PR #6, opened 2016-03-29
replace memory region with memory pool API

facao:
Check if an new branch is created.


whchung:
This one still asks to go to master branch

PR #15, opened 2016-04-06
Add is_peer and get_peers API to hc::accelerator

facao:
As title described.

### DIFF
--- a/include/hc.hpp
+++ b/include/hc.hpp
@@ -765,25 +765,10 @@ public:
      *
      * @return true if other can access this accelerator's device memory pool or false if not
      */
-    bool is_peer(const accelerator& other) const {
-      hsa_amd_memory_pool_t* self_pool = static_cast<hsa_amd_memory_pool_t*>(get_hsa_am_region());  
-      // If pool handle is set during initialization, then it means, no local pool, return false.
-      if(self_pool->handle == uint64_t(-1))
-          return false;
-
-      hsa_agent_t* other_agent = static_cast<hsa_agent_t*>(other.get_hsa_agent());
-      hsa_amd_memory_pool_access_t access;
-
-      hsa_status_t status = hsa_amd_agent_memory_pool_get_info(*other_agent, *self_pool, HSA_AMD_AGENT_MEMORY_POOL_INFO_ACCESS, &access);
-      if(HSA_STATUS_SUCCESS != status)
-          return false;
-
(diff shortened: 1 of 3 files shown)


whchung:
In general I get the idea of this pull request and the logic looks fine. However I'd like to ask you to help refactor it a bit to move the implementation away from the interface (hc.hpp). Reason of it is to possibly allow other targets (ex: OpenCL) for the same interface.

You should have noticed implementations of many member functions in hc::accelerator and hc::accelerator_view are actually not carried out in hc.hpp, but in mcwamp_hsa.cpp in lib/hsa. I made the design so we could have one interface but with multiple targets (x86 CPU / OpenCL / HSA).

Steps to do so would be:

- define new virtual member functions in kalmar_runtime.h , provide a default dummy implementation there. please do not declare them as pure virtual, just virtual.
- place implementation logic of the virtual member functions inside mcwamp_hsa.cpp .
- call the virtual member function from the interface (hc.hpp).

include/hc.hpp
```diff
@@ -23,6 +23,7 @@
 
 #include <hsa_atomic.h>
 #include <kalmar_cpu_launch.h>
+#include <hsa_ext_amd.h>
```
whchung: This line marks a direct dependency from HC API to AMD HSA extension API, which is too strong a coupling, and would not be a good idea from API implementation point of view.

include/hc.hpp
```diff
@@ -755,10 +756,52 @@ class accelerator
...
         return pDev->getHSAAgent();
     }
 
+    /**
+     * Check if @p other is peer of this accelerator.
+     *
+     * @return true if other can access this accelerator's device memory pool or false if not
+     */
+    bool is_peer(const accelerator& other) const {
+      hsa_amd_memory_pool_t* self_pool = static_cast<hsa_amd_memory_pool_t*>(get_hsa_am_region());
```
whchung: The implementation of hc::accelerator::is_peer() directly uses AMD HSA extension API. It breaks separation of interface and implementation we have so far, and would make supporting other targets (x86 CPU / OpenCL) rather hard. Please try define a virtual member function in KalmarDevice, and have the HSA implementation placed in mcwamp_hsa.cpp .

facao:
Move to HSA back end.

include/hc.hpp
whchung: @facao Could you remove this line from hc.hpp ? As there is no HSA runtime API directly used in hc.hpp we should remove it to completely remove the dependency.

whchung:
I've changed this pull request by removing AMD HSA extension header from hc.hpp .

PR #19, opened 2016-04-11
Added feature for locking host memory

adityaatluri:


whchung:
could you make it go to a branch first? also it would be good if a unit test in hcc layer be provided.

adityaatluri:
Hi,
Can you provide the directory where the test goes?

whchung:
@adityaatluri Please place them under tests/Unit directory. You may want to discuss with @facao on a common directory name as both of you are working on AM-related features.

include/hc_am.hpp
```diff
@@ -155,5 +155,9 @@ void am_memtracker_print();
 void am_memtracker_sizeinfo(const hc::accelerator &acc, size_t *deviceMemSize, size_t *hostMemSize, size_t *userMemSize);
 
 
+am_status_t am_memtracker_host_memory_lock(hc::accelerator &acc, void *ptr, size_t size);
```
whchung: @adityaatluri As these two new functions are public APIs and would supposedly be invoked by user applications, could you help provide some comments for them?

PR #21, opened 2016-04-12
add unit test and modify map_to_peer api

facao:


include/hc_am.hpp
```diff
@@ -19,7 +19,7 @@ struct AmPointerInfo {
     void *      _hostPointer;   ///< Host pointer.  If host access is not allowed, NULL.
     void *      _devicePointer; ///< Device pointer.  
     size_t      _sizeBytes;     ///< Size of allocation.
-    hc::accelerator &_acc;       ///< Device / Accelerator to use.
```
whchung: @facao , could you help elaborate why we need a copy instead of a reference?

whchung: @facao and @bensander . after our discussion I think it's better to change this field to hc::accelerator* , instead of hc::accelerator& (so we don't have undefined semantic), or hc::accelerator (so we save memory).

tests/Unit/Pool/accelerator_get_is_peer_api.cpp
```diff
@@ -0,0 +1,29 @@
+#include <hc.hpp>
```
whchung: @facao missing lit macro. could you help follow the examples of other unit tests and apply them in the tests in this pull request?

tests/Unit/Pool/accelerator_get_peers.cpp
```diff
@@ -0,0 +1,26 @@
+#include <hc.hpp>
```
whchung: @facao missing lit macro

tests/Unit/Pool/map_to_peers_device_ptr.cpp
```diff
@@ -0,0 +1,21 @@
+#include <hc_am.hpp>
```
whchung: @facao missing lit macro

tests/Unit/Pool/map_to_peers_host_ptr.cpp
```diff
@@ -0,0 +1,21 @@
+#include <hc_am.hpp>
```
whchung: @facao missing lit macro

whchung:
@facao Per our discussion I'm for the idea moving to std::shared_ptr<hc::accelerator> . This should be a coordinated change in both HIP and HCC. So I'd suggest you:

- forfeit the current pull request.
- submit a new one with only the test cases in, as they shouldn't be affected by the change in AmPointerInfo.
- make the changes in both HIP and HCC, and submit new pull requests to both HIP and HCC, for the proposed change in AmPointerInfo.

include/hc_am.hpp
whchung: As @facao pointed out, hc::accelerator is just a wrapper of KalmarDevice* , and I've verified that sizeof(hc::accelerator) is only 8. So we should be fine copying it around.

whchung:
@facao There are conflicts in this pull request. Could you help resolve them so we can adopt this pull request?

tests/Unit/Pool/accelerator_get_is_peer_api.cpp
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Test if hc::accelerator::get_is_peer() works fine.
+ * Create the default accelerator and check if others 
+ * is peer of it.
+ * accelerator is peer of itself.
+ * FIXME: on current system, dGPU is peer of any each  
+ * other, we should expect is_get_peer() return true.
+ */
```
whchung: @facao  the unit test assume all dGPUs are peers to each other, which may be true for supported configuration. but for systems without P2P support peers.size() would be zero and this test would fail.

as the result of the test depends on the hardware configuration, and in many cases won't simply work, so I'd propose ways to improve this test case. And I'm more inclined to ask you improve the test as the 2nd method.

1. relax this test so as long as long as acc.get_is_peer() could return a valid value, then call the test pass. this would reduce the test case to merely check whether the API is available in the header file.
1. change the test case, and use HSA runtime API to query if the hardware do have HSA agents who are peers, and then check if the result of acc.get_is_peer() agree with it. You can refer to the following unit test as an example:

https://github.com/RadeonOpenCompute/hcc/blob/master/tests/Unit/AsyncPFE/accelerator_view_wait.cpp

PR #24, opened 2016-04-13
added feature for locking host memory

adityaatluri:


tests/Unit/Lock/lock_host_pointer.cpp
```diff
@@ -0,0 +1,24 @@
+// XFAIL: Linux
+// RUN: %hc %s -o %t.out %t.out
```
whchung: @adityaatluri this line looks suspicious. The test is supposed to be linked with -lhc_am but it's not specified here.

whchung:
@adityaatluri could you help:

1. address my comment inside the codes
1. alter the pull request so it goes to a feature branch "mem_lock" first
1. provide test results of "HCC_RUNTIME=HSA make test"
1. provide test results of HIP unit tests

tests/Unit/Lock/unlock_host_pointer.cpp
```diff
@@ -0,0 +1,34 @@
+// XFAIL: Linux
+// RUN: %hc %s -o %t.out %t.out
```
whchung: @adityaatluri same here. perhaps -lhc_am is needed so lit could correct build the test

adityaatluri:
Hi @whchung ,
Except for point 1, I have no idea about the rest you are talking about. I don't know about the unit test infrastructure of hcc. I want to add this functionality for friday release.

whchung:
Hi Aditya,

For point 2, normally I'd like to have new features go to a branch first.
And I'll merge them into master once I verify it again. And this pull
request wants to merge 1 commit into RadeonOpenCompute:master from
adityaatluri:mem_lock . I'd like to ask you change it so it goes to
RadeonOpenCompute:mem_lock . On bitbucket it's quite intuitive to do so,
but it seems Github UI is a bit strange here. You should be able to find
"mem_lock" when you create a pull request.

For point 3 & 4, please run "HCC_RUNTIME=HSA make test" and HIP unit tests
as a guarantee that your new check-ins don't break existing stuffs. You can
refer to this pull request as an example.

https://github.com/RadeonOpenCompute/hcc/pull/20

Jack

2016-04-14 10:18 GMT+08:00 Aditya Atluri <notifications@github.com>:

tests/Unit/Lock/lock_host_pointer.cpp
```diff
@@ -0,0 +1,24 @@
...
+#include <hc.hpp>
+
+int main()
+{
+    hc::accelerator acc;
+    const auto& all = hc::accelerator::get_all();
+    if(all.size() == 0){
+        return -1;
+    }
+    std::vector<hsa_agent_t> agentVec;
```
whchung: please consider not using HSA data structures directly in hcc unit tests.

if you must use them. please refer to:
https://github.com/RadeonOpenCompute/hcc/blob/master/tests/Unit/HC/create_marker.cpp
for an example. you'll basically need to change lit macro to make your test build.

include/hc_am.hpp
```diff
@@ -174,6 +174,28 @@ void am_memtracker_update_peers(const hc::accelerator &acc, int peerCnt, hsa_age
...
+ * Locks a host pointer to a vector of agents
+ * 
+ * @p ac acclerator corresponding to current device
+ * @p hostPtr pointer to host memory which should be page-locked
+ * @p size size of hostPtr to be page-locked
+ * @p accVec vector to hsa agents to which the hostPtr should be visible
+ * @return AM_SUCCESS if lock is successfully.
+ * @return AM_ERROR_MISC if lock is unsuccessful.
+ */
+am_status_t am_memory_host_lock(hc::accelerator &ac, void *hostPtr, size_t size, std::vector<hsa_agent_s> &accVec);
```
whchung: could you consider remove the dependency of hsa_agent_s here? for example, change them to std::vector<void*> ? normally we want to avoid having direct HSA dependency in HCC public headers, so it's possible to build x86 simulation path or OpenCL path.

adityaatluri: Sure

bensander: Right - for an HC interface we should use a vector of accelerators rather than agents.  Then in the lock implementation create the agent list from the accelerator (there is an hc::accelerator:: get_hsa_agent() that might help here).
Also if does the agent where the pointer is allocated also need to be on the visibility list?  I think the answer is YES – we should make this more clear in the description.
-ben

PR #38, opened 2016-04-23
Cleanup wrapper IR functions

arsenm:
Use new workitem intrinsics + range metadata, correct
some attributes on functions, and canonicalize.

Correct range metadata to be maximum theoretical workgroup size.

Change-Id: I9dedbe2dd62753858ccd0eb7841e228873a2c031


whchung:
@arsenm could you alter this pull request so it goes to "develop" branch? We are moving to a branching strategy more aligned with HIP, where all the latest stuffs go to "develop" first.

arsenm:
Maybe the default github branch should be changed?

whchung:
master should still be the default branch because that's where users fetch production-ready commits.

GitHub UI is less favorable than Bitbucket here because you can't create a new topic branch along with a pull request.

I'll create a CONTRIBUTE.md to explain the rules better.

PR #39, opened 2016-04-23
Cleanup wrapper IR functions

arsenm:
Use new workitem intrinsics + range metadata, correct
some attributes on functions, and canonicalize.

Correct range metadata to be maximum theoretical workgroup size.

Change-Id: I9dedbe2dd62753858ccd0eb7841e228873a2c031

### DIFF
--- a/lib/hsail-amdgpu-wrapper.ll
+++ b/lib/hsail-amdgpu-wrapper.ll
@@ -1,77 +1,73 @@
 ; ModuleID = 'hsail-amdgpu'
 
-; Function Attrs: alwaysinline
+@"&hcc_static_group_segment_size" = addrspace(1) global i32 0, align 4
+@"&hcc_dynamic_group_segment_size" = addrspace(1) global i32 0, align 4
+
+; Function Attrs: alwaysinline nounwind readonly
 define linkonce_odr spir_func i32 @__hsail_get_global_id(i32) #0 {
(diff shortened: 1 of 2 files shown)


lib/hsail-amdgpu-wrapper.ll
```diff
@@ -393,170 +388,155 @@ define linkonce_odr spir_func float @__hsail_ftz_f32(float) #1 {
...
+  ret i32 %6
 
-; <label>:6                                      ; preds = %1
-  %size_z_ptr = getelementptr i32 ,i32 addrspace(2)* %dispatch_ptr_i32, i32 2
-  %7 = load i32, i32 addrspace(2)* %size_z_ptr, align 4, !invariant.load !0, !range !1
-  ret i32 %7
+; <label>:7:                                      ; preds = %1
+  %size_z_ptr = getelementptr i8, i8 addrspace(2)* %dispatch_ptr, i64 8
+  %8 = bitcast i8 addrspace(2)* %size_z_ptr to i32 addrspace(2)*
+  %9 = load i32, i32 addrspace(2)* %8, align 4, !range !2, !invariant.load !0
```
scchan: @arsenm  There's no metadata !2 in the cleaned up version, could you take a look?

arsenm: It should be !1

PR #43, opened 2016-04-27
Update debian package

scchan:
- don't create symlinks for clang and clang++ in /opt/rocm/bin
- add "section devel" to the control file


whchung:
@scchan should this pull request go to master directly, or could it go to develop branch?

scchan:
I'd put it in master because they went into the roc-1.0.x branch as well

whchung:
I see. Let me get it into master.

PR #44, opened 2016-04-29
use FNV-1a for kernel indexing instead of md5

yan-ming:
this will improve hcc runtime performance when multiple kernels are used
in a program

`make test`:

```
********************
Testing Time: 633.70s
********************
Failing Tests (7):
    CPPAMP :: Unit/AmpShortVectors/hc_short_vector_device.cpp
    CPPAMP :: Unit/HC/wg_size.cpp
    CPPAMP :: Unit/HSAIL/shfl_xor.cpp
    CPPAMP :: Unit/SharedLibrary/shared_library2.cpp
    CPPAMP :: Unit/SharedLibrary/shared_library3.cpp

  Expected Passes    : 661
  Expected Failures  : 25
  Unsupported Tests  : 10
  Unexpected Failures: 7
```

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -72,6 +72,9 @@
 #define ASYNCOPS_VECTOR_GC_SIZE (1024)
 
 
+// whether to use MD5 as kernel indexing hash function
+// default set as 0 (use faster FNV-1a hash instead)
+#define USE_MD5_HASH (0)
 
 static const char* getHSAErrorString(hsa_status_t s) {
 
@@ -1323,6 +1326,7 @@ public:
 
     // calculate MD5 checksum
     std::string MD5Sum(size_t size, void* source) {
+#if USE_MD5_HASH
         unsigned char md5_hash[16];
         memset(md5_hash, 0, sizeof(unsigned char) * 16);
         MD5_CTX md5ctx;
@@ -1337,6 +1341,22 @@ public:
         }
 
         return checksum.str();
+#else
+        // FNV-1a hashing, 64-bit version
+        const uint64_t FNV_prime = 0x100000001b3;
+        const uint64_t FNV_basis = 0xcbf29ce484222325;
+        uint64_t hash = FNV_prime;
+
+        const char *str = static_cast<const char *>(source);
+
+        const int N = 104;
+        size = size > N ? N : size;
+        for (auto i = 0; i < size; ++i) {
+            hash ^= *str++;
+            hash *= FNV_prime;
+        }
+        return std::to_string(hash);
+#endif
     }
 
     void BuildProgram(void* size, void* source, bool needsCompilation = true) override {
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -1325,7 +1325,7 @@ public:
     }
 
     // calculate MD5 checksum
-    std::string MD5Sum(size_t size, void* source) {
+    std::string kernel_checksum(size_t size, void* source) {
 #if USE_MD5_HASH
         unsigned char md5_hash[16];
         memset(md5_hash, 0, sizeof(unsigned char) * 16);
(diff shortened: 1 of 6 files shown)


yan-ming:
Hey @whchung,

Could you please review this PR?

From my experiments of this patch in the gpu_deconvolution work, the total execution time has been reduced from 3m10s to 1m59s, which is a big improvement.

whchung:
@yan-ming thanks for this contribution.  Could you alter this PR so it goes to "develop" branch? We are trying to adopt a new branching strategy now.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1337,6 +1341,23 @@ class HSADevice final : public KalmarDevice
...
         return checksum.str();
+#else
+        // FNV-1a hashing, 64-bit version
+        const uint64_t FNV_prime = 0x100000001b3;
+        const uint64_t FNV_basis = 0xcbf29ce484222325;
+        uint64_t hash = FNV_basis;
+
+        const char *str = static_cast<const char *>(source);
+
+        // 104 is the proper size from Jack's research
```
whchung: It's 140, not 104

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1323,6 +1326,7 @@ class HSADevice final : public KalmarDevice
 
     // calculate MD5 checksum
     std::string MD5Sum(size_t size, void* source) {
```
whchung: For better clarify, how about change the function name as well?

yan-ming:
Hi @whchung,

Sure let me change the PR destination after I make sure everything is okay.

lib/hsa/mcwamp_hsa.cpp
yan-ming: Oops, I just quoted that directly from your mail. Let me fix it.

whchung: Sorry, 104 should be the correct number.

whchung: On a second thought, could you make it a macro? The rationale is because a new code object format is being developed so this value may have to be changed soon.

yan-ming: Could you give me the desired name of the macro?

whchung: I'm still not very comfortable at merely passing the size of header to the hash algorithm. Take BrigModuleHeader as an example, originally it shall contain the hash of the BRIG module itself. But it's all filled by 0 in the current implementation. Also in LC backend we are moving toward completely ELF-compatible format so it's likely we run into same ELF headers for similar kernels.

Could you help study the impact to performance if we don't use this 104 heuristic?

whchung: For reference purpose, here are sources where I got this 104 magic number. It's the larger value of these 2 kinds of headers.

- HSA BRIG module header: http://www.hsafoundation.com/html/Content/PRM/Topics/18_BRIG/BrigModuleHeader.htm
- ELF64 header: https://refspecs.linuxfoundation.org/elf/gabi4+/ch4.eheader.html

yan-ming: So the gpu_deconvolution work is actually using LC backend (due to hcFFT). If I use the original size in FNV-loop, the execution time is about 2m37s.

Here's the comparison of each condition:

- original: 3m10s
- FNV hash: 2m37s
- FNV hash + 104 cutoff: 1m59s

Personally I would prefer having FNV hash with some fixed size cutoff for the best performance.

whchung: Thanks for providing the values. To boost performance let's make a macro to hold this magic number then. How about call it "KERNEL_CODE_OBJECT_HEADER_SIZE"? And put comments so we know how it's derived (bigger value of HSA BrigModuleHeader and Elf64_Ehdr).

whchung: @yan-ming what you proposed ( ```FNV1A_CUTOFF_SIZE``` ) may be a better name after all.

yan-ming:
@whchung it seems like github doesn't allow me to change PR destination branch directly, please look to PR #45, sorry for the inconvenience.

whchung:
@yan-ming it's alright. github UI is less favorable compared to Bitbucket in this regard.

PR #51, opened 2016-05-05
clamp-hsatools: Use lld as the liner for AMDGPU

tstellarAMD:
We need to wait until co-v2 support lands in LLVM before merging this.

### DIFF
--- a/lib/clamp-hsatools.in
+++ b/lib/clamp-hsatools.in
@@ -47,7 +47,7 @@ fi
 HSA_USE_AMDGPU_BACKEND=@HSA_USE_AMDGPU_BACKEND@
 
 if [ $HSA_USE_AMDGPU_BACKEND == "ON" ]; then
-  LLD=@AMDPHDRS@
+  LLD=@HSA_LLVM_BIN_DIR@/ld.lld
   KM_USE_AMDGPU="${KM_USE_AMDGPU:=1}"
 fi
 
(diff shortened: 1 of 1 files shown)


whchung:
@tstellarAMD could you help revise the pull request so this goes to "develop" branch?

tstellarAMD:
These needs to go in the master branch, otherwise we will break master when llvm updates to code object v2.

whchung:
@tstellarAMD We are changing to a new branching strategy now. Every new feature goes to develop first. Periodically we'll promote production-ready features to master.

tstellarAMD:
Would it be possible to merge it into both branches?  If we only merge it into develop, then master won't work at all with lightning, which would defeat the purpose of having a stable master branch.  I think this commit is a special case, because it changes the interface between hcc and lightning.  I don't think we are going to have many commits like this.

whchung:
Ok, let me reopen this pull request. Please give me a signal when it's good to me merged into master.  I can then cherry-pick it to develop.

tstellarAMD:
This can be merged now.  I'm going to push the LLVM changes in a few minutes.

whchung:
@tstellarAMD are those changes in LLVM side take place in LLVM trunk? could you let me know the commit # so I'm sure my env is up to speed?

tstellarAMD:
It was committed as svn revision r268647.

whchung:
@tstellarAMD In terms of git commit, that's:
https://github.com/llvm-mirror/llvm/commit/66eb4d17bba8e55395434deb192592b1a5edbdff
right?

PR #57, opened 2016-05-06
hsail-amdgpu-wrapper: Add range metadata to mbcnt intrinsics

tstellarAMD:
As used in the hsail_get_lane_id() intrinsic, these intrinsics
always return a value between 0 and 63.  The range metadata
gives this information compiler to help improve the optimizer.


scchan:
let's get this into master as well, will help shfl functions

whchung:
@tstellarAMD @scchan both master and develop contains this now.

PR #58, opened 2016-05-07
avoid loading compute platform library during platform detection

scchan:
- rocr team requests hcc not to load the libhsa-runtime64 library from its soname
- platform detection could be achieved by loading the hcc runtime library since it depends on either libhsa-runtime or libOpenCL


whchung:
@scchan should this go into master now?

scchan:
yep

whchung:
@scchan both master and develop contains this now.

PR #61, opened 2016-05-20
Revert "[builtin][LC] use inline assembly for 32-bit FMA operation"

arsenm:
This reverts commit 0213640450ee5af2a50643e5bf6e93c74c3bdd18.

This is entirely incorrect. v_mad_f32 is not equivalent to fma.
We also very much do not want to be using inline asm for common
operations.

### DIFF
--- a/lib/hsail-amdgpu-wrapper.ll
+++ b/lib/hsail-amdgpu-wrapper.ll
@@ -146,12 +146,8 @@ declare double @llvm.floor.f64(double) #1
 
 ; Function Attrs: alwaysinline nounwind readnone
 define linkonce_odr spir_func float @__hsail_fma_f32(float, float, float) #2 {
-  ; Use inline assembly to leverage v_mad_f32 which offers better precision than @llvm.fma.f32()
-  %ret = tail call float asm "v_mad_f32 $0, $1, $2, $3", "=v,v,v,v"(float %0, float %1, float %2) #
-  ret float %ret
-
-  ;%4 = call float @llvm.fma.f32(float %0, float %1, float %2)
-  ;ret float %4
+  %4 = call float @llvm.fma.f32(float %0, float %1, float %2)
+  ret float %4
 }
 
 ; Function Attrs: nounwind readnone


adityaatluri:
Hi @arsenm ,
As a general question, why "no inlining" for common operations? How different is it from code translation (One I can think of is compiler optimization).

arsenm:
None of the optimizers will see it, and inline asm pessimizes many things. I believe this is also unduly constraining the register classes of the operands

arsenm:
This also won't use source modifiers when possible

adityaatluri:
Awesome! 
Question 2: How can I use the source from inline asm with amdphdrs? Is it compatible? I am compiling to .o file using llc.

arsenm:
I'm not sure what you mean by using the source from inline asm.

adityaatluri:
I mean, if I use this code[1], can I use it with amdphdrs? 
[1]. https://gist.github.com/adityaatluri/2b911cbbc878d2d4b0305e75231acbb7

arsenm:
I'm not sure what you would do with it. It's not a kernel and doesn't have any of the necessary ABI to work. There are also no functions or linking of external functions

whchung:
@arsenm @adityaatluri The commit was created in order to fix an issue reported by @NEELMCW for applications he's been working on. You can find the test at:

https://gist.github.com/whchung/135cf3e3ebabeaec1d5c9784ecdd74b7

The commit fixes this test case.

If the semantic of v_mad_f32 is different from v_fma_f32, do you think should hcc frontend expose another API (ex: hc::precise_math::mad() ) and let applications use it? Also if it's not a good idea to use inline assembly, could you help recommend an LLVM intrinsic to use?

arsenm:
There is no need for an intrinsic. v_mad_f32 is less precise and has an intermediate rounding step, and never respects denormals. v_mad_f32 is already matched from a basic multiply and add. This test looks bogus and is testing two different functions

arsenm:
That testcase would only be expected to produce different results for the f64 version if fp-contract or fast math were enabled

adityaatluri:
Hi @whchung @arsenm ,
If I want to do something like this, https://gist.github.com/adityaatluri/2b911cbbc878d2d4b0305e75231acbb7#file-vadd-asm-ll
How should I make it work? (I know that code cannot be compiled, added to give an idea of doing vector add).

whchung:
@arsenm could you try build and run the test on gist? As reported by @NEELMCW , we do see incorrect results on f32 version. So that's why this commit was created to experiment whether using v_mad_f32 be a better idea than v_fma_f32.

@NEELMCW , could you also chime in here? @arsenm made some good points on why inline asm wouldn't be a good idea . And to my knowledge you are still mostly working on HSAIL backend. So say if we revert this commit would your application development be blocked?

arsenm:
Using inline asm for normal instructions like this is fine

arsenm:
This test behaves as expected. If I fix it t consistently use mad or fma between host and device it passes.

whchung:
@arsenm let me try summarize here. could you find if my understanding is correct? I'll merge this pull request after your confirmation.

- The test case itself needs be fixed, not math functions here. The root cause of failure is discrepancy between host code (which uses vanilla * followed by +) and kernel code (which uses fma()). The number of rounding caused the difference.
- In general it's not a good idea to use inline assembly because it forbids all kinds of optimization, plus it puts strict constraints on register types used. But for this particular case it is ok.

@NEELMCW I've revised the test and it passes now on LC even without this specific commit. Please take a look at:
https://gist.github.com/whchung/135cf3e3ebabeaec1d5c9784ecdd74b7/revisions

arsenm:
Yes, the test needs to be fixed.

Yes, inline assembly is bad for those reasons. I'm not sure what you mean by in this particular case it is OK. The asm in this case will prevent any optimization or operand folding as would any other asm.

whchung:
You mentioned "Using inline asm for normal instructions like this is fine" in one of your earlier comment, so I thought this particular case is ok.

Anyway, let me go ahead and merge this pull request.

arsenm:
That was only referring to that using inline assembly would work technically, not that it is a good idea to do so

NEELMCW:
@arsenm  @whchung  Thanks for the clarification.

PR #68, opened 2016-06-01
remove redundant runtime boostrap object

scchan:
remove redundant runtime boostrap object

### DIFF
--- a/lib/mcwamp.cpp
+++ b/lib/mcwamp.cpp
@@ -438,9 +438,6 @@ public:
   }
 };
 
-// this would initialize Kalmar runtime before main() in user program begins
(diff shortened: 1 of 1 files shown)


lib/mcwamp.cpp
```diff
@@ -438,9 +438,6 @@ class KalmarBootstrap {
   }
 };
 
-// this would initialize Kalmar runtime before main() in user program begins
-static KalmarBootstrap boot;
```
whchung: __hcc_shared_library_init() achieves the same functionality, right?

scchan: yep, does the same thing

PR #77, opened 2016-06-20
add half precision type

scchan:
add half precision floating point


whchung:
This one should be harmless. But I think for experimental stuffs we should keep them in topic branches, and merge them only when we have enough tests available.

PR #82, opened 2016-06-24
[HSA] Move checksum computation from runtime to compile time.

UncleHandsome:
Hello, Jack
This is my idea to avoid checksum computation at runtime.

Currently, I use utility program md5sum to generate the checksum.
Should I write another little program to generate the checksum in case md5sum is not present??
Thank you


whchung:
md5sum is in coreutils so it should be universally available.

on the other hand, could you:

1. remove other traces of "md5.h" dependency in the source tree? OpenCL runtime, CMakeLists.txt, md5.cpp to be particular
2. provide test results

UncleHandsome:
Hello, Jack
I cannot run the all conformance-all because the HSA stack will freeze some how....
Currently, among 2173 tests, there are 189 failed case.

For the OpenCL Runtime, since I don't have OpenCL machine immediately, I may not be able to deliver right now.

Thank you

whchung:
@UncleHandsome, many thanks for submitting this PR. We will soon move to a new kernel archive format so arbitrary number of GPU architectures be supported. Therefore this PR wouldn't be needed.

PR #84, opened 2016-06-24
[hcc rt] clear the dependent aysnc op vector correctly

scchan:
clear the dependency vector instead of its copy


scchan:
https://github.com/RadeonOpenCompute/hcc/pull/85

PR #85, opened 2016-06-24
Fix performance problem when launching lots of kernels

pfultz2:
This fixes a problem where the `dependentAsyncOpVector` in the `bufferKernelMap` is never cleared, instead just a local copy is cleared. This drastically improves performances when the program launches a lot of kernels.


pfultz2:
Oops, I guess this is already addressed in PR#84.

scchan:
I closed mine and I'll let Jack to review this patch

whchung:
@pfultz2 thanks for spotting this bug. I'll take it.

PR #86, opened 2016-06-27
HCC CU mask APIs

facao:
Added one Unit test to tests/Unit/Pool, looks like it is disabled? ' make test' output "Expected Failure".


whchung:
@facao could you file the PR to develop branch?

PR #87, opened 2016-06-27
HCC CU Mask APIs

facao:


whchung:
If I understand this PR correctly, set_cu_mask operates on accelerator_view, and get_cu_count operates on accelerator, right?

There is a test case for get_cu_count, could you also provide a test case which uses set_cu_mask? At least the test should:
- uses the set_cu_mask API
- check its returning value be true
- launches a kernel
- verify the calculation is correct

PR #91, opened 2016-06-30
implement --help for hcc-config / clamp-config

aunzat:
Please note, that I have omitted --verbose & --brief since they appear to have no function.
The descriptions are based on the source code; I hope they are ok like this.

Also C++11 is needed, since raw string literals are used (I can change this, if needed).


whchung:
@aunzat Thanks a lot for your contribution!

PR #98, opened 2016-07-19
added fix for signal overflow for kernels

adityaatluri:
With this change, errors and bad dmesg errors generated by HIP samples which launch more than 32k kernels is fixed.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -2627,11 +2627,7 @@ HSADispatch::dispatchKernelAsync(Kalmar::HSAQueue* hsaQueue) {
 inline void
 HSADispatch::dispose() {
     hsa_status_t status;
-    if (kernargMemory != nullptr) {
-      device->releaseKernargBuffer(kernargMemory, kernargMemoryIndex);
-      kernargMemory = nullptr;
-    }
-
+    
     clearArgs();
     std::vector<uint8_t>().swap(arg_vec);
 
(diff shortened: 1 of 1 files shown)


whchung:
@adityaatluri Could you provide information on which applications / tests be fixed with this PR? And would it be possible to come up with a reduced test case to depict the issue?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2557,7 +2560,7 @@ HSADispatch::waitComplete() {
         printf("Signal wait returned unexpected value\n");
         exit(0);
     }
-
+    dispose();
```
whchung: call dispose() here would render the "if (kernargMemory != nullptr)" block below redundant. could you remove it?

whchung:
@adityaatluri as we discussed over email, generally I like the idea behind this PR so HSA signals get released sooner to the resource pool. Could you also do the same for HSABarrier?

lib/hsa/mcwamp_hsa.cpp
adityaatluri: Hi Jack, by removing it, we can't reclaim the signals. I couldn't get what you want me to change.

adityaatluri:
Hi Jack,
Sure. I'll create another pull request for it. The impact of this change resolves lots of issues.

lib/hsa/mcwamp_hsa.cpp
whchung: sorry I should be more clear. I'm not referring to remove this line, but the block "if (kernargMemory != nullptr)" below it. (line 2565 - 2568). the logic would be redundant if we call dispose() here.

adityaatluri: Oh ya! Do you want me to create a new pull request?

whchung:
@adityaatluri Thanks. I'll take this now. Please submit another PR with:

- changes to HSABarrier.
- a reduced test case.

PR #107, opened 2016-07-28
cmake cleanup

jvesely:


whchung:
@jvesely thanks for providing this PR. Could you:

1) shortly describe what this PR would like to achieve?
2) change this PR so it goes to "develop" branch first?

CMakeLists.txt
```diff
@@ -516,8 +516,7 @@ if (HSA_USE_AMDGPU_BACKEND AND HSA_LLVM_BIN_DIR)
 
   # create a symlink to the LLVM binaries for the amdgpu backend in the project directory
   add_custom_target(lightning_bin
-    COMMAND mkdir -p "${PROJECT_BINARY_DIR}/llvm/bin"
-    COMMAND cp -rp "${HSA_LLVM_BIN_DIR}/*" "${PROJECT_BINARY_DIR}/llvm/bin/"
+    COMMAND ln -sf "${HSA_LLVM_BIN_DIR}" "${PROJECT_BINARY_DIR}/llvm/bin"
```
whchung: @scchan I believe using "cp" was made by you earlier this month. Would it affect HCC if we change to "ln" here?

scchan: If I recall correctly, symlinks would break the debian packaging

jvesely: copying binaries does not work, and breaks build if llvm is in non-default location.
$ chrpath -l bin/llvm-link
bin/llvm-link: RUNPATH=$ORIGIN/../lib

resulting in "failed to load library" errors during build
the script either needs to change rpath, or copy libraries to a corresponding location as well. or fix the packaging script, no idea which one is the easiest.
I'd argue that packaging binaries without corresponding libraries is a bad idea anyway

jvesely:
Not sure if I can change the target of PR. I'll close this one and open a new one against develop pending the outcome of this discussion

More details are provided in individual commits.
0953112 adds pseudo targets for opt and llvm-link. these are not specified if configured to USE_AMDGPU_BACKEND, and every target that adds dependencies on them results in cmake complaining about non-existent targets opt and llvm-link
0640d33 discussed inline
61696ed cleans up MCWAMP handling. the original code would modifiy global CXX flags, as a result the commandline would be append 3 times and used for every compilation target. (you can check CMakeFiles/xxxxx.target/flags)

jvesely:
droppped the symlink patch and moved to develop in https://github.com/RadeonOpenCompute/hcc/pull/108

PR #110, opened 2016-08-04
enable new linking algorithm to support static library with gpu code

scchan:


whchung:
@scchan could you help double check if the new linking algorithm do contain latest features I added in clamp-link? Ex: AMDGPU target selection.

whchung:
also it seems in this PR, we simply turn the new algorithm back on. does it truly mean all issues we saw in applications gone?

scchan:
Those failures are gone.   I suspect it was exposing some build issues in hctorch.  Now it builds and runs fine with the new linking algorithm.

I updated the logic for AMDGPU_TARGET and HCC_AMDGPU_TARGET and saw the isa-to-name translation works fine

PR #126, opened 2016-09-21
submodule: Update the submodule config to allow anonymous checkouts

tstellarAMD:
The current urls don't seem to work if I don't have an ssh key.

I don't have any experience using submodules, but this seems to work for me.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,12 +1,12 @@
 [submodule "llvm"]
 	path = compiler
-	url = git@github.com:RadeonOpenCompute/llvm.git
+	url = https://github.com/RadeonOpenCompute/llvm.git
 	branch = amd-hcc
 [submodule "lld"]
 	path = lld
-	url = git@github.com:RadeonOpenCompute/lld.git
+	url = https://github.com/RadeonOpenCompute/lld.git
 	branch = amd-common
 [submodule "clang"]
 	path = clang
-	url = git@github.com:RadeonOpenCompute/hcc-clang-upgrade.git
+	url = https://github.com/RadeonOpenCompute/hcc-clang-upgrade.git
 	branch = clang_tot_upgrade


pfultz2:
This will no longer clone the submodules using ssh.

tstellarAMD:
> This will no longer clone the submodules using ssh.

That was my goal.  Is there some downside to not using ssh?

pfultz2:
Yes, you have to enter the password everytime, which besides the annoyance, can also hinder automation.

whchung:
Like @pfultz2 said, using SSH bring much more benefits (no password, automation) than HTTPS. I'll close this pull request for now. Once ToT HCC becomes more stable, and when we separate SDK/runtime/compiler we can consider going back to HTTPS.

tstellarAMD:
Hi,

The URL I used is an anonymous URL, so no password is required.  This patch actually makes it possible for me to do automation with minimal setup.  With the current url I have to install an ssh key on my tests systems.

whchung:
@tstellarAMD looking through the documentation ( https://git-scm.com/book/en/v2/Git-Tools-Submodules ) I found this:

git config submodule.DbConnector.url PRIVATE_URL

Let me do some tests first before merging this PR.

tstellarAMD:
Did you get a chance to look at this?

whchung:
I tested it the other day but it doesn't really work the way I anticipated. I'm still looking at other ways. Bottom line is I don't want to make my routine synchronization with upstream obscured by adopting HTTPS, as the process involves pushing to all submodules.

tstellarAMD:
The problem with the current URLs is that in order to clone the submodules, you need a github account and need to have configured ssh so that you can access github via ssh without a password.  I think this excludes a large number of people from using hcc.

It's still not clear to me why switching to the https urls will make things more difficult.

whchung:
@tstellarAMD I have to commit to submodules to keep up with upstream changes, and HTTPS bothers me a lot because I have to enter password every time for every submodule, so that's why I used SSH in the module config file. I do understand the fact it imposes restrictions to people that a key is needed, and I'm looking into ways to locally change git remote for submodules, so we can use HTTPS by default (great for users), plus still be able to push commits easily (great for developers / maintainers).

whchung:
So the correct way for me to do is to go inside each submodule and do:

git remote set-url --push origin "SSH-based URL"

With that users can still clone with HTTPS, and I can push to SSH-based URLs.

PR #127, opened 2016-09-22
Utilize all the cores when building and running tests

pfultz2:
This sets up the cmake to call the 'ProcessorCount' module to build and test with all the cores available, instead of just setting it to 8.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,7 +28,8 @@ ENDIF (NOT CLANG_URL)
 set(CMAKE_MACOSX_RPATH 1)
 
 IF (NOT NUM_BUILD_THREADS)
-  SET(NUM_BUILD_THREADS 8)
+  include(ProcessorCount)
+  ProcessorCount(NUM_BUILD_THREADS)
 ENDIF(NOT NUM_BUILD_THREADS)
 
 include (EnsureLLVMisPresent)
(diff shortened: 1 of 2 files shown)


whchung:
@pfultz2 thanks. learnt a new function in CMake.

CMakeLists.txt
```diff
@@ -28,7 +28,8 @@ ENDIF (NOT CLANG_URL)
 set(CMAKE_MACOSX_RPATH 1)
 
 IF (NOT NUM_BUILD_THREADS)
-  SET(NUM_BUILD_THREADS 8)
+  include(ProcessorCount)
+  ProcessorCount(NUM_BUILD_THREADS)
```
whchung: @pfultz2 Would it make sense if we separate this variable to 2:

- NUM_BUILD_THREADS sets the number of cores used for building HCC
- NUM_TEST_THREADS sets the number of cores used for testing HCC

To my empirical experience, setting NUM_BUILD_THREADS to number of available CPU cores "times 2" yields the best CPU utilization on my build cluster consisting of 3 x86 boxes.

pfultz2: Yea, it already is two separate variables. The `NUM_BUILD_THREADS` does not affect the threads for testing. Right now, I just called the variable `N`, but it does make sense to rename it to `NUM_TEST_THREADS`.

PR #128, opened 2016-09-22
Unpinned copy update2

bensander:
Fixes necessary for Peer staging buffers.

### DIFF
--- a/include/hc.hpp
+++ b/include/hc.hpp
@@ -39,6 +39,8 @@
  */
 namespace hc {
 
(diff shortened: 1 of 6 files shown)


whchung:
@bensander Is there a test case to depict how this PR improves things? Or is there a particular existing test case get fixed by this?

bensander:
Test cases are in HIP.  HCC P2P will need a pointer info feature in ROCR 1.3 – one we have that we can expand the HCC peer test cases including this one.

PR #130, opened 2016-09-22
[LIGHTNING-306] Fix amp-conformance tests, on which Clang 4.0 reasonably reports error for ill-formed const object declaration.

emankov:
[Problem]
Clang 4.0 reports an error "default initialization of an object of const type 'const XXX' without a user-provided default constructor" on some of const declarations without explicit initialization. 

[Explanation]
All the problematic declarations are of the objects inherited from POD classes, which are not initialized by default. 
C++ spec says: "If a program calls for the default-initialization of an object of a const-qualified type T, T shall be a class type with user-provided default constructor" ... 7.1.7.1 The cv-qualifiers ... "the definition of an object or subobject of const-qualified type must specify an initializer or be subject to default initialization."

[Conclusion]
The error reported by Clang 4.0 is valid and conforms to C++ spec, because the reported const declarations are ill-formed in fact. Clang 3.5 allows such ill-formed declarations and thus is wrong (Bug in the compiler). 

[Solution]
Add default empty ctor to the base POD classes in order to make them non-POD. Fix doesn't change tests' logic.

### DIFF
--- a/amp-conformance/Tests/2_Cxx_Lang_Exte/2_3_Expr_Invo_Rest_Func/2_3_2_Func_Over/Special_member_functions/defaulted_copy_assign_op.01/test.cpp
+++ b/amp-conformance/Tests/2_Cxx_Lang_Exte/2_3_Expr_Invo_Rest_Func/2_3_2_Func_Over/Special_member_functions/defaulted_copy_assign_op.01/test.cpp
@@ -78,6 +78,7 @@ struct A7 : A7_base
 // Empty class with base classes having both defaulted and user-defined copy op=
 struct A8_base_1
 {
+	A8_base_1() {}
 	int i;
 };
 class A8_base_2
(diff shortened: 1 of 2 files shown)


pfultz2:
> Add default empty ctor to the base POD classes in order to make them non-POD. 

This is not what this commit does. It adds an empty user-defined default ctor. This changes the test, fundamentally, and no longer tests for defaulted constructors. Instead, the solution should be to initialize the variables or remove cv-qualifiers where they are used, or even better both.

emankov:
Well, the commit actually does what is written in its description. It makes POD classes non-POD. There are a few ways to do it, you know, ctor, dtor or virtual function. As for the tests themselves, one of them tests defaulted copy assignment operators, second - defaulted copy constructors. None of them tests defaulted constructors. Thus I suggest that the tests' logic is not changed. Agree, that removing a const qualifier is yet another option, which eliminates the issue. But I'm not sure about initializing the variables. Could you please show how to initialize POD object? Thanks.

pfultz2:
> Thus I suggest that the tests' logic is not changed. 

I disagree that the logic of the test is not changed, because now it requires a user-defined constructor. 

> Could you please show how to initialize POD object?

```cpp
struct pod {};

int main() {
    const pod x = {};
}
```

emankov:
Thanks, Paul. I worried about initializing POD descendant objects, but it works too. I changed the fix in [133](https://github.com/RadeonOpenCompute/hcc/pull/133) pull request.

Issue #99, opened 2016-07-20
HC restriction attribute causes compilation failures


mcopik:
Adding [[hc]] attribute to template functions results in unexpected compilation errors, even in non-GPU HPX stuff. Those errors doesn't make sense most of the time, they can't be reproduced with any other compiler or hcc when neither [[hc]] nor [[hc, cpu]] is used.

One of the problems could be repeated with this simple sample:
https://gist.github.com/mcopik/6e1cf89c44df783b08a125b83df80b83
If std::forward_as_tuple is used, everything compiles fine. Our functions and tuple functionalities are defined with an additional macro HPX_HOST_DEVICE in front, defined here:
https://github.com/mcopik/hpx/blob/compute/hpx/config/compiler_specific.hpp#L85
If HPX_HOST_DEVICE is [[cpu]], this example compiles fine. If it is [[cpu, hc]], which is required for invoking algorithms on GPU (those are purely template calls!), we get unexpected error:
"rvalue reference to type 'int' cannot bind to lvalue of type 'int'"

To reproduce the example, please clone the repository with branch compute. It does not require building the project, only configuring build with cmake:
 -DHPX_WITH_HCC=(hcc location) -DHPX_WITH_HCC_HC=On
And then compilation of example requires additional include flags: -I$(HPX_SOURCE) -I$(HPX_BUILD)

mcopik:
I tried using HCC branch with updated clang (clang_tot_upgrade), but hcc doesn't build at all.

whchung:
@mcopik Sorry for not being able to address your issues. You surely uncovered a lot of them for HCC. I've achieved some important milestone in clang_tot_upgrade so my time there could be greatly reduced from now on.

clang_tot_upgrade is still in its early phase so I wouldn't recommend try your application on it yet.

mcopik:
Thanks @whchung 
If you need any help with debugging this issue, please let me know. I'll help as much as I can, but it can't be reproduced without HPX.

whchung:
I'm looking into this issue now. For some reason, std::move(lvalue) becomes a l-value instead of an x-value when [[hc]] is annotated in HPX_HOST_DEVICE macro. So the compiler thinks the whole initializer expression as an l-value and rejects it.

whchung:
I'm still working on this issue. Since it's related to type system is quite complicated. Let me jot down my findings so far:

[...]

PR #136, opened 2016-09-29
[LIGHTNING-308] Fix - Clang 4.0 reports linker error 'undefined reference to' ctor

emankov:
_[Code snippet]_

```
class restrict {
  int size;
  …
public:
  restrict(unsigned int _size) __GPU {
    this->size = _size;
  }
  …
private:
  int size;
};
…
restrict *y = new restrict(20); // error: undefined reference to `restrict::restrict(int)'
```
_[Analysis]_
The error is reported for conversion constructor only, neither for class virtual functions, nor for non-virtual functions. The restriction specifier is not the case. Constructor call for stack variable leads to the same error:
```
restrict y(20);
restrict y = 20;
restrict y = restrict(20);
```
But changing the type of member "size" from int to unsigned int eliminates the issue. It is very strange. Looks like a bug in Clang 4.0 compiler. MSVC doesn’t emit even a warning on the same code.

_[Conclusion]_
Anyway, either it is a Clang 4.0 itself bug or a correct behavior with conversion constructors (evidence in C++ Spec is not found), the variable size type in the code is unsigned semantically, and there is no need for test in implicit type casting from unsigned int to int and vice versa, as the "test checks that a function qualifier can be used as a class/struct name".

2 alternative options:
1) `restrict(int _size) __GPU`
2) `restrict *y = new restrict(20U);`

### DIFF
--- a/amp-conformance/Tests/2_Cxx_Lang_Exte/2_1_Synt/Contextual_restrict/test.05/test.cpp
+++ b/amp-conformance/Tests/2_Cxx_Lang_Exte/2_1_Synt/Contextual_restrict/test.05/test.cpp
@@ -14,7 +14,7 @@ using namespace std;
 
 class restrict
 {
-   int size;
+   unsigned int size;
 
 public:
     restrict() __GPU
--- a/amp-conformance/Tests/2_Cxx_Lang_Exte/2_1_Synt/Contextual_restrict/test.06/test.cpp
+++ b/amp-conformance/Tests/2_Cxx_Lang_Exte/2_1_Synt/Contextual_restrict/test.06/test.cpp
@@ -41,7 +41,7 @@ struct restrict
     }
 
 private:
-   int size;
+   unsigned int size;
 };
 
 


pfultz2:
This is a bug in the hcc clang frontend, and its very likely related to issue #99. When I compile this snippet with hcc(from develop), I get the 'undefined reference' error about conversion ctor:

```cpp
class restrict {
  int size;
  
public:
  restrict(unsigned int _size) [[hc]] {
    this->size = _size;
  }
  
};

int main() {
    restrict *y = new restrict(20);
}
```

Removing the `[[hc]]` attribute, the snippet compiles successfully with hcc, clang 4.0, and gcc. This is valid C++ code, and the tests shouldn't be update to compensate for bugs in the frontend.

whchung:
I agree with Paul, this issue seems to be a bug caused by HCC-specific logic, instead of a bug in the test case. We may need to look into how SMF gets generated in the frontend and whether they carry correct function attribute ([[hcc]]) in AST tree so overloading resolution logic could find them.

pfultz2:
BTW, the code snippet is not entirely correct. I forgot that `__GPU` means `[[hc]] [[cpu]]`. Changing it to use the attribute for both cpu and gpu:

```cpp
class restrict {
  int size;

public:
  restrict(unsigned int _size) [[hc]] [[cpu]] {
    this->size = _size;
  }

};

int main() {
    restrict *y = new restrict(20);
}
```

This code does compile with hcc from rocm 1.2. I don't if it fails on tot.(I assume so)

> We may need to look into how SMF gets generated

SMF? Static Member Function?

whchung:
Hi Paul, Special Member Function ( ctor / dtor / copy ctor / copy assignment / move ctor / move assignment ) it is.

emankov:
Paul, removing restrict specifiers from ctor declaration doesn't eliminate the error, anyway I'll double check it.

PR #149, opened 2016-10-25
[HCC Runtime] Performance tuning for dispatch speed.

weixingzhang:
The hotspot is the function DetermineAndGetProgram() which is called by function CreateKernel. The following changes are made.
- include/kalmar_runtime.h : change KalmarDevice::CreateKernel() to remove void\* size, void \* source from the function signature, so raw kernel bytes won't be used in parallel_for_each().
- lib/has/mcwamp_has.cpp : change HSADevice::CreateKernel() to honor the new interface, and change its logic as below:
  a. Don't call kernel_checksum() anymore, as the raw kernel bytes won't be passed to HSADevice::CreateKernel() anymore.
  b. Walks through the map of HSAExecutables to find the instance of HSAExecutable which has the symbol. If not found, create create HSAKernel instance.
  c. Create HSADispatch.

With the change above, the dispatch latencey measured on my system by app emptyKernelBench can be dramatically decreased from 144 micro-seconds to 10.6 micro-seconds.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -2059,13 +2059,25 @@ public:
         std::string str(fun);
         HSAKernel *kernel = programs[str];
         if (!kernel) {
-            size_t kernel_size = (size_t)((void *)size);
-            char *kernel_source = (char*)malloc(kernel_size+1);
-            memcpy(kernel_source, source, kernel_size);
-            kernel_source[kernel_size] = '\0';
-            //std::cerr << "HSADevice::CreateKernel(): Creating kernel: " << fun << "\n";
-            kernel = CreateOfflineFinalizedKernelImpl(kernel_source, kernel_size, fun);
-            free(kernel_source);
+            if (executables.size() != 0) {
+                for (auto executable_iterator : executables) {
+                    HSAExecutable *executable = executable_iterator.second;
+
+                    // Get symbol handle.
+                    hsa_status_t status;
+                    hsa_executable_symbol_t kernelSymbol;
+                    status = hsa_executable_get_symbol(executable->hsaExecutable, NULL, fun, agent, 0, &kernelSymbol);
+                    STATUS_CHECK(status, __LINE__);
+
+                    // Get code handle.
+                    uint64_t kernelCodeHandle;
+                    status = hsa_executable_symbol_get_info(kernelSymbol, HSA_EXECUTABLE_SYMBOL_INFO_KERNEL_OBJECT, &kernelCodeHandle);
+                    STATUS_CHECK(status, __LINE__);
+
+                    kernel =  new HSAKernel(executable, kernelSymbol, kernelCodeHandle);
+                }
+            }
+         
             if (!kernel) {
                 std::cerr << "HSADevice::CreateKernel(): Unable to create kernel\n";
                 abort();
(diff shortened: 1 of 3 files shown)


whchung:
@weixingzhang Please help provide:
- HCC unit test result: `make test`

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2059,13 +2059,25 @@ class HSADevice final : public KalmarDevice
         std::string str(fun);
```
whchung, line 2059: @weixingzhang it appears `HSADevice::CreateKernel()` doesn't really honor the revised interface in `kalmar_runtime.h`.

whchung (changes requested):
Please help revise `HSADevice::CreateKernel()` so it honors the new interface defined in `kalmar_runtime.h` . Also please provide the result of `make test`.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2059,13 +2059,25 @@ class HSADevice final : public KalmarDevice
...
+                    hsa_executable_symbol_t kernelSymbol;
+                    status = hsa_executable_get_symbol(executable->hsaExecutable, NULL, fun, agent, 0, &kernelSymbol);
+                    STATUS_CHECK(status, __LINE__);
+
+                    // Get code handle.
+                    uint64_t kernelCodeHandle;
+                    status = hsa_executable_symbol_get_info(kernelSymbol, HSA_EXECUTABLE_SYMBOL_INFO_KERNEL_OBJECT, &kernelCodeHandle);
+                    STATUS_CHECK(status, __LINE__);
+
+                    kernel =  new HSAKernel(executable, kernelSymbol, kernelCodeHandle);
```
whchung, line 2077: logic here doesn't seem correct to me. once an HSAKernel instance is created, shouldn't we exit the loop?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2059,13 +2059,25 @@ class HSADevice final : public KalmarDevice
...
-            free(kernel_source);
+            if (executables.size() != 0) {
+                for (auto executable_iterator : executables) {
+                    HSAExecutable *executable = executable_iterator.second;
+
+                    // Get symbol handle.
+                    hsa_status_t status;
+                    hsa_executable_symbol_t kernelSymbol;
+                    status = hsa_executable_get_symbol(executable->hsaExecutable, NULL, fun, agent, 0, &kernelSymbol);
+                    STATUS_CHECK(status, __LINE__);
```
whchung, line 2070: logic here doesn't seem correct to me. in case there are multiple HSAExecutable instances associated with one HSADevice, there might be cases where `hsa_executable_get_symbol()` may return error, but the for loop shall continue to find the the symbol for the next HSAExecutable until all HSAExecutable has been traversed.

whchung (changes requested):

whchung (changes requested):
@weixingzhang could you create additional test cases which:
- checks building HCC with multiple ISA (via `--amdgpu-target=` command line options) into a single executable
- checks building HCC with multiple ISA into several static libraries
- checks building HCC with multiple ISA into several shared libraries

you can consult unit tests in `tests/Unit/StaticLibrary` and `tests/Unit/SharedLibrary` to find how to create relevant unit tests.

scchan:
Perhaps let's pull this in on a separate branch instead of clang_tot_upgrade?

whchung:
@scchan For now we can leave this pull request open, and merge it once @weixingzhang updates the implementation, plus new tests & test results.

PR #159, opened 2016-11-10
[HCC Runtime] Check whether the number of SGPRs and VGPRs in a workgr…

weixingzhang:
…oup exceeds the max number that HW can support

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -26,6 +26,8 @@
 #include <hsa/hsa.h>
 #include <hsa/hsa_ext_finalize.h>
 #include <hsa/hsa_ext_amd.h>
+#include <hsa/amd_hsa_kernel_code.h>
+#include <hsa/hsa_ven_amd_loader.h>
 
 #include <hcc/kalmar_runtime.h>
 #include <hcc/kalmar_aligned_alloc.h>
@@ -313,6 +315,8 @@ private:
     hsa_executable_symbol_t hsaExecutableSymbol;
     uint32_t static_group_segment_size; 
     uint32_t private_segment_size;
+    uint16_t wavefront_sgpr_count;
+    uint16_t workitem_vgpr_count;
     friend class HSADispatch;
 
 public:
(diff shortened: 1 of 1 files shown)


whchung:
@weixingzhang could you briefly explain the purpose of this pull request? I'm not that familiar with API being used here.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -3171,6 +3197,11 @@ HSADispatch::setLaunchConfiguration(int dims, size_t *globalDims, size_t *localD
       workgroup_total_size = workgroup_size[0] * workgroup_size[1] * workgroup_size[2];
     }
 
+    if (kernel->wavefront_sgpr_count > 128 ||
+        kernel->workitem_vgpr_count*workgroup_total_size > 4*64*256) {
```
whchung, line 3201: @weixingzhang would these values change per ASIC?

scchan, line 3201: Could you comment on what "4 \* 64 \* 256" calculates?  Is it   4 (SIMD/CU) \* 64 (lane/SIMD) \* 256 (vgpr/lane) ?

I think we need to round-up the group size (if it hasn't been done in other places) to the next wave size because the HW schedule as the wavefront granularity.  

I think we need to calculate how many waves a workgroup would produce and see whether we have enough SIMDs to fit them in.    I believe there are cases where if we only look at the total # of gprs on a CU, it might be fine but the workgoup may still not fit in a CU because the remaining gprs are spread across different SIMDs

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -323,20 +327,42 @@ class HSAKernel {
...
+        bool ext_supported = false;
+        status = hsa_system_extension_supported(HSA_EXTENSION_AMD_LOADER, 1, 0, &ext_supported);
+        STATUS_CHECK_Q(status, commandQueue, __LINE__);
 
-            status = 
-                hsa_executable_symbol_get_info(
-                    _hsaExecutableSymbol,
-                    HSA_EXECUTABLE_SYMBOL_INFO_KERNEL_PRIVATE_SEGMENT_SIZE,
-                    &this->private_segment_size);
+        if (ext_supported) {
```
scchan, line 351: Maybe we'd want to strictly enforcing this since we know we always have this extension on rocm

PR #160, opened 2016-11-11
Fix p2p

bensander:


whchung:
Description for this pull request per @bensander 

```
I have done some major work on the HCC and HIP copy APis in light of what we learned from running P2P in CAFFE recently.  Basically, we need to separate the stream/queue (controls order of copy) from the optimal copy engine (depends on where the data is and is selected by the runtime).  Previously we picked the copy engine based on the user’s stream – turns out apps don’t really pick the engine very well and we had a host of issues with P2P as result.    HIP has the most information and knows which device should do the copy, and whether we need to force copy through staging buffer.  HCC now has interfaces so that HIP can directly convey this information, rather than the previous creaky solution where control logic was spread between HIP and HCC.  These are the new “copy_ext” and “copy_async_ext” interfaces.  These are on the hcc ‘fix_p2p’ branch – Jack please review my pull request.

The new HIP develop uses these new HCC APIs.  I have attempted to maintain backward compatibility with the compile define USE_COPY_EXT_V2.  The default is 0 until we get the required interfaces onto HCC develop branch.  (and clang40)    P2P may not work with USE_COPY_EXT_V2==0 but other copies should.  Let me know if you see issues.  With USE_COPY_EXT_V2=1 and the new HCC, all the HIP tests + 
```

test result:

```

********************
Testing Time: 379.16s
********************
Failing Tests (18):
    CPPAMP :: Unit/AsyncPFE/accelerator_view_wait.cpp
    CPPAMP :: Unit/AsyncPFE/async_av_dependent1.cpp
    CPPAMP :: Unit/AsyncPFE/async_av_dependent2.cpp
    CPPAMP :: Unit/AsyncPFE/async_av_dependent4.cpp
    CPPAMP :: Unit/AsyncPFE/async_av_dependent5.cpp
    CPPAMP :: Unit/AsyncPFE/async_av_dependent8.cpp
    CPPAMP :: Unit/AsyncPFE/async_av_independent1.cpp
    CPPAMP :: Unit/AsyncPFE/async_av_independent4.cpp
    CPPAMP :: Unit/AsyncPFE/completion_future_wait.cpp
    CPPAMP :: Unit/DynamicTileStatic/test11.cpp
    CPPAMP :: Unit/HC/create_marker.cpp
    CPPAMP :: Unit/HC/memcpy_symbol2.cpp
    CPPAMP :: Unit/HC/memcpy_symbol4.cpp
    CPPAMP :: Unit/HC/test2.cpp
    CPPAMP :: Unit/Pool/accelerator_get_is_peer_api.cpp
    CPPAMP :: Unit/Pool/accelerator_view_set_cu_mask.cpp

  Expected Passes    : 682
  Expected Failures  : 22
  Unsupported Tests  : 10
  Unexpected Failures: 18




Also passes HIP directed tests, hipSimpleP2p, and p2pbandwidthlatency
```

include/hc.hpp
```diff
@@ -316,6 +318,33 @@ class accelerator_view {
...
     completion_future copy_async(const void *src, void *dst, size_t size_bytes);
 
+
+    /**
+     * Copies size_bytes bytes from src to dst.  
+     * Src and dst must not overlap.  
+     * Note the src is the first parameter and dst is second, following C++ convention.  
+     * This is an asynchronous copy command, and this call may return before the copy operation completes.
+     *
+     * The copy command will be implicitly ordered with respect to commands previously enqueued to this accelerator_view:
```
scchan, line 328: This API takes a pointer of an hc::accelerator rather than a pointer of an hc::accelerator_view so how would the copy command or the user know which accelerator_view the copy command is synchronized with?  Unless the copy command is synchronized with all accelerator_views associated with that specific accelerator?

scchan: ah.. . ok it's member of the accelerator_view so please ignore my previous comment

bensander:
Right – do we need to update the docs to make this more clear?  The copy_async_ext has a better description:

The copy command will be implicitly ordered with respect to commands previously enqueued to this accelerator_view:
     \* - If the queue execute_order is execute_in_order (the default), then the copy will execute after all previously sent commands finish execution.
     \* - If the queue execute_order is execute_any_order, then the copy will start after all previously send commands start but can execute in any order.
     \*   The copyAcc determines where the copy is executed and does not affect the ordering.
     *

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2995,6 +2991,121 @@ HSAQueue::getHSAKernargRegion() override {
...
+    // create a HSACopy instance
+    HSACopy* copyCommand = new HSACopy(src, dst, size_bytes);
+
+    // synchronously do copy
+    // FIX me, pull from constructor.
+    copyCommand->syncCopyExt(this, copyDir, srcPtrInfo, dstPtrInfo, copyDeviceHsa, forceUnpinnedCopy);
+
+    // TODO - should remove from queue instead?
+    delete(copyCommand);
+
```
scchan, line 3014: Suggest to change from using new/delete to automatic variable.  This will prevent memory leak when syncCopyExt throws an exception  and copyCommand will still be deallocated properly when we unwind the stack:

   HSACopy copyCommand(src, dst, size_bytes);

```
// synchronously do copy
// FIX me, pull from constructor.
copyCommand.syncCopyExt(this, copyDir, srcPtrInfo, dstPtrInfo, copyDeviceHsa, forceUnpinnedCopy);
```

bensander:
Good idea.  I think this existed before this pull request as well.

PR #163, opened 2016-11-15
[HCC Runtime] Check whether the number of SGPRs and VGPRs in a workgr…

weixingzhang:
…oup exceeds the max number that HW can support

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -26,6 +26,8 @@
 #include <hsa/hsa.h>
 #include <hsa/hsa_ext_finalize.h>
 #include <hsa/hsa_ext_amd.h>
+#include <hsa/amd_hsa_kernel_code.h>
+#include <hsa/hsa_ven_amd_loader.h>
 
 #include <hcc/kalmar_runtime.h>
 #include <hcc/kalmar_aligned_alloc.h>
@@ -313,6 +315,8 @@ private:
     hsa_executable_symbol_t hsaExecutableSymbol;
     uint32_t static_group_segment_size; 
     uint32_t private_segment_size;
+    uint16_t wavefront_sgpr_count;
+    uint16_t workitem_vgpr_count;
     friend class HSADispatch;
 
 public:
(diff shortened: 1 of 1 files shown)


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -3171,6 +3191,21 @@ HSADispatch::setLaunchConfiguration(int dims, size_t *globalDims, size_t *localD
...
     }
 
+    static const size_t num_work_items_per_simd = 64;
+    static const size_t num_simds_per_cu = 4;
+    static const size_t num_work_items_per_cu = num_work_items_per_simd * num_simds_per_cu;
+    size_t num = (workgroup_total_size + num_work_items_per_cu - 1) / num_work_items_per_cu;
+    size_t num_sgprs = num * kernel->wavefront_sgpr_count;
+    size_t num_vgprs = num * kernel->workitem_vgpr_count;
+    // Every wavefront is allocated up to a maximum of 104 SGPRs on the GPUs prior to VI and 102 SGPRs on VI.
+    if (num_sgprs > 102) {
```
whchung, line 3201: @weixingzhang  could you revise this hard-coded value as a macro? Also it would be better if we use ISA version (ex: GFX8), instead of ASIC family names (ex: VI).

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -3171,6 +3191,21 @@ HSADispatch::setLaunchConfiguration(int dims, size_t *globalDims, size_t *localD
...
+        throw Kalmar::runtime_exception("The number of SGPRs needed by this launch exceeds HW limit due to big work group size!", 0);
+    }
+    // Every work-item has access to some number of VGPRs, up to a maximum of 256.
+    if (num_vgprs > 256) {
```
whchung, line 3205: @weixingzhang could you revise this value as a macro, with comment to describe which ISA version does this value apply to (ex: GFX7 above , or only for GFX8).

whchung:
@weixingzhang is this pull request superseded by #167 ?

weixingzhang:
@whchung Yes, this pull request is superseded by #167.

PR #165, opened 2016-11-16
[benchEmptyKernel] Remove the dependency of HIP and generate code obj…

weixingzhang:
…ect file during compiling the test.


whchung:
@weixingzhang is this pull request superseded by #168 ?

weixingzhang:
@whchung Yes.

PR #169, opened 2016-11-18
[HCC Tests] Adding mulitple ISA test cases

weixingzhang:
- multiple ISA (via --amdgpu-target= command line options) into a single executable
    - multiple ISA into static library
    - multiple ISA into shared library


tests/Unit/HC/multi_acc.cpp
```diff
@@ -1,5 +1,5 @@
 
-// RUN: %hc %s -o %t.out && %t.out
```
scchan, line 2: I thought we want to make a new copy of this and then add the multiple isa options instead of modify the existing test?

tests/Unit/HC/multi_acc_array.cpp
```diff
@@ -1,5 +1,5 @@
 
-// RUN: %hc %s -o %t.out && %t.out
```
scchan, line 2: See my previous comment...

whchung:
@weixingzhang I concur what @scchan said in the review. Please create new tests instead of modifying existing ones.

weixingzhang:
Sure, I just revised my change and submitted a new pull request #171.

PR #181, opened 2016-12-02
Revert: Revert "[modules] PR28752: Do not instantiate variable declarations which are not visible."

AlexVlx:
This re-enables the upstream patch, after fixing the issues it was spawning:
    - upstream did not include our C++ AMP / HC specific bits - this was corrected via merging;
    - the new behaviour brought to the fore some dubious behaviour in the conformance suite we inherited - this was corrected by small edits in some of the helper functions; ideally, we will end up fully refactoring the conformance suite.
Thank you.


whchung:
@AlexVlx if I understand correctly, this PR is to be applied after applying PR#9 in hcc-clang-upgrade, right?

AlexVlx:
Hello Jack.

Yup, that is correct, sorry for not clarifying!

PR #184, opened 2016-12-07
Added version header file

aditya4d:
Useful to implement compiler specific features in HIP.
Example, half types, rocm-device-libs etc.,


whchung:
would clang-builtin macros such as `__clang_major__`, `__clang_minor__` be sufficient? please refer to: https://github.com/RadeonOpenCompute/hcc-clang-upgrade/blob/master/lib/Frontend/InitPreprocessor.cpp#L548

aditya4d:
I didn't know that. Thanks!

PR #185, opened 2016-12-07
Added version header

aditya4d:
Useful for implementing compiler specific features in HIP


whchung:
would clang-builtin macros such as __clang_major__, __clang_minor__ be sufficient? please refer to: https://github.com/RadeonOpenCompute/hcc-clang/blob/master/lib/Frontend/InitPreprocessor.cpp#L457

PR #190, opened 2016-12-14
Added CodeXL Activity Logger header file and library to the CMake files

ot-amd:
Allows CodeXL Activity Logger to be used in HCC.
To enable set the environment variable USE_CODEXL_ACTIVITY_LOGGER to 1 when building HCC

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,3 +139,20 @@ dpkg -i libc++-dev_3.9.0-3_amd64.deb
 ```
 
 This replaces previous version of libc++.
+
+CodeXL Activity Logger
+======================
+To enable the [CodeXL Activity Logger](https://github.com/RadeonOpenCompute/ROCm-Profiler/tree/master/CXLActivityLogger), use `USE_CODEXL_ACTIVITY_LOGGER` env var.
+
+Configure the build like: 
+
+```
+cmake \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DHSA_AMDGPU_GPU_TARGET=<AMD GPU ISA version string> \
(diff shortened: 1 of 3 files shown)


whchung:
@otauqir could you provide documentation about how to use this feature? For example, you can change `README.md` to add a section on how to build HCC with CodeXL profiling support. You might need to provide information about:

- supported CodeXL version. where to download/install it.
- how to configure HCC with CodeXL.
- how to profile HCC applications with CodeXL.
- how to see profiling result within CodeXL GUI.

whchung:
@otauqir thanks for providing the documentation. could you revise it a bit so readers could better understand the rationale of this functionality. for example, when would you need to consider use this feature. how this feature could help you understand the behavior of your application.

scchan:
These are just cmake changes to build the runtime with the codexl library linked in.  No profiling instrumentation has been added to the runtime yet

whchung:
Since this PR seems to introduce a new feature, so I want to understand better before adopting the proposed change, especially from user perspective.

Based on the discussion so far, could you help clarify:

a) right now this PR won't bring any visible effect from user perspective?

b) future PR(s) may change HCC runtime so users may have better profiling support? and what would those improvements be, given CodeXL can already profile ROCR runtime API calls / kernel invocations?

ot-amd:
This change allows the Activity Logger to be used to profile the hcc runtime itself not a user application. It links the Activity Logger library and header to the generated libraries of hcc, mcwamp_hsa.so and hc_am.so

whchung (approved):

PR #193, opened 2016-12-15
Fixes for some unit tests, coupled with slight refactoring.

AlexVlx:
This cleans up some of the unit tests, getting us down to only 4 failing tests, out of which at least one (list2) appears to be a legitimate compiler bug. Most changes are trivial, primarily dealing with disambiguating our usage of math functions. hc_math.cpp gets some more surgery, being refactored into a less macro-ised, slightly more modern form. Overall, this is an intermediate step, which can and should be followed by proper enhancements, as detailed by some of the TODOs.


whchung (approved):
Thanks for submitting this PR. They look good.

PR #198, opened 2016-12-20
build hcc-config with g++

scchan:
- build hcc-config with g++ instead of hcc's clang
- replace clamp-config with a symlink to hcc-config
- renamed hcc-config's source file name and moved to a new directory
- fix a few other symlinks


CMakeLists.txt
```diff
@@ -358,7 +359,7 @@ if ("${DISTRO}" STREQUAL "ubuntu")
   set(CPACK_DEBIAN_PACKAGE_CONTROL_EXTRA
     "${PROJECT_BINARY_DIR}/packaging/debian/postinst;${PROJECT_BINARY_DIR}/packaging/debian/prerm")
 
-  set(HCC_GENERAL_DEBIAN_DEP "libstdc++-4.8-dev, libc++1, libc++-dev, libc++abi1, libc++abi-dev, elfutils")
+  #set(HCC_GENERAL_DEBIAN_DEP "libstdc++-dev, libc++1, libc++-dev, libc++abi1, libc++abi-dev, elfutils")
```
whchung, line 362: @scchan could you clarify why this line is commented out now?

hcc_config/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+add_custom_command(TARGET hcc-config POST_BUILD
+  COMMAND ${CMAKE_COMMAND} -E create_symlink ./hcc-config ./clamp-config
+  WORKING_DIRECTORY ${PROJECT_BINARY_DIR}/bin
+)
+
+install(TARGETS hcc-config
+    RUNTIME DESTINATION bin
+)
+
+install(FILES ${PROJECT_BINARY_DIR}/bin/clamp-config DESTINATION bin)
```
whchung, line 23: @scchan could you comment if `/opt/rocm/hcc/bin/clamp-config` would be correctly linked to `/opt/rocm/hcc/bin/hcc-config`, not the one in the build directory?

whchung (commented):
Based on the discussion with @scchan , this PR changes "hcc-config" & "clamp-config" so they can be built without dependency to libc++. The dependency of HCC runtime implementations are beyond the scope of this PR.

CMakeLists.txt
scchan: because it would trigger apt-get to try to install these packages when testing on Ubuntu 16.04

hcc_config/CMakeLists.txt
scchan: Yes because the symlink is relative path, here's the content of the install dir created by the debian package

scchan@d7e2ff423cf3:/opt/rocm/hcc-1.0/bin$ ls -la
total 36
drwxrwxr-x 2 root root  4096 Dec 20 06:36 .
drwxrwxr-x 6 root root  4096 Dec 20 06:36 ..
lrwxrwxrwx 1 root root    12 Dec 20 05:56 clamp-config -> ./hcc-config
lrwxrwxrwx 1 root root    21 Dec 20 05:56 clang -> ../compiler/bin/clang
lrwxrwxrwx 1 root root    23 Dec 20 05:56 clang++ -> ../compiler/bin/clang++
-rwxr-xr-x 1 root root  5309 Dec 16 16:27 extractkernel
lrwxrwxrwx 1 root root    19 Dec 20 05:56 hcc -> ../compiler/bin/hcc
-rwxr-xr-x 1 root root 20136 Dec 20 05:55 hcc-config
lrwxrwxrwx 1 root root    19 Dec 20 05:56 lld -> ../compiler/bin/lld
lrwxrwxrwx 1 root root    23 Dec 20 05:56 llvm-mc -> ../compiler/bin/llvm-mc
lrwxrwxrwx 1 root root    28 Dec 20 05:56 llvm-objdump -> ../compiler/bin/llvm-objdump

whchung (approved):

PR #200, opened 2016-12-22
Modified hcc-config so that applications compiled with hcc that has CodeXL enabled can use the Activity Logger

ot-amd:
Applications compiled with a build of hcc that has the CodeXL Activity Logger enabled using the `USE_CODEXL_ACTIVITY_LOGGER` variable, can use the Activity Logger in the application by adding the header file.


whchung (approved):
Thanks for submitting this PR. Looks good to me.

PR #206, opened 2017-01-09
Export cmake targets so they can be used by other libraries

pfultz2:
With this, users should be able to write:

```cmake
find_package(hcc)
add_library(myLibrary ${SOURCES})
target_link_libraries(myLibrary ${hcc_LIBRARIES})
```

And they will be able to use the hcc runtime.


pfultz2:
Is there any feedback on this pull request?

scripts/cmake/MCWAMP.cmake
```diff
@@ -20,7 +20,10 @@ set(MCWAMP_INC_DIR "${PROJECT_SOURCE_DIR}/include")
 
 macro(amp_target name )
 	target_compile_definitions(${name} PRIVATE "GTEST_HAS_TR1_TUPLE=0")
-	target_include_directories(${name} PRIVATE ${GTEST_INC_DIR} ${LIBCXX_INC_DIR} ${MCWAMP_INC_DIR})
+  target_include_directories(${name} SYSTEM PRIVATE ${GTEST_INC_DIR} ${LIBCXX_INC_DIR})
+	target_include_directories(${name} PRIVATE ${MCWAMP_INC_DIR})
+  target_include_directories(${name} SYSTEM INTERFACE $<INSTALL_INTERFACE:$<INSTALL_PREFIX>/include>)
+  target_include_directories(${name} SYSTEM INTERFACE /usr/include/libcxxabi)
```
whchung: @pfultz2 may I understand the reason behind these additional lines?

scripts/cmake/MCWAMP.cmake
```diff
@@ -55,7 +58,7 @@ macro(add_mcwamp_library_hsa name )
   CMAKE_FORCE_CXX_COMPILER("${PROJECT_BINARY_DIR}/compiler/bin/clang++" MCWAMPCC)
   # add HSA headers
   add_library( ${name} SHARED ${ARGN} )
-  target_include_directories(${name} PRIVATE ${HSA_HEADER})
+  target_include_directories(${name} SYSTEM PUBLIC ${HSA_HEADER})
```
whchung, line 60: @pfultz2 may I understand the reason behind going from "PRIVATE" to "SYSTEM PUBLIC"?

whchung:
@pfultz2 I've checked this PR, could you help clarify the questions I mentioned in the review before I merge it? Thanks.

pfultz2:
I don't see any questions nor any reviews?

scchan:
@pfultz2  Could you contribute a unit test for this?  It doesn't have to be part of "make test" since it would probably be a post-install test

scripts/cmake/MCWAMP.cmake
pfultz2: The first two lines make the gtest and libcxx includes(which I don't think this variable is set) system include, however the mcwamp include is not since its part of the project. 

The next line adds include to the directory where it will be installed(but only when its installed and not built). This is so consumers of the target can include the header files. 

The last line adds an include for libc++abi, I recall this was needed at one point. Perhaps not, I will drop it as I don't see any libc++abi header included.

scripts/cmake/MCWAMP.cmake
pfultz2: Because `hsa_atomic.h` is used be `hc.hpp`: https://github.com/RadeonOpenCompute/hcc/blob/clang_tot_upgrade/include/hc.hpp#L24

So the include needs to be public, so consumers of the target gets the correct includes as well.

pfultz2:
> @pfultz2 Could you contribute a unit test for this? It doesn't have to be part of "make test" since it would probably be a post-install test

I am not sure how to setup a test to check after installation, but I can at least check that doing `target_link_libraries(simpleExample mcwamp_hsa hc_am)` will work correctly. Where should I put a test like that?

pfultz2:
Ok, so I added some test in a new directory called `cmake-test`. You can move it to another directory if you want. I also added linking against libdl for mcwamp because that is missing.

whchung (approved):

PR #207, opened 2017-01-11
Using libstdc++ as C++ runtime

scchan:
This is a patch that allows hcc to use libstdc++ on a distro with g++ version > 5.0.  It also contains various fixes in the hcc RT and tests exposed by libstdc++.

This is currently OFF by default and could be enabled by running cmake with  -DUSE_LIBCXX=ON.
It has been tested on Ubuntu 16.04 and on Fedora 24 with libstdc++

It has been tested on Ubuntu 14.04 with libc++ to ensure no regression has been introduced.


amp-conformance/CMakeLists.txt
```diff
@@ -1,4 +1,7 @@
-include_directories(${CMAKE_CURRENT_SOURCE_DIR}/amp_test_lib/inc ${LIBCXX_HEADER})
+include_directories(${CMAKE_CURRENT_SOURCE_DIR}/amp_test_lib/inc)
+if (USE_LIBCXX)
+  include_directories(${LIBCXX_HEADER})
```
pfultz2, line 3: If this include is necessary to use the amp libraries, then the include should be added to the libraries' target, and not here. Otherwise, the user will need to add this line every time they link against the libraries.

lib/CMakeLists.txt
```diff
@@ -6,7 +6,11 @@ if (HCC_RUNTIME_DEBUG)
   add_definitions(-DKALMAR_DEBUG=1)
 endif (HCC_RUNTIME_DEBUG)
 
-include_directories(${CMAKE_CURRENT_BINARY_DIR} ${LIBCXX_HEADER} )
+include_directories(${CMAKE_CURRENT_BINARY_DIR})
+
+if (USE_LIBCXX)
+  include_directories(${LIBCXX_HEADER})
```
pfultz2: This include should also be added to each target, instead of at directory scope.

amp-conformance/Tests/4_Basic_Data_Elmnts/4_2_extent/4_2_3_m/tile/Negative/pad.01/test.cpp
```diff
@@ -11,6 +11,7 @@
 #include <amptest.h>
 #include <amptest_main.h>
 #include <string>
+#include <climits>
```
whchung, line 14: @scchan may I understand these changes to C++AMP conformance tests are to make them be properly built with libstdc++?

include/hc_math.hpp
```diff
@@ -71,6 +71,8 @@ function(Q arg1) __attribute__((hc,cpu)) { \
   return hc::precise_math::function(arg1); \
 }
 
+#define HC_MATH_WRAPPER_FP_OVERLOAD_TQ_STD(function, T, arg1)  HC_MATH_WRAPPER_FP_OVERLOAD_TQ(function, T, arg1)
+
```
whchung, line 75: @scchan may I understand if changes in this header are meant to make the file built with libstdc++?

include/kalmar_math.h
```diff
@@ -502,7 +502,7 @@ namespace fast_math {
     #endif
   }
 
-  inline int host_isfinite(float x) restrict(cpu) { return ::isfinite(x); }
+  inline int host_isfinite(float x) restrict(cpu) { return std::isfinite(x); }
```
whchung, line 505: @scchan same as above, could you confirm if changes here (::xxx() -> std::xxx() ) are to make sure the header can be built with libstdc++?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -339,25 +339,25 @@ class HSAKernel {
                 _hsaExecutableSymbol,
                 HSA_EXECUTABLE_SYMBOL_INFO_KERNEL_GROUP_SEGMENT_SIZE,
                 &this->static_group_segment_size);
-        STATUS_CHECK_Q(status, commandQueue, __LINE__);
+        STATUS_CHECK(status, __LINE__);
```
whchung, line 342: @scchan could you clarify why we need to change macro being used here?

scripts/cmake/MCWAMP.cmake
```diff
@@ -78,29 +94,13 @@ macro(add_mcwamp_library_hc_am name )
...
 
-####################
-# C++AMP config (clamp-config)
-####################
+  if (USE_LIBCXX)
+    target_link_libraries(${name} c++)
+    target_link_libraries(${name} c++abi)
+  endif (USE_LIBCXX)
 
-macro(add_mcwamp_executable name )
```
whchung, line 91: @scchan `add_mcwamp_executable()` macro is removed in this PR. IIRC this macro was used to build `hcc-config` and `clamp-config`. could you clarify this macro is no longer needed?

tests/Unit/HC/amdgcn_ds_bpermute.cpp
```diff
@@ -77,7 +78,7 @@ int main() {
 
   pass &= run_test<unsigned int>(1024*1024);
   pass &= run_test<int>(1024*1024);
-  pass &= run_test<float>(1024*1024);
+  //pass &= run_test<float>(1024*1024);
```
whchung: @scchan it seems this test is disabled. could you confirm if it doesn't work with libstdc++ anymore?

tests/Unit/HC/amdgcn_ds_bpermute.cpp
```diff
@@ -35,7 +35,8 @@ bool run_test(const int num) {
   T max = std::is_signed<T>::value ? INT_MAX : UINT_MAX;
   T min = std::is_signed<T>::value ? INT_MIN : 0;
 
-  std::uniform_int_distribution<T> distribution(min, max);
+  typedef typename std::conditional<std::is_signed<T>::value, int, unsigned int>::type _RangeType;
+  std::uniform_int_distribution<_RangeType> distribution(static_cast<_RangeType>(min), static_cast<_RangeType>(max));
```
whchung, line 39: @scchan is this change to make libstdc++ happy?

tests/Unit/HC/reduction_hc.cpp
```diff
@@ -1,594 +1,595 @@
```
whchung, line 1: @scchan could you clarify what was changed in this test case? it seems the whole file was changed but I couldn't spot differences between the old version and the new one.

tests/Unit/SharedLibrary/shared_library2.cpp
```diff
@@ -1,7 +1,7 @@
 
 // RUN: %hc -fPIC -Wl,-Bsymbolic -shared -DSHARED_LIBRARY_1 %s -o %T/libtest2_foo.so
 // RUN: %hc -fPIC -Wl,-Bsymbolic -shared -DSHARED_LIBRARY_2 %s -o %T/libtest2_bar.so
-// RUN: %clang -std=c++11 -stdlib=libc++ -lc++ -lc++abi -ldl -lpthread %s -o %t.out -ldl && LD_LIBRARY_PATH=%T %t.out
+// RUN: %clang -std=c++11 -ldl -lpthread %s -o %t.out -ldl && LD_LIBRARY_PATH=%T %t.out
```
whchung, line 4: @scchan would this test case still work in case "USE_LIBCXX" is ON?

tests/lit.cfg
```diff
@@ -70,7 +70,6 @@ def inferClang(PATH):
 cxx_options = ' ' + ' '.join([
   "-I%s" % config.gtest_src_dir,
   "-DGTEST_HAS_TR1_TUPLE=0",
-  "-stdlib=libc++",
```
whchung, line 73: @scchan would HCC unit tests still work if "USE_LIBCXX" is ON?

amp-conformance/CMakeLists.txt
whchung: If I understand the intention of this change correctly, this change only affects libraries of C++AMP conformance suite. They will not be shipped with HCC binary packages. So I think it shall be fine.

@scchan could you confirm if `test_one.pl` and `run_tests.pl` still function properly on both cases (USE_LIBCXX = ON / OFF)?

lib/CMakeLists.txt
whchung: @pfultz2 I noticed your PR #206 also heavily changes this file. could you help advise how it shall be changed? My rough understanding is that every `install(TARGETS foo EXPORT hcc-targets ...)` may need to be changed, so `hcc-targets.cmake` generated would properly contain paths to libc++ headers?

amp-conformance/CMakeLists.txt
scchan: yes, C++AMP conformance was run with OFF on ubuntu 14.04 and with ON on ubuntu 16.06.  Issues exposed by libstdc++ in those tests were also fixed

amp-conformance/Tests/4_Basic_Data_Elmnts/4_2_extent/4_2_3_m/tile/Negative/pad.01/test.cpp
scchan: Yes, these issues are exposed by the libstdc++ headers due to variations in the implementations and also they are less tolerant to these small bugs in the tests

include/hc_math.hpp
scchan: Some of the math functions are only defined within the std namespace in libstdc++

include/kalmar_math.h
scchan: yes

lib/hsa/mcwamp_hsa.cpp
scchan: This is a bug.  commandQueue is not defined in HSAKernel.  Somehow it was only exposed on Fedora

scripts/cmake/MCWAMP.cmake
scchan: This is not needed anymore. hcc-config has its own cmakefile hcc_config/CMakeLists.txt that doesn't depend on this macro

tests/Unit/HC/amdgcn_ds_bpermute.cpp
scchan: libstdc++'s uniform_int_distribution enforces the input template type to be integers.  This change is to allow the unit test template to work with floats

tests/Unit/HC/amdgcn_ds_bpermute.cpp
scchan: was turned off to debug the uniform_int_distribution issue with floats.  It works now with the change above.  I'll submit a commit to re-enable it.

tests/lit.cfg
scchan: yes, this branch has been tested on Ubuntu 14.04 with USE_LIBCXX=ON.   -stdlib=libc++ is provided by the hcc-config tool

tests/Unit/SharedLibrary/shared_library2.cpp
scchan: yes it works

tests/Unit/HC/amdgcn_ds_bpermute.cpp
scchan: amendment submitted

tests/Unit/HC/reduction_hc.cpp
scchan: it just cleans up the carriage return ^M character after each /n probably introduced by an editor on Windows

amp-conformance/CMakeLists.txt
pfultz2: Ok, I guess the tests don't link against the amp runtime in cmake, but its done by the compiler.

lib/CMakeLists.txt
pfultz2: You should use `target_include_directories` for each target that needs libc++. This way when I say `target_link_libraries(myLib hcc::mcwamp_hsa hcc::hc_am)` it will add the includes for libc++ if necessary.

whchung: @scchan I've merged PR #206 from @pfultz2 , could you rebase and modify this PR according to what Paul recommends?

whchung: @scchan I noticed you just rebased. For this `include_directories` would you consider change it to `target_include_directories` per @pfultz2 's suggestion?

whchung:
@scchan are we ready to merge this PR? I believe you are well aware of @pfultz2 's changes.

scchan:
@whchung this PR is ready now

whchung (approved):
LGTM

PR #208, opened 2017-01-17
update llvm to get the static grid launch kernel fix, add 2 tests

scchan:
-update llvm to get the static grid launch kernel fix
-add 2 tests for static grid launch kernels


whchung (approved):
LGTM

PR #209, opened 2017-01-17
Enable passing generic lambdas as arguments to p_f_e

AlexVlx:
This handles the AMP/HC runtime part of enabling passing generic lambdas as arguments to p_f_e. It is matched with an associated PR against the Clang upgrade repository, since changes are needed in Clang as well. The enabled functionality is predicated on also enabling C++14+ support, which we've yet to do. The main change is to do manual overload resolution against the correct index type on the kernel path within the p_f_e against the Callable's operator(), since in a generic lambda scenario this is no longer singular, but rather a function template that stands in for a family of functions. It also fixes some tests which were ill-specified and adds unit tests for verifying functionality both in C++AMP mode and in HC mode. The latter are only enabled when running in in no less than C++14 mode. It temporarily bubbles up some Promote pass errors in the functor4, functor5 and functor6 unit tests, which will be subsequently fixed, and is otherwise unit test neutral. While I believe that a cleaner design is possible and desirable in the Runtime, that would be too big a leap at the moment. Please review and merge if adequate. Thank you.

### DIFF
--- a/include/amp.h
+++ b/include/amp.h
@@ -21,6 +21,8 @@
 #include "kalmar_launch.h"
 #include "kalmar_cpu_launch.h"
 
+#include <utility> // For std::declval.
+
 // forward declaration
 namespace Concurrency {
 class completion_future;
@@ -95,7 +97,7 @@ public:
         pQueue = other.pQueue;
         return *this;
     }
-  
+
     /**
      * Returns the queuing mode that this accelerator_view was created with.
      * See "Queuing Mode".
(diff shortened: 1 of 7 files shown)


whchung:
@AlexVlx it's about time to give this PR a try. could you help resolve the conflicts?

AlexVlx:
@whchung let me rebase on current and issue a new PR. Thank you for taking the time to look.

AlexVlx:
@whchung please see #251. Thank you.

PR #210, opened 2017-01-17
Updates to export targets

pfultz2:
This adds a `hccrt` target which has the proper linking for using hcc runtime. This also adds a test for installation and using `find_package`.


lib/CMakeLists.txt
```diff
@@ -19,7 +19,13 @@ set( CONFIG_PACKAGE_INSTALL_DIR ${LIB_INSTALL_DIR}/cmake/hcc )
 add_mcwamp_library(mcwamp mcwamp.cpp)
 add_mcwamp_library(mcwamp_atomic mcwamp_atomic.cpp)
 
-target_link_libraries(mcwamp dl)
+# Library interface to use runtime
+add_library(hccrt INTERFACE)
+target_link_libraries(hccrt INTERFACE "-Wl,-rpath $<TARGET_FILE_DIR:$<TARGET_NAME:mcwamp>>" "-Wl,--whole-archive" mcwamp "-Wl,--no-whole-archive")
+target_link_libraries(hccrt INTERFACE dl)
+target_link_libraries(hccrt INTERFACE c++)
+target_link_libraries(hccrt INTERFACE c++abi)
```
whchung, line 28: @scchan could you comment on how these lines be changed if we enable libstdc++ as the default C++ runtime for HCC runtime? would it be just an `if(USE_LIBCXX)` check?

whchung (commented):
This PR generally looks good to me. Adding @scchan to take a look so he is aware of potential changes needed when we introduce libstdc++ support.

cmake-tests/test-install.sh
```diff
@@ -0,0 +1,29 @@
...
+
+INSTALL_DIR=$TMP_DIR/hcc
+
+mkdir $TMP_DIR/hcc-build && cd $TMP_DIR/hcc-build
+
+cmake $SOURCE_DIR -DCMAKE_INSTALL_PREFIX=$INSTALL_DIR $@
+make -j`nproc` world
+make -j`nproc`
+make install
+
```
scchan: This should be a post-installation test, so I think we should skip the installation and assume hcc has already been installed

lib/CMakeLists.txt
scchan: Yes, just guard c++ and c++abi with USE_LIBCXX

cmake-tests/test-install.sh
scchan: Except the requested change above in test-install.sh, other stuffs LGTM

pfultz2: The test needs to know where hcc was installed.

scchan: Let's say I want to add find_package(hcc) to a project's cmakefile.  I have installed the deb hcc package on the system, then how would the project know where hcc was installed?

pfultz2: Since we install in a non-standard path, the library will need to append the `CMAKE_PREFIX_PATH` with `/opt/rocm` since that is a common location for the runtime. That still won't prevent the user from setting `CMAKE_PREFIX_PATH` themselves when they have installed the runtime in an alternative location.

scchan: Execution of this test should be decoupled from the compilation/installation of hcc.  This test should be able to run in an environment where an hcc compiler has already been installed to a known location without having access to hcc build tree.
The most common use case would be a project using the hcc compiler that has been installed to /opt/rocm/ (or any other known alternative location).  The project needs to be installed to a different directory other than /opt/rocm.  We want a standalone test case that has a find_package(hcc), is able to find the hcc-config from the existing hcc installation and then gets compiled successfully.

pfultz2: I am not sure what you are asking here. I updated the cmake-test to append `/opt/rocm` to the prefix path. So now you can test it post installation by doing something like:

```
mkdir build
cd build
CXX=/opt/rocm/bin/hcc cmake ..
make
```

pfultz2:
Ok, so I removed the test-install script. Now, you should be able to test it with cmake with `CXX=/opt/rocm/bin/hcc cmake ..`, although the symlinks may need to be updated.

scchan:
reviewed, thanks

PR #215, opened 2017-01-20
Feature hcc opt flush2

bensander:
After cleaning up indentation nightmare.
Passes HIP tests as well, including with HCC_MAX_QUEUES=2.

Default results from HCC tests below:

********************
Testing Time: 702.16s
********************
Unexpected Passing Tests (4):
    HCC :: Unit/FilePath/file path_test2.cpp
    HCC :: Unit/FilePath/file_path_test1.cpp
    HCC :: Unit/FilePath/file_path_test3.cpp
    HCC :: Unit/FilePath/file_path_test4.cpp

********************
Failing Tests (3):
    HCC :: Unit/HSA/list2.cpp
    HCC :: benchEmptyKernel/bench.cpp
    HCC :: benchEmptyKernel/nullkernel.cpp

  Expected Passes    : 691
  Expected Failures  : 21
  Unexpected Passes  : 4
  Unexpected Failures: 3
make[3]: *** [tests/CMakeFiles/test] Error 1
make[2]: *** [tests/CMakeFiles/test.dir/all] Error 2
make[1]: *** [tests/CMakeFiles/test.dir/rule] Error 2

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -39,6 +39,7 @@
 #include <time.h>
 #include <iomanip>
 
+#define KALMAR_DEBUG (0)
 #ifndef KALMAR_DEBUG
 #define KALMAR_DEBUG (0)
 #endif
@@ -82,6 +83,12 @@
 #define ASYNCOPS_VECTOR_GC_SIZE (1024)
 
 
+//---
+// Environment variables:
+int HCC_PRINT_ENV=0;
+
+int HCC_UNPINNED_COPY_MODE = UnpinnedCopyEngine::ChooseBest;
+
 // Copy thresholds, in KB.  These are used for "choose-best" copy mode.
 long int HCC_H2D_STAGING_THRESHOLD    = 64; 
 long int HCC_H2D_PININPLACE_THRESHOLD = 4096; 
@@ -90,7 +97,12 @@ long int HCC_D2H_PININPLACE_THRESHOLD = 1024;
 int HCC_SERIALIZE_KERNEL = 0;
 int HCC_SERIALIZE_COPY = 0;
 
+int HCC_OPT_FLUSH=0;
+
+#define DB_SYNC 0x2
 unsigned HCC_DB = 0;
+
+int HCC_MAX_QUEUES = 1024;
 
 
 
@@ -176,13 +188,6 @@ static const char* getHSAErrorString(hsa_status_t s) {
 		abort();\
 	}
 
-#define STATUS_CHECK_Q(s,q,line) if (s != HSA_STATUS_SUCCESS) {\
-    const char* error_string = getHSAErrorString(s);\
-		printf("### HCC STATUS_CHECK_Q Error: %s (0x%x) at file:%s line:%d\n", error_string, s, __FILE__, line);\
-                assert(HSA_STATUS_SUCCESS == hsa_queue_destroy(q));\
-                assert(HSA_STATUS_SUCCESS == hsa_shut_down());\
-		abort();\
-	}
 
 // debug function to dump information on an HSA agent
 static void dumpHSAAgentInfo(hsa_agent_t agent, const char* extra_string = (const char*)"") {
@@ -339,25 +344,25 @@ public:
                 _hsaExecutableSymbol,
                 HSA_EXECUTABLE_SYMBOL_INFO_KERNEL_GROUP_SEGMENT_SIZE,
                 &this->static_group_segment_size);
-        STATUS_CHECK_Q(status, commandQueue, __LINE__);
+        STATUS_CHECK(status, __LINE__);
 
         status =
             hsa_executable_symbol_get_info(
                 _hsaExecutableSymbol,
                 HSA_EXECUTABLE_SYMBOL_INFO_KERNEL_PRIVATE_SEGMENT_SIZE,
                 &this->private_segment_size);
-        STATUS_CHECK_Q(status, commandQueue, __LINE__);
+        STATUS_CHECK(status, __LINE__);
 
         workitem_vgpr_count = 0;
 
         hsa_ven_amd_loader_1_00_pfn_t ext_table = {nullptr};
         status = hsa_system_get_extension_table(HSA_EXTENSION_AMD_LOADER, 1, 0, &ext_table);
-        STATUS_CHECK_Q(status, commandQueue, __LINE__);
+        STATUS_CHECK(status, __LINE__);
 
         if (nullptr != ext_table.hsa_ven_amd_loader_query_host_address) {
             const amd_kernel_code_t* akc = nullptr;
             status = ext_table.hsa_ven_amd_loader_query_host_address(reinterpret_cast<const void*>(kernelCodeHandle), reinterpret_cast<const void**>(&akc));
-            STATUS_CHECK_Q(status, commandQueue, __LINE__);
+            STATUS_CHECK(status, __LINE__);
 
             workitem_vgpr_count = akc->workitem_vgpr_count;
         }
@@ -550,9 +555,8 @@ public:
         dispose();
     }
 
-    hsa_status_t enqueueBarrier(hsa_queue_t* queue);
 
-    hsa_status_t enqueueAsync(Kalmar::HSAQueue*);
+    hsa_status_t enqueueAsync(Kalmar::HSAQueue*, bool forceFlush);
 
     // wait for the barrier to complete
     hsa_status_t waitComplete();
@@ -658,7 +662,7 @@ public:
     hsa_status_t dispatchKernelAsyncFromOp(Kalmar::HSAQueue* hsaQueue);
 
     // dispatch a kernel asynchronously
-    hsa_status_t dispatchKernel(hsa_queue_t* commandQueue, const void *hostKernarg,
+    hsa_status_t dispatchKernel(hsa_queue_t* lockedHsaQueue, const void *hostKernarg,
                                int hostKernargSize, bool allocSignal);
 
     // wait for the kernel to finish execution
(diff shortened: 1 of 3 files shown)


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -39,6 +39,7 @@
 #include <time.h>
 #include <iomanip>
 
+#define KALMAR_DEBUG (0)
```
whchung: this line seems redundant?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -763,11 +767,56 @@ pool_iterator::pool_iterator()
...
+    RocrQueue(hsa_queue_t *hwQueue, HSAQueue *hccQueue) : 
+        _hwQueue(hwQueue), _hccQueue(hccQueue) 
+    {
+        assignHccQueue(hccQueue);
+    }
+
+    ~RocrQueue() {
+
+#if KALMAR_DEBUG
+        std::cerr << "HSAQueue::dispose(): destroy an HSA command queue: " << _hwQueue << "\n";
```
whchung, line 806: should be better changed to `RocrQueue::~RocrQueue()` to avoid confusion

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -781,11 +830,15 @@ class HSAQueue final : public KalmarQueue
 
     uint64_t                                      opSeqNums;
 
+    bool                                            valid;
```
whchung, line 856: @bensander could you provide a comment on the meaning of `valid` here? seems it's being used in quite a few places.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1021,7 +1008,7 @@ class HSAQueue final : public KalmarQueue
     int getPendingAsyncOps() override {
         int count = 0;
         for (int i = 0; i < asyncOps.size(); ++i) {
-            auto asyncOp = asyncOps[i];
+            auto asyncOp = asyncOps[i]; // TODO _ shared_ptr copy, can we optimize?
```
whchung: would `auto&` help reduce the copy?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1034,6 +1021,24 @@ class HSAQueue final : public KalmarQueue
         return count;
     }
 
+
+    int isEmpty() override {
+        int count = 0;
+        // Have to walk asyncOps since it can contain null pointers.
+        for (int i = 0; i < asyncOps.size(); ++i) {
+            if (asyncOps[i] != nullptr) {
+                auto asyncOp = asyncOps[i];
```
whchung: would `auto&` reduce the copy?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1607,6 +1673,106 @@ class HSADevice final : public KalmarDevice
...
+                                      UINT32_MAX, UINT32_MAX, &hwQueue);
+#if KALMAR_DEBUG
+            std::cerr << "HSAQueue::HSAQueue(): created an HSA command queue: " << hwQueue << "\n";
+#endif
+            STATUS_CHECK(status, __LINE__);
+
+            status = hsa_amd_profiling_set_profiler_enabled(hwQueue, 1);
+
+
+            auto rq = new RocrQueue(hwQueue, thief);
```
whchung: @bensander could you explain the logic here a bit? It seems you are creating 2 `RocrQueue` instances, `rq` and `rocrQueue`, and push them into `rocrQueues`. Is this really your intended behavior?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1607,6 +1673,106 @@ class HSADevice final : public KalmarDevice
...
+            /// Enable profiling support for the queue.
+            status = hsa_amd_profiling_set_profiler_enabled(hwQueue, 1);
+
+            auto rocrQueue = new RocrQueue(hwQueue, thief);
+            rocrQueues.push_back(thief->rocrQueue);
+            rocrQueue->assignHccQueue(thief);
+            
+
+        } else {
+            while (1) {
```
whchung: could we use a `bool` flag? I feel a bit uneasy to see `while(1)`...

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1607,6 +1673,106 @@ class HSADevice final : public KalmarDevice
...
+
+                                assert (victimHccQueue->rocrQueue == rq);  // ensure the link is consistent.
+                                victimHccQueue->rocrQueue = nullptr; 
+                            }
+                        }
+                            
+                        // update the queue pointers to indicate the theft:
+                        rq->assignHccQueue(thief);
+
+                        return;
```
whchung: and change the flag to break the while loop.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -763,11 +767,56 @@ pool_iterator::pool_iterator()
...
 class HSAQueue final : public KalmarQueue
 {
 private:
-    // HSA commmand queue associated with this HSAQueue instance
-    hsa_queue_t* commandQueue;
+    friend class Kalmar::HSADevice;
+    friend class RocrQueue;
+
+    // ROCR queue associated with this HSAQueue instance. 
+    RocrQueue    *rocrQueue;
```
whchung, line 838: @bensander could you add a bit more comment on the relationship between `HSAQueue` and `RocrQueue`? It seems now `RocrQueue` holds the real handle returned by `hsa_queue_create()`, and `HSAQueue` is a logical instance, and multiple `HSAQueue` instances may share one `RocrQueue`. Is this understanding correct?

whchung (changes requested):
@bensander I've finished my first round of review. could you help check my questions? thanks.

lib/hsa/mcwamp_hsa.cpp
pfultz2: Actually, since you don't use the index, you could just write:

```cpp
for(auto&& asyncOp:asyncOps) {
    if (asyncOp != nullptr) {
        hsa_signal_t signal = *(static_cast <hsa_signal_t*> (asyncOp->getNativeHandle()));
        hsa_signal_value_t v = hsa_signal_load_relaxed(signal);
        if (v != 0) {
            ++count;
        }
    }
}
```

pfultz2: Or even better:

```cpp
return std::count_if(asyncOps.begin(), asyncOps.end(), [](auto&& asyncOp)
{
    return asyncOp != nullptr and 
        hsa_signal_load_relaxed(*(static_cast <hsa_signal_t*> (asyncOp->getNativeHandle()))) != 0;
});
```

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1034,6 +1021,24 @@ class HSAQueue final : public KalmarQueue
         return count;
     }
 
+
+    int isEmpty() override {
+        int count = 0;
```
pfultz2: Count is not used here.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -763,11 +767,56 @@ pool_iterator::pool_iterator()
...
+
+
+// Small wrapper around the hsa hardware queue (ie returned from hsa_queue_create(...).
+// This allows us to see which accelerator_view owns the hsa queue, and
+// also tracks the state of the cu mask, profiling, priority of the HW queue.
+struct RocrQueue {
+    RocrQueue(hsa_queue_t *hwQueue, HSAQueue *hccQueue) : 
+        _hwQueue(hwQueue), _hccQueue(hccQueue) 
+    {
+        assignHccQueue(hccQueue);
```
scchan, line 800: Can we move the hsa_queue creation into the RocrQueue's constructor rather than passing it in as an argument?  I think this is more consistent with the ~RocrQueue where the hsa_queue gets destroyed.

lib/hsa/mcwamp_hsa.cpp
pfultz2: Likewise this could be written as:

```cpp
return not std::any_of(asyncOps.begin(), asyncOps.end(), [](auto&& asyncOp)
{
    return asyncOp != nullptr and 
        hsa_signal_load_relaxed(*(static_cast <hsa_signal_t*> (asyncOp->getNativeHandle()))) != 0;
});
```

lib/hsa/mcwamp_hsa.cpp
scchan: Just wondering if we need to release and then re-acquire the mutex for rocrQueuesMutex after each iteration to give a chance for other threads to release their queue resource while the current thread is spinning looking for an empty queue>

lib/hsa/mcwamp_hsa.cpp
bensander: This should be removed.  We just need the one inside the ifdef.

lib/hsa/mcwamp_hsa.cpp
bensander: This is inside the structure definition so can't use full namespace path.

lib/hsa/mcwamp_hsa.cpp
bensander: Yes.  
Really I think we should rename HSAQueue to something like HSAAcceleratorView - this is actually the implementation of accelerator_view for the "hsa" backend.  
Then RocrQueue could be HSAQueue, this is thin wrapper around the hsa_queue_t.

lib/hsa/mcwamp_hsa.cpp
bensander: Looks like the merge duplicated some code, I will clean it up in next push.

lib/hsa/mcwamp_hsa.cpp
bensander: I change to reference.   The index i is useful for gdb debug at times.

lib/hsa/mcwamp_hsa.cpp
bensander: Good idea.

bensander:
Folks please review updated code.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -763,11 +771,83 @@ pool_iterator::pool_iterator()
...
+    RocrQueue(hsa_agent_t agent, HSAQueue *hccQueue) 
+    {
+        /// Query the maximum size of the queue.
+        uint32_t queue_size = 0;
+        hsa_status_t status = hsa_agent_get_info(agent, HSA_AGENT_INFO_QUEUE_MAX_SIZE, &queue_size);
+        STATUS_CHECK(status, __LINE__);
+
+        assert (__builtin_popcount(MAX_INFLIGHT_COMMANDS_PER_QUEUE) == 1); // make sure this is power of 2.
+        assert(queue_size > MAX_INFLIGHT_COMMANDS_PER_QUEUE*2);
+
```
scchan: Maybe we should move these asserts into HSADevice such that we would only perform them once per device rather than for every queue creation?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1439,44 +1489,69 @@ class HSAQueue final : public KalmarQueue
...
         hsa_status_t status = HSA_STATUS_SUCCESS;
 
         // create shared_ptr instance
         std::shared_ptr<HSABarrier> barrier = std::make_shared<HSABarrier>();
 
         // enqueue the barrier
-        status = barrier.get()->enqueueAsync(this);
+        status = barrier.get()->enqueueAsync(this, false/*agent-flush*/);
+        STATUS_CHECK(status, __LINE__);
```
scchan, line 1526: I think this is changing the semantics of EnqueueMarker (and also EnqueueMarkerWithDependency) by changing it from system scope to agent scope fence.  I think we expose these through the hc::create_marker API so this will break user code the marker was used as the synchronization mechanism between the host and the device?

bensander: The logic is a little confusing but this second parm is "forceFlush" and only force the fences to agent scope if HCC_OPT_FLUSH=1 - which is not enabled yet.

    unsigned fenceBits = (!forceFlush && HCC_OPT_FLUSH) ?
        ((HSA_FENCE_SCOPE_AGENT) << HSA_PACKET_HEADER_ACQUIRE_FENCE_SCOPE) |
        ((HSA_FENCE_SCOPE_AGENT) << HSA_PACKET_HEADER_RELEASE_FENCE_SCOPE)
        :
        ((HSA_FENCE_SCOPE_SYSTEM) << HSA_PACKET_HEADER_ACQUIRE_FENCE_SCOPE) |
        ((HSA_FENCE_SCOPE_SYSTEM) << HSA_PACKET_HEADER_RELEASE_FENCE_SCOPE);

Eventual behavior will be that Markers have System scope by default, but a flag to create_marker can select agent scope in cases where this is all that matters.

scchan:
Approved, thanks!

PR #217, opened 2017-01-21
Exclude llvm static library from the installation

scchan:
hcc doesn't need the llvm static library to function.  Removing them from the installer package saves around 40MB


whchung (approved):
LGTM

PR #218, opened 2017-01-24
Short vector implementation based on llvm native short vector types

scchan:
This is a new implementation of short vectors based on LLVM's short vector data types.  This provides better performance over the current structure based implementation.  For example, the compiler backend could now leverage vector loads/stores instead of generate scalar loads/stores.  It could also leverage vector swizzling instructions.  This also provides the foundation for exploiting packed fp16.

One test case (tests/Unit/AmpShortVectors/amp_short_vectors_2files.cpp) was modified to replace codes relying on a MS C++ extension (replacing access to private members with a getter/setter functions):

v.x = 123;   changed to                  v.set_x(123);
int foo = v.x;   changed to              foo = v.get_x();

This above code pattern just happen to work with the structure based implementation.

All other short vector related tests passed without any modification.


whchung (commented):
@scchan Many thanks for bringing this highly-anticipated change!

include/amp_short_vectors.h
```diff
@@ -3,14 +3,23 @@
 #ifndef _AMP_SHORT_VECTORS_H
 #define _AMP_SHORT_VECTORS_H
 
+#include <cstddef>
+#include <type_traits>
+
```
whchung, line 8: @scchan these 2 lines seems redundant. could them be removed?

include/amp_short_vectors.h
```diff
@@ -3,14 +3,23 @@
...
 {
 
 #define __CPU_GPU__ restrict(cpu, amp)
 
+#if 1
+#include <cstddef>
+#include <type_traits>
+#include "hc_short_vector.inl"
+#else
 #include "kalmar_short_vectors.inl"
```
whchung, line 21: @scchan do you plan to remove `kalmar_short_vectors.inl`?

whchung (approved):
apart from a few minor cosmetic stuffs I'm fine with this PR. I'll take it.

include/amp_short_vectors.h
scchan: Yes, there will be a 1-2 follow up PR for this.  I'm planning to add a few more tests, add half precision support and remove the old implementation.

PR #222, opened 2017-01-26
Allow multiple search paths for ROCm Device Libs

scchan:
This is to allow multiple search paths (a semi-colon separated list) for ROCm device libs because the location of the library may be different between the build environment and the production environment.

Here's a cmake example to configure the search paths for device libs:

cmake -DROCM_DEVICE_LIB_DIR="/build/rocm_device_libs/lib;/opt/rocm/lib"  ....

The search list is also added to clamp-device to detect the library when running hcc.


whchung (approved):
LGTM. Many thanks for this contribution!

PR #228, opened 2017-01-31
Fix tests extract kernel

scchan:
- change nullkernel test to invoke extractkernel from the build directory
- remove XFAIL from all file path tests as they are expected to pass


whchung (approved):
LGTM

PR #231, opened 2017-02-03
New debug messages interface

ot-amd:
A new interface for runtime debug messages that provides more control over debug messages such as turning the messages on and off without having to recompile.

Also allows debug messages to be categorized and only displayed when their corresponding flag is turned on with an environment variable.

The main method of displaying a message would use the C++ standard output style but the printf style is also supported with a different name.

After the interface is finalized, new categories might be added, and current ones modified. Followed by converting existing debug messages to this interface.

[JIRA Ticket SWDEV-109204](http://ontrack-internal.amd.com/browse/SWDEV-109204)

### DIFF
--- a/include/hcc_debug_messages.hpp
+++ b/include/hcc_debug_messages.hpp
@@ -1,0 +1,53 @@
+#pragma once
+
+#include <iostream>
+#include <string>
+#include <stdio.h>
+
+#define DB_MISC		1U << 0
+#define DB_SYNC		1U << 1
+#define DB_AQL		1U << 2
+#define DB_QUEUE	1U << 3
+
+#define DEBUG_ENABLED 1
+
+char* hcc_db;
+std::ostream &err = std::cerr;
+
+#define DEBUG_MESSAGE_PRINTF(category, format, ...) \
+hcc_db = std::getenv("HCC_DB"); \
+if (hcc_db != NULL && DEBUG_ENABLED) { \
+	int db_bitmask = atoi(hcc_db); \
+	if (db_bitmask & category) {  \
+		fprintf (stderr, format, __VA_ARGS__); \
+	} \
+}
+
+template <typename T>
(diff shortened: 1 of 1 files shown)


include/hcc_debug_messages.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+#define DB_MISC         1U << 0
+#define DB_SYNC         1U << 1
+#define DB_AQL          1U << 2
+#define DB_QUEUE        1U << 3
+
+#define DEBUG_ENABLED 1
+
+char* hcc_db;
+std::ostream &err = std::cerr;
```
whchung, line 15: considering the cast where multiple object files may include this header file, could you put them into an anonymous namespace to prevent from breaking ODR rule?

whchung (changes requested):
thanks for submitting this PR. as there is no description about this PR, and it seems it's not completed yet, please kindly summarize what you want to achieve in the further commits / PR.

bensander:
Also take a look at tip clang_tot_upgrade, there have been other improvements in the debug infrastructure (but still more work to do).  One specific item I’d like to see is to replace the KALMAR_DEBUG compile-time macro with new runtime debug.  This will require reviewing existing debug message and categorizing them appropriately.
Also I’d like to keep the macro for wrapping debug messages short to keep the code clean, I believe currently we are using “DBOUT”.
-ben

include/hcc_debug_messages.hpp
ot-amd: Updated the description.

For the anonymous namespace, wouldn't it have no effect? Because ostream &err is never changed in the code and hcc_db is getting the environment variable value which isn't going to change during runtime.

whchung:
As @bensander has improved incumbent debug interfaces, I'm closing this one. @otauqir , please work with @bensander to consolidate your efforts, thanks.

PR #232, opened 2017-02-06
Mark atomic add/sub tests with fp as invalid (with XFAIL) 

scchan:
C++ atomic_fetch_add/sub only support integers.  
These tests apply atomic_fetch_add/sub on floating-point types and the clang FE would do implicit conversion into integral types, which cause the tests to fail.  
We should change these tests to use compare-and-swap instead in the future


whchung (approved):
LGTM

PR #233, opened 2017-02-06
Increase checksum cutoff

scchan:
shared_library2 and shared_library3 tests started to fail after recent clang and llvm update.  

It turns out the checksum value for the GPU binary blobs in these tests collide and therefore, one of the GPU blobs didn't get loaded.

Increase the cutoff size to avoid the collision.


whchung (approved):
LGTM

whchung:
@scchan The algorithm is now executed much less frequently than before. Do you think would it be a good idea if we simply drop this cutoff value, and use the whole kernel source buffer to calculate the hash?

scchan:
@whchung   yea I agree, it would be less risky without the cutoff.  Also it saves time and energy... it took me a while to figure that out :)

PR #234, opened 2017-02-07
remove cutoff value when computing kernel checksum

scchan:
as discussed, remove the cutoff value in kernel checksum


whchung (approved):
LGTM

PR #235, opened 2017-02-07
Enable Jenkins build CI with GitHub integration

kknox:
I am enabling automated build and tests of hcc commits and pull requests though our internal libraries jenkins farm.  The build farm includes client machines capable of running ROCm tests as part of the build process. The jenkins infrastructure will mark each PR/commit with in-progress/pass/fail results, to let maintainers know whether it is safe to merge or commit new code.  The intent is to provide a basic set of build files, but my hope is that hcc maintainers can help maintain the build scripts and update tests to benefit continued automation.  The stage() called 'unit testing' is empty on check-in, but could be used to execute sanity checks on a successful build.  The pass/fail results are clickable and a web page with more detailed build results and logs are provided given that the client is behind the corporate firewall.

This commit includes the addition of 5 new files.

**docker-compose.yml**

A convenience file for building docker images and running containers from the command line.  This file is not necessary for jenkins continuous automation testing, but is helpful to debug images and containers.  docker-compose may be installed on a linux system like so:  
```bash
sudo curl -L $(curl -s https://api.github.com/repos/docker/compose/releases/latest | grep browser_download_url.*-Linux | cut -d\" -f 4) -o /usr/local/bin/docker-compose
sudo chmod +x /usr/local/bin/docker-compose

sudo curl -L https://raw.githubusercontent.com/docker/compose/$(docker-compose version --short)/contrib/completion/bash/docker-compose -o /etc/bash_completion.d/docker-compose
```

The dockerfiles in the /docker directory may receive arguments to change build types or build paths, so the docker-compose.yml file is a nice way to manage all the arguments and provide default values for build.

**Jenkinsfile**

The primary build script for the Jenkins farm.  This utilizes the docker infrastructure to build hcc-lc inside of a docker container.  Modify this file as necessary to change the way hcc-lc builds.

**docker/dockerfile-build-ubuntu-16.04**

Docker build script for building an image with all the necessary dependencies to build hcc-lc on ubuntu 16.04

**docker/dockerfile-build-ubuntu-14.04**

Docker build script for building an image with all the necessary dependencies to build hcc-lc on ubuntu 14.04

**docker/dockerfile-hcc-lc-ubuntu-14.04**

Docker build script for building a clean image with hcc-lc and dependencies.  No build dependencies included.

Unfortunately, it is not possible to use variable interpolation in a dockerfile **FROM** statement, so there are separate build files for each platform.  However, in practice this is turns out to be OK, because the build instructions for installing on Ubuntu are significantly different than building on Fedora.  The package names are different, the commands to update software are different, the compiler and library versions are different. It may turn out to be more maintainable in the long term to keep distinct dockerfiles for different platforms.


scchan:
@kknox could we push this into a branch first?  I'll have to make a couple of changes to the build instructions in ubuntu16.04 dockerfile

kknox:
Feel free to change the PR in whatever way you need.  I'm glad to hear that the build code is understandable.

You should be able to retarget the destination branch in the PR by clicking the edit button at the top of this page.  You can also push additional commits into the PR; I think you just clone kknox/hcc and add the necessary edits.

pfultz2:
@scchan I updated it to go to the feature-jenkins branch.

PR #239, opened 2017-02-09
Improve and simplify finding dependencies in cmake

pfultz2:
* Rocm device libs are found using `PATH_SUFFIXES` so that it can be found along the dependency path. 

* The `ROCM_ROOT` is searched after searching the dependency path, so that when the user specifies the path to some locally installed rocm deps(ie `-DCMAKE_PREFIX_PATH=/path/to/local/rocm/stuff`) it will find those over the globally installed rocm. 

* Although the variables `HSA_HEADER_DIR`, `HSA_LIBRARY_DIR`, and `ROCM_DEVICE_LIB_DIR` can still be used, these variables are completely unnecessary and should be deprecated as cmake already creates a cached variable when calling the find commands. So, the message now directs the user to set `HSA_HEADER`, `HSA_LIBRARY`, and `ROCM_DEVICE_LIB` instead, which cannot only be set at the command line, but also shows up in tools like `ccmake`, which the user can set as well.


CMakeLists.txt
```diff
@@ -64,56 +64,45 @@ LINK_DIRECTORIES( ${LLVM_LIB_DIR} )
...
-
-  find_path(ROCM_DEVICE_LIB ocml.amdgcn.bc PATHS ${ROCM_ROOT}/lib NO_DEFAULT_PATH)
-  find_path(ROCM_DEVICE_LIB ocml.amdgcn.bc)
-
-  if (NOT ROCM_DEVICE_LIB)
-    MESSAGE(FATAL_ERROR "ROCm Device Libs not found. Use -DROCM_DEVICE_LIB_DIR=<path_to_ocml.amdgcn.bc>.")
-  endif (NOT ROCM_DEVICE_LIB)
-
-endif(ROCM_DEVICE_LIB_DIR)
+find_path(ROCM_DEVICE_LIB ocml.amdgcn.bc
```
scchan, line 94: The path to the ROCm device lib may be different on a build system and on a production system so we need the capability to be able to find rocm device lib from a list of potential locations during the execution of hcc.  That's the reason ROCM_DEVICE_LIB has to stay as a list of paths rather than the path determined by cmake when building hcc.  See https://github.com/RadeonOpenCompute/hcc/blob/clang_tot_upgrade/lib/clamp-device.in#L59

If you decide to modify this part of cmake, please make sure you maintain this functionality.

scchan (changes requested):

CMakeLists.txt
pfultz2: I don't understand the need for it to be a list. `ROCM_DEVICE_LIB` can be set for whats needed for a build system when building on a build system, and then set for whats needed for a production system when building for production. You don't want to set both at the same time as you don't want a production system searching for paths that are for the build system, right?

pfultz2:
So I updated it to search the paths set by `ROCM_DEVICE_LIB_DIR`.

CMakeLists.txt
scchan: yes, it needs to be searched when hcc links in the libraries since tests may be run on the build system.  This is how the internal build is being setup unfortunately where the device libs are located at a non-default location

CMakeLists.txt
```diff
@@ -64,56 +64,48 @@ LINK_DIRECTORIES( ${LLVM_LIB_DIR} )
...
+  PATHS
+    ${ROCM_ROOT}/lib
+)
+
+if (NOT ROCM_DEVICE_LIB)
+  MESSAGE(FATAL_ERROR "ROCm Device Libs not found. Use -DROCM_DEVICE_LIB=<path_to_ocml.amdgcn.bc>.")
+endif (NOT ROCM_DEVICE_LIB)
+
+set(ROCM_DEVICE_LIB_PATHS ${ROCM_DEVICE_LIB})
+list(APPEND ROCM_DEVICE_LIB_PATHS ${ROCM_DEVICE_LIB_DIR})
```
scchan: Can we also append ${ROCM_ROOT}/lib ?  This covers the case where cmake didn't find device libs in ROCM_ROOT/lib but hcc would still search in ${ROCM_ROOT}/lib during linking?

scchan (changes requested):

CMakeLists.txt
pfultz2: Yea

scchan (approved):

Issue #92, opened 2016-07-02
Capturing temporary object by reference in completion_future


mcopik:
Method then() in completion_future, responsible for launching a callback functor when completion_future is ready, takes a const reference to functor and then passes it **by reference to newly created thread:**
https://github.com/RadeonOpenCompute/hcc/blob/master/include/hc.hpp#L953

Hence a reference to a temporary object may be saved by the thread. The result is quite obvious to foresee - a mysterious memory corruption when the callback is launched. It took us three days to figure it out why this piece of code is failing:

`
some_data * ptr = ....;
auto fut = device.create_marker();
fut.then([ptr]() { /* do something with ptr which is already garbage */ });
` 

C++AMP API declares this function as taking a const reference, not universal reference which would enforce moving an rvalue) but the whole issue could be solved just by capturing the functor by value in the lambda passed to std::thread.

david-salinas:
@AlexVlx @scchan I believe we can close this issue, as it is resolved with  PRs #353 / #354  ?

AlexVlx:
I'm not sure I see how any of those PRs does anything for this issue, which is one of interface design...

mcopik:
@david-salinas @AlexVlx I don't think it's a problem with interface design since this API was almost entirely copied from C++AMP. It might not be perfect but there was an obvious bug in the implementation - catching a possibly temporary object by reference for later execution. Is it fixed?

mcopik:
I don't know which branch is the main one but `clang_tot_upgrade` seems to be the most active one. The bug is **not** fixed there:
https://github.com/RadeonOpenCompute/hcc/blob/clang_tot_upgrade/include/hc.hpp#L1284

AlexVlx:
Whilst I worked on C++AMP (the original) and I have great appreciation for it, I don't think it got everything right, so saying that something was copy&pasted from there, whilst flattering does not axiomatically guarantee correctness. IMHO, the functor should've been passed by value and treated as a sink. Anyhow, nothing changed in the ::then implementation, and this issue is not addressed, hence my confusion in the original question I asked.

PR #240, opened 2017-02-13
create_blocking_marker from array_view range

AlexVlx:
This is a first iteration of the extended create_blocking_marker interface suggested in current conversation. It introduces an alternate form for create_blocking_marker which accepts a range of accelerator_views as arguments, thus making it possible to straightforwardly create a barrier that crosses array_view boundaries. This latter variant is not yet finalised as I somewhat doubt that I captured the design intent fully. A number of stylistic adjustments are also included (e.g. using = default where reasonable), alongside a fair share of trailing whitespace removal. I would appreciate a review here, as opposed to immediate adoption. In its current form this commit creates issues with some of the grid_launch tests. Thank you for the review.

### DIFF
--- a/include/hc.hpp
+++ b/include/hc.hpp
@@ -24,6 +24,12 @@
 #include "hsa_atomic.h"
 #include "kalmar_cpu_launch.h"
 
+#include <future>      // For std::async, std::future and std::shared_future.
+#include <iterator>    // For std::iterator_traits.
+#include <memory>      // For std::shared_ptr.
+#include <type_traits> // For std::false_type, std::is_same, and std::true_type.
+#include <utility>     // For std::move.
+
 #ifndef __HC__
 #   define __HC__ [[hc]]
 #endif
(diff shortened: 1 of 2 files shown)


whchung (commented):
@AlexVlx is this PR a squashed commit? As the diff is quite large, I'm wondering would it be possible to not squash your commits, and separate functional changes and stylistic ones into different PRs?

AlexVlx:
@whchung it's not squashed, but I've only now realised that the automatic trailing whitespace removal creates a ton of noise. I can definitely break things up, and I apologise for the inconvenience however I was wondering if it would be terribly onerous of me to ask for a peek / assessment of the changes around completion_future, accelerator_view and accelerator_view::create_blocking_marker, where the functional meat is. They are all in hc.hpp, and since this is the part which will require further refinement before adoption, it would be helpful to fuse the separation with review feedback incorporation for the upcoming PRs. I understand if this is not feasible on your end. Thank you!

include/hc.hpp
```diff
@@ -117,6 +451,188 @@ inline uint64_t get_tick_frequency() {
...
+
+    template<typename InputIterator>
+    completion_future create_blocking_marker_(InputIterator f,
+                                              InputIterator l,
+                                              memory_scope scope,
+                                              std::true_type) const
+    {   // Precondition: readable_bounded_range(f, l) && non_empty(f, l).
+        static_assert(std::is_same<typename std::iterator_traits<InputIterator>::value_type,
+                                   completion_future>::value,
+                      "This requires accelerator_view arguments.");
```
bensander, line 567: Error should say "This requires completion_future arguments"

include/hc.hpp
```diff
@@ -117,6 +451,188 @@ inline uint64_t get_tick_frequency() {
...
+
+    explicit
+    accelerator_view(std::shared_ptr<Kalmar::KalmarQueue> queue)
+        : queue_{std::move(queue)} {}
+
+    template<typename InputIterator>
+    completion_future create_blocking_marker_(InputIterator f,
+                                              InputIterator l,
+                                              memory_scope scope,
+                                              std::true_type) const
```
pfultz2, line 563: It's really not clear that `true_type` and `false_type` is related to the `static_assert` below, especially since they aren't mutually exclusive. I think using `enable_if` would be cleaner.

```cpp
template<typename InputIterator, typename std::enable_if<(
    std::is_same<typename std::iterator_traits<InputIterator>::value_type, completion_future>{}
), int>::type = 0> 
```

And I think a macro and template alias can help reduce the noise:

```cpp
#define HCC_REQUIRES(...) typename std::enable_if<(__VA_ARGS__), int>::type = 0

template<class Iterator>
using iterator_value = typename std::iterator_traits<Iterator>::value_type;

template<typename InputIterator, 
    HCC_REQUIRES(std::is_same<iterator_value<InputIterator>, completion_future>{})
>
```

Plus this puts the type requirements for the template parameter next to it. 

But, I don't know how everyone else thinks about it?

include/hc.hpp
```diff
@@ -117,6 +451,188 @@ inline uint64_t get_tick_frequency() {
...
+    {   // Precondition: readable_bounded_range(f, l) && non_empty(f, l).
```
pfultz2, line 564: Is there a reason why it doesn't handle empty ranges? There should probably be some kind of assert for this condition.

include/hc.hpp
```diff
@@ -108,6 +114,334 @@ inline uint64_t get_tick_frequency() {
...
+    std::future<void> then(F&& func)
+    {   // TODO: this should be re-assessed. For now it treats completion_future
+        //       as an equivalent of std::experimental::shared_future i.e. it is
+        //       safe to call .then() multiple times, with different
+        //       continuations
+        if (valid()) {
+            return std::async([](const completion_future& fut, F&& fn) {
+                fut.wait();
+                fn();
+            }, *this, std::forward<F>(func));
```
pfultz2, line 372: I think this fixes issue #92.

include/hc.hpp
```diff
@@ -255,95 +789,219 @@ class accelerator_view {
...
+     *         been completed.
      */
-    void copy(const void *src, void *dst, size_t size_bytes) {
-        pQueue->copy(src, dst, size_bytes);
+    template<typename InputIterator>
+    completion_future create_blocking_marker(InputIterator first,
+                                             InputIterator last,
+                                             memory_scope scope) const
+    {
+        if (first == last) return completion_future{};
```
pfultz2, line 859: I see, you take care of the empty condition here.

include/hc.hpp
AlexVlx: Very good eyes, thank you for the observation. true_type and false_type are placeholders, in the final form I'll use proper tag dispatch (i.e. there will be a completion_future_tag and an accelerator_view_tag), which I think is a better fit here. I would absolutely emphatically not use a macro for requires / concept checking (although I love the latter to bits). In what regards type functions, those are something else that are to be added but I've kept out so as to minimise noise and because it's not very clear (to me, yet) where they should live - they are generally useful, so I would like to incorporate them in the code-base in general.

include/hc.hpp
AlexVlx: Right, it does not handle empty-ranges because it is an implementation detail, and the outer caller (create_blocking_marker) already deals with that itself. Properly asserting around this is a pretty great idea indeed; they will show up before final PR, thank you.

include/hc.hpp
```diff
@@ -117,6 +451,188 @@ inline uint64_t get_tick_frequency() {
...
+                    deps[n] = f->queue_->EnqueueMarker(scope);
+                    ++n;
+                    ++f;
+                }
+            } while (f != l && n != max_async_ops);
+
+            if (f == l) {
+                return completion_future{queue_->EnqueueMarkerWithDependency(n,
+                                                                             deps,
+                                                                             scope)};
```
bensander, line 620: Should we move the code out of the header?
There was some indirection that HCC runtime typically provides for other targets (ie CPU), I think this code may only work with the hsa/rocr back-end?

bensander (changes requested):

include/hc.hpp
AlexVlx: Thank you, good catch!

include/hc.hpp
AlexVlx: Right, this would fail for e.g. a CPUQueue, however I'm not sure we're ever using a CPU device anywhere at this point? If we are, it might be cleaner to provide a token implementation there, perhaps? Or just disable this code for anything but a HSAQueue. It's worthwhile to note that I do not quite believe that create_blocking_marker works with a CPUQueue anyhow, even in its current implementation, therefore we might be straying into uncharted territory.

bensander:
I think we could & should preserve the layer of indirection where all the HSA-backend-specific code is inside mcwamp_hsa.cpp, we will likely use this in the future even if it doesn’t work now.

include/hc.hpp
scchan: @pfultz2 thanks for the hint! :)

Is it possible to solve this with partial specialization?

include/hc.hpp
scchan: I asked Weixing to look into how we could improve the architecture of the runtime, potentially removing the "plug-in" approach.  We should have some discussion on this...

scchan:
Could you provide some test results, including HCC direct test, HIP direct test, Rodinia, etc?

scchan (changes requested):
It looks your PR includes changes from previous recent commits.  Perhaps you should rebase, isolate your changes and create a PR again.

whchung:
@AlexVlx is this PR is relevant, after recent PRs #353 / #354 from @bensander ?

AlexVlx:
@whchung it is quite stale independently from those, I forgot to close it.

PR #241, opened 2017-02-14
enable short vector for half fp

scchan:


whchung (changes requested):
@scchan could you add unit tests for this PR? Or change existing tests and add new function templates?

scchan:
sure, will add half precision short vector tests

kknox:
Just a reminder, if you add unit tests, you should add logic to [run the tests in the jenkinsfile](https://github.com/RadeonOpenCompute/hcc/blob/clang_tot_upgrade/Jenkinsfile#L82) so that unit tests are considered part of a successful build

scchan:
@whchung I modified an existing test to include half and created a new unit test with half

scchan:
@whchung I added new tests as you suggested, could you review this again?  thanks

whchung (approved):
LGTM. Thanks!

PR #243, opened 2017-02-16
Fix bug which impacted copy performance.

bensander:
For copies involving a single GPU (no P2P), specify host agent
when appropriate.  Previously the code used copy-agent for both src agent
and dst agent, this would cause ROCR to use a blit kernel. With
fix, we are using SDMA again.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -509,7 +509,7 @@ public:
 
 
 private:
-  hsa_status_t hcc_memory_async_copy(const Kalmar::HSADevice *copyDevice, void *dst, const void *src, size_t sizeBytes,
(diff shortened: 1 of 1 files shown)


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4000,11 +4000,46 @@ hsa_status_t HSACopy::hcc_memory_async_copy(const Kalmar::HSADevice *copyDeviceA
...
+            throw Kalmar::runtime_exception("bad copyKind in hcc_memory_async_copy", copyKind);
+    };
+
+
+    /* ROCR logic to select the copy agent:
+     *  
+     *  Decide which copy agent to use : 
+     *   
+     *   1. Pick source agent as long as the src agent is a GPU (regardless of the dst agent).
+     *   2. Pick destination agent only if the dst agent is a GPU.
```
scchan: just nickpicking here... should it be read as  "Pick destination agent **if only** the dst agent is a GPU"?

scchan (commented):
LGTM, let me know if we want to update the comment before merging

bensander:
Updated comments in file (related to ROCR handling)

PR #247, opened 2017-02-23
Jenkins docker ci

kknox:
Closes #235 

This is PR 235 rebased to the top of clang_tot_upgrade.  Same description applies, but in addition
* the hcc cmakefile autodetects the hosts distribution
* automatically determines if it should link to libc++ or libstdc++ based on g++ version
* appends c++ runtime lib name to the package


docker/dockerfile-build-ubuntu-16.04
```diff
@@ -0,0 +1,79 @@
...
+    g++-multilib \
+    libncurses5-dev \
+    libelf-dev \
+    libc++abi-dev \
+    libc++-dev \
+    hsa-rocr-dev && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/*
+
+# This installs the new cmake side-by-side with the old cmake using update-alternatives
```
scchan, line 37: We probably don't need to update the cmake on 16.04

scchan:
LGTM overall.  The comment regarding cmake is a minor issue that we could address later on.

docker/dockerfile-build-ubuntu-16.04
kknox: It's true; updating cmake for 16.04 is not necessary.  However, i wrote the cmake update logic anyways for 14.04, then I thought it might as well be used in 16.04 just to have the latest.  I don't clobber the cmake installed on the system; i install to /opt and use update-alternatives to default to the newer cmake.  If the shipping cmake is desired, update-alternatives could be used to switch back to the distro cmake.

If you'd rather not suffer the compile times on image build, it could be ripped out safely.

PR #252, opened 2017-02-27
add compiler-rt to hcc

scchan:
hcc needs conversion builtins from double/single fp to half fp on the host, which are provided by compiler-rt.

- Add compiler-rt to hcc as a submodule
- Update cmakefile to build and to install the builtin library
- Update clang driver to link in the builtin library
- Add unit tests to test the conversion builtins


.gitmodules
```diff
@@ -10,3 +10,6 @@
 	path = clang
 	url = https://github.com/RadeonOpenCompute/hcc-clang-upgrade.git
 	branch = clang_tot_upgrade
+[submodule "compiler-rt"]
```
kknox, line 13: You are adding a new submodule here, but the Jenkinsfile I recently checked in checks out the submodules manually.  If you want CI to also build compiler-rt, you need to add a new [git clone into Jenksinsfile](https://github.com/scchan/hcc/blob/add_compiler-rt/Jenkinsfile#L32).

I know this is sub-optimal, but cloning with `git clone --recursive -b <name>` takes greater than 15 minutes per build.  Sometimes, this clone step would time out and fail the build, also sub-optimal.  As a compromise, I manually clone each submodule with depth 1 which makes the whole clone operation take < 2 minutes.  The downside is two files to update when new submodules are added/deleted (hopefully infrequently).

whchung: @kknox from checkout speed perspective, do you think moving to repo bring any benefit?

kknox: Iff the repo command allows an easy way to clone these submodules with depth=1, then yes.  For CI, all we care about for each submodule is the HEAD commit.  I feel like the [series of git clone commands](https://github.com/scchan/hcc/blob/add_compiler-rt/Jenkinsfile#L32) I hardcoded basically does what repo does.

I have tried (and I know Paul has tried to) to clone hcc with the --shallow-submodules flag, and it doesn't work.  We also tried a lot of different permutations with `git submodule update --depth=1`, but those also fail.

scchan: updated the Jenkinsfile...I had this in my mind but so how still forgot to update it :)

kknox:
I'm a little disappointed that the jenkins CI hasn't kicked off for this PR yet.  I assume it doesn't try because of the merge conflicts.

scchan:
I fixed the merge conflict

kknox:
Huh, the dashboard still says there are conflicts (with submodules?), but you fixed something because you got a green checkmark.

scchan:
I fixed the merge conflict again (which was caused my merge update clang_tot_upgrade).

@whchung If you have a chance, could you review this?  Otherwise I'd have to maintain this PR as well after I did the daily merge update... thanks!

whchung (approved):
LGTM, especially SYNC_UPSTREAM.md is changed to cover the steps to sync with upstream compiler-rt.

scchan:
Thanks, this PR also depends on https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/29

whchung:
Thanks for your notification. It's done.

PR #254, opened 2017-03-02
enable amdgpu-internalize-symbols optimization

scchan:
Enable amdgpu-internalize-symbols which optimize the object code size by removing unreferenced symbols.  This is probably more useful to the promotion-free pass since the clang_tot_upgrade has the erase-no-kernel pass.


whchung (approved):
LGTM

PR #256, opened 2017-03-04
Support for profiling using CodeXL Activity Logger

ot-amd:
Adds support for the CodeXL Activity Logger for profiling hcc.
The activity logger is initialized in the HSAContext object and a marker is placed there as an example.
Additional markers will be placed in other locations for profiling in future commits.

In user applications, only the marker objects have to be created. No need to initialize the activity logger if hcc is built with the activity logger macro enabled as outlined in the readme.


whchung:
I'm wondering would it be possible to design the feature so it sits inside a dynamically loaded .so, and enabled via an env var?

This way we can make this feature always be built by default as long as cmake can locate codexl. Users who use hcc as binary packages can also benefit from codexl profiling immediately without having to reconfigure and rebuild hcc.

whchung:
@otauqir ping

whchung:
close after getting consent from @otauqir

PR #258, opened 2017-03-07
Avstress fix

sunway513:
This PR fix the memory leaks issue for avstress tests.


tests/Unit/AcceleratorViewCopy/avstress_0x18.cpp
```diff
@@ -68,25 +68,33 @@ int main(){
         }
     }
 
-    if (testsToRun & 0x8) {
+// Create a vector of hc::completion_future for async copy synconization
+    std::vector<hc::completion_future> cfs;
+    
+    if (testsToRun & 0x18) {
```
scchan: Could we use more meaningful names instead of an hex value?  
For example, if (testsToRun & HostToDeviceAsyncCopyTest) ?

tests/Unit/AcceleratorViewCopy/avstress_0x18.cpp
```diff
@@ -68,25 +68,33 @@ int main(){
...
-            Test4(av);
+            cfs.push_back(Test4(av));
         }
+        for(uint32_t i=0;i<testIters;i++)
+	    cfs[i].wait();
     }
 
-    if (testsToRun & 0x10) {
+    cfs.clear(); 
+    if (testsToRun & 0xFF) {
```
scchan: When we choose to use a mask bits, usually we'd want to be able to enable certain capabilities individually (i.e. run test 1 and test 5 only).  In this unit test, the selection of the conditions doesn't allow that.  For example, if you want to enable test5, then it's mandatory to run test1 through test5 (if I understand this test, running just test5 should be fine).

tests/Unit/AcceleratorViewCopy/avstress_0x18.cpp
```diff
@@ -68,25 +68,33 @@ int main(){
...
+    if (testsToRun & 0x18) {
         for(uint32_t i=0;i<testIters;i++){
             if ((i%1000 == 0)) {
                 printf ("info: running Test4 %5d/%5d\n", i, testIters);
             }
-            Test4(av);
+            cfs.push_back(Test4(av));
         }
+        for(uint32_t i=0;i<testIters;i++)
+	    cfs[i].wait();
```
scchan, line 82: Instead of keeping all the completion_futures (unless you want to stress the completion_future creation) in a vector, you could simply call the wait() on the accelerator_view at the end av.wait();
which will wait on the last command send to the accelerator_view

scchan (changes requested):

tests/Unit/AcceleratorViewCopy/avstress_0x18.cpp
sunway513: Thanks, that'll be simpler. Does HCC covers stress test of completion_futures in the other unit tests? If not, maybe keep it as is?

tests/Unit/AcceleratorViewCopy/avstress_0x18.cpp
sunway513: @adityaatluri , can you clarify here?

whchung (changes requested):
@sunway513 could you address points raised by @scchan? Once they are done I can take it

sunway513:
@whchung , I've made changes to use meaningful test names instead of hex bits. 
The other changes might need more discussion.

whchung (approved):
LGTM. Thanks for finishing up this PR.

tests/Unit/AcceleratorViewCopy/common.h
```diff
@@ -11,6 +11,12 @@
     printf ("error: TEST FAILED\n%s", KNRM );\
     abort();
 
+#define HostToDeviceCopyTest 0x1
+#define DeviceToDeviceCopyTest 0x2
+#define DeviceToHostCopyTest 0x4
+#define HostToDeviceAsyncCopyTest 0x18
+#define DeviceToHostAsyncCopyTest 0xFF
+
```
scchan: I think you've actually introduced 2 bugs to these bit mask.  If you refer to line 71 and line 80 in the original avstress_0xFF.cpp, HostToDeviceAsyncCopyTest should be 0x8 and DeviceToHostAsyncCopytTest should be 0x10.

The default RUNMASK should be 0x1F to enable all the tests

sunway513: Hi @scchan , thanks for pointing it out. It's actually not a bug, as it won't affect any units of the two avstress tests as what I've tested locally. But I've changed them back, for your concern here. Please review.

scchan: It was a bug because it lost the ability to be able to run a particular test individually as I pointed out previously.  For example, after you switched back to the old hex values, now someone could run the DeviceToHostAsyncCopyTest alone without having to run all the previous tests.

scchan (approved):

tests/Unit/AcceleratorViewCopy/common.h
sunway513: Thanks SiuChi, that makes sense.

PR #261, opened 2017-03-10
Installer package improvements

scchan:
- fix library path issue on Fedora
- updated package dependency list


whchung (approved):
LGTM. Thanks!

PR #263, opened 2017-03-10
cmake: Use ${CMAKE_INSTALL_LIBDIR} consistently

jvesely:
Fixes build on fedora where .so libs were installed in 'lib64' but hcc-config pointed to 'lib'
Also fix clang builtin location

Signed-off-by: Jan Vesely <jan.vesely@rutgers.edu>


scchan:
#261 also covers this, could you take a look?

jvesely:
yeah, e039e9724 works for me.
It misses the includedir change, though.

scchan:
The include path is set here and it is not using CMAKE_INSTALL_INCLUDEDIR https://github.com/RadeonOpenCompute/hcc/blob/clang_tot_upgrade/include/CMakeLists.txt#L3

jvesely:
right, I stand corrected.
thanks

PR #266, opened 2017-03-15
add gfx900 option, remove gpu code names

scchan:


scchan:
This PR depends on https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/31

scchan:
@whchung could you take a look when you got a chance?

whchung (approved):
LGTM. Processors.td in AMDGPU backend has incorporated these processor IDs so it's no longer to do such mapping in the frontend.

PR #267, opened 2017-03-20
refine cmake files:

dfukalov:
1. use add_subdirectory to include llvm/clang tree instead of add_custom_target.
   there were explicit cmake&make invokes for each hcc build
2. version string building code moved out to function add_version_info_from_git
3. removed confusing instruction to do "make world" for first time
4. removed a number of not used variables (like EXECUTABLE_OUTPUT_PATH)
5. removed use of ProcessorCount() since the llvm tree make is a part of whole
   hcc build
6. removed using CMAKE_FORCE_CXX_COMPILER - it is deprecated in cmake 3.6


cmake-tests/CMakeLists.txt
```diff
@@ -1,8 +1,9 @@
+set(CMAKE_CXX_COMPILER "${PROJECT_BINARY_DIR}/compiler/bin/clang++")
 
-if(COMMAND CMAKE_FORCE_CXX_COMPILER)
-CMAKE_FORCE_CXX_COMPILER("${PROJECT_BINARY_DIR}/compiler/bin/clang++" MCWAMPCC)
-endif()
 add_executable(cmake-test cmake-test.cpp)
+
+add_dependencies(cmake-test world)
```
pfultz2: This breaks the test when ran as standalone, which is useful for testing the `find_package(hcc)` logic. Either this line should be moved into the `if(TARGET hccrt)` conditional block or hccrt should be update to depend on the world target, so this line can be removed.

CMakeLists.txt
```diff
@@ -325,29 +287,19 @@ set(KALMAR_BACKEND "HCC_BACKEND_AMDGPU")
...
-          -DKALMAR_BACKEND_COMMIT=${KALMAR_BACKEND_COMMIT}
-          -DKALMAR_BACKEND=${KALMAR_BACKEND}
-          -DLLVM_TARGETS_TO_BUILD="AMDGPU\;X86"
-          -DLLVM_INCLUDE_EXAMPLES=off
-  COMMAND make -j ${NUM_BUILD_THREADS} # not portable, but we need it this way
-  WORKING_DIRECTORY ${CLANG_BIN_DIR}
-)
+
+set(LLVM_TARGETS_TO_BUILD "AMDGPU" "X86"
+    CACHE STRING "Semicolon-separated list of targets to build, or \"all\"." FORCE)
```
pfultz2: Is using `FORCE` necessary here? It seems this won't allow the user to choose different targets from the command-line.

cmake-tests/CMakeLists.txt
scchan: @dfukalov could you address this?

scchan (changes requested):
Please address concerns raised by @pfultz2

CMakeLists.txt
dfukalov: you're right, I'll remove it

cmake-tests/CMakeLists.txt
dfukalov: moving `add_dependencies(cmake-test world)` into `if(TARGET hccrt)` block doesn't work for me since cmake-test binary wants .so and .bc files in right places. I'm reworking the change now

dfukalov:
I've just updated the change according to Paul' comments. `cmake-test` dependency of `world` target moved into `hccrt` case it works for `all` build.
`make cmake-test` in clean hcc' build directory fails, I'm going to fix it in separate change since it'll modify more files

pfultz2:
> make cmake-test in clean hcc' build directory fails,

Why does that fail?

dfukalov:
this target at least uses clamp-device script that invokes opt and llc and they uses our passes from .so files and all of them are located in install bin directory.
but cmake-test target in hcc build tree has no dependency from these (llc opt etc.) targets

dfukalov:
Paul, I've just prepared next change that fixes cmake-test target clean build. Is this branch ready to land? Or I should merge cmake-test fix into the change?

pfultz2:
> Paul, I've just prepared next change that fixes cmake-test target clean build. Is this branch ready to land? Or I should merge cmake-test fix into the change?

If you have the cmake-test fix ready, then why not?

whchung (approved):
LGTM. @scchan ?

scchan (approved):

PR #269, opened 2017-03-24
add llvm libs needed by hip

scchan:
include certain llvm static libs needed by HIP to extract metadata


pfultz2:
The cmake files needs to be installed as well, so we can call `find_package(LLVM)`

aditya4d:
@pfultz2 It is added.
https://github.com/RadeonOpenCompute/hcc/pull/269/commits/1244fa3afc77392da197f9f7bf504071fa5b8e65#diff-af3b638bc2a3e6c650974192a53c7291L364

whchung (approved):

PR #270, opened 2017-03-27
update clang for inline asm, add inline asm tests

scchan:
Enable inline asm support

Note this is for promote-free branches only


whchung:
@scchan build log says:
```
manifest invalid: manifest invalid
```

whchung (approved):
LGTM. I'll merge it even though build server fails. The log suggests it's a configuration issue.

PR #273, opened 2017-03-28
CPACK improvements

scchan:
- allow generating rpm package on Ubuntu
- allow overriding the installer package destination (set CPACK_SET_DESTDIR to OFF and set the desired  path with CPACK_PACKAGING_INSTALL_PREFIX)


whchung (approved):
LGTM

PR #274, opened 2017-03-29
update dockerfile to build remove-promote-change-addr-space

scchan:


whchung:
merged even if one check failed.

PR #276, opened 2017-03-31
Use imported target for hsa to make find_package relocatable

pfultz2:
Currently, it was using an absolute path to hsa in the exported targets. This makes an imported target of hsa libraries, so that`hcc-cmake.config`(ie `find_package(hcc)`) can be relocatable.


whchung:
@pfultz2 the PR looks good to me. I'll take it even though CI tests show failure. Its execution log suggests it's a network I/O issue and not related with this PR.

whchung (approved):
LGTM

PR #281, opened 2017-04-06
Unpinned copy mode related fixes

gargrahul:
1. Enabled setting of unpinned copy mode through env variable
2. Guarded PinInPlace in case of host buffer already locked. In such case if choose best is the copy mode , it will fall back to staging copy logic. If copy mode is set to PinInPlace then exception will be thrown.

### DIFF
--- a/lib/hsa/unpinned_copy_engine.cpp
+++ b/lib/hsa/unpinned_copy_engine.cpp
@@ -225,7 +225,7 @@ void UnpinnedCopyEngine::CopyHostToDevice(UnpinnedCopyEngine::CopyMode copyMode,
         THROW_ERROR(hipErrorInvalidValue, HSA_STATUS_ERROR_INVALID_ARGUMENT);
     }
     tprintf (DB_COPY2, "Unpinned H2D: pointer type =%d\n", info.type);
-    if((info.type == 1) || (info.type == 2)) {
+    if((info.type == HSA_EXT_POINTER_TYPE_HSA) || (info.type == HSA_EXT_POINTER_TYPE_LOCKED)) {
         isLocked = true;
     }
     if (copyMode == ChooseBest) {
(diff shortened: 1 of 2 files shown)


whchung:
@gargrahul could you specify which HCC unit test, or HIP directed tests would be improved from this PR?

lib/hsa/unpinned_copy_engine.cpp
```diff
@@ -216,10 +216,22 @@ void UnpinnedCopyEngine::CopyHostToDeviceMemcpy(void* dst, const void* src, size
...
+    hsa_amd_pointer_info_t info;
+    hsa_status_t hsa_status;
+    bool isLocked = false;
+    const char *srcp = static_cast<const char*> (src);
+    hsa_status = hsa_amd_pointer_info(const_cast<char*> (srcp), &info, nullptr, nullptr, nullptr);
+    if(hsa_status != HSA_STATUS_SUCCESS) {
+        THROW_ERROR(hipErrorInvalidValue, HSA_STATUS_ERROR_INVALID_ARGUMENT);
+    }
+    tprintf (DB_COPY2, "Unpinned H2D: pointer type =%d\n", info.type);
+    if((info.type == 1) || (info.type == 2)) {
```
whchung: please use enum, such as HSA_EXT_POINTER_TYPE_HSA or HSA_EXT_POINTER_TYPE_LOCKED

whchung (changes requested):

whchung (changes requested):
In order to keep git commit history more comprehensible in the future, please revise the commit message so it better describes the purpose of the commit. The description used in this PR should be good.

whchung (approved):
LGTM

whchung:
@gargrahul I have to revert this PR because it breaks 16 HIP directed tests. Please make sure you submit test results in future PRs.

PR #284, opened 2017-04-07
[Cmake] pass LLVM_ENABLE_ASSERTIONS to the clang target

scchan:
Change-Id: I142876c963832668635dea739df561631f915514


whchung (approved):
LGTM

PR #287, opened 2017-04-11
feature_use_rocm_agent_enumerator_to_pick_gfxip

AlexVlx:
This handles the HCC side of enabling the use of the rocm_agent_detector to pick the target for AMDGPU codegen. It is matched with another PR against clang itself, and depends on adoption of the detector into rocm-smi.


scchan:
In CMakeLists.txt, could you add rocm-smi to the lines starting with "set(HCC_ROCR_DEP ..."?  That will add a dependency on rocm-smi to the installer package

scchan:
Combining with the patch in the driver https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/33 , I don't think I'm getting a clear idea on how this works and perhaps you could explain here...   

Let's take an example.  Suppose a user builds hcc on a system with a Hawaii GPU, then cmake would invoke the agent enumerator and set the default GPU target to gfx701?

Now if the user copies the hcc binaries to a system with a Fiji and compiles an application without specifying a GPU target.  I think in this case, hcc will use the default target, which is gfx701 set by cmake?  Or will the hcc driver invoke the agent enumerator to detect the targets available?

AlexVlx:
@scchan very good question, I should have clarified from the get-go, apologies:
1) yes, the highest GFXip found in the build system is set as the default (in hindsight, it may be equally valid to set all GFXips found in the build system as a default target list, although that could be inefficient wrt making compiles by default potentially expensive);
2) at runtime, there are 4 possible cases:
    a) the user manually passes --amdgpu-target=foo - then the detector is not run, and the default is not queried, and codegen is done for foo;
    b) the user does not pass --amdgpu-target - the detector is run, if it succeeds then all found GFXips are passed as --amdgpu-target(s); if it cannot find any valid target it bails out with an error message; if it does not run successfully for other reasons (not found, runtime error) it falls back to default; this is the "auto" case;
    c) the user has set the HCC_AMDGPU_TARGET environment variable - then (a) and (b) are overridden by the targets in the environment variable (this is current behaviour, I have not changed it - not sure if it is intentional);
    d) if neither (a) nor (b) nor (c) have produced --amdgpu-target(s) then the CMake default (in your example gfx701) would end up as the target - again, this is in line with current behaviour.
I would expect (b) to cover most common use cases, and between itself and (a) to get all but the quirkiest of scenarios. It was my understanding that this was more of less what we were after, but perhaps I am incorrect? Please let me know. Thank you.

scchan:
I'm thinking that maybe we should just set the default to gfx803 or any other sensible one in cmake (instead of invoking the detector tool).  That would have taken the build machine config out of the picture and we'll get a more predictable behavior.

whchung:
@AlexVlx may I understand the status of getting the detector tool into `https://github.com/RadeonOpenCompute/roc-smi` ? Just checked and it seems it's not there yet.

whchung:
@AlexVlx I'm for @scchan 's idea to just set the default to `gfx803` in CMake, without invoking the detector tool. This should decouple the configuration used by the build machine from the machine being deployed.

CMakeLists.txt
```diff
@@ -285,35 +292,75 @@ else ((NOT HSA_HEADER) OR (NOT HSA_LIBRARY))
...
 endif (HAS_ROCM EQUAL 0)
 
 #################
 # Detect AMDGPU backend for native codegen
 #################
 
-set(HSA_AMDGPU_GPU_TARGET "gfx803" CACHE STRING "Target AMDGPU ISA version")
+set(default_target "gfx803")
+
+# Attempt to detect any available HSA Agents, picking the highest ISA they
```
whchung: @AlexVlx I'm for @scchan 's idea to fix `default_target` to `gfx803` without invoking `rocm_agent_enumerator` at CMake time. This could help decouple the configuration used in the build machine from the machines being deployed.

whchung (changes requested):

bensander:
If detector tool is not present, can we default to 0x803?   (actually I thought his was already the plan?)
-ben

whchung:
@bensander `gfx803` would be the default value in CMake.

whchung (approved):
LGTM. Waiting for https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/33 be merged before taking this PR.

whchung:
@AlexVlx build check failure is a configuration issue not related to the PR itself. This PR will still be held for now until additional changes requested in https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/33 are fulfilled.

scchan:
One issue I found is that setting the cmake variable HSA_AMDGPU_GPU_TARGET to "auto" doesn't work.    That's probably because clang doesn't use HSA_AMDGPU_GPU_TARGET to set the default for --amdgpu-target.  Could you take a look at how to fix that?

scchan:
@AlexVlx do we still need this?

AlexVlx:
@scchan I think it is useful in its present form (I've slightly updated it), in conjunction with: https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/58.

whchung:
This PR is now obsolete.

PR #289, opened 2017-04-12
Fixed unpinned copy mode issue

gargrahul:
- Enabled Copy mode selection through env variable
- Guarded PinInPlace logic from being called with already locked pointer.
[TestResults.txt](https://github.com/RadeonOpenCompute/hcc/files/915560/TestResults.txt)


whchung (approved):
LGTM. Thanks for providing HIP test results.

PR #290, opened 2017-04-12
add backtrace support to the runtime

scchan:
Dump a backtrace when the hcc RT fails to launch a kernel.  If the host path is compiled with debug information, the addr2line tool could be used to map the PC back to source line.  

Example backtrace:

Backtrace:
0x00007fce6d54ee49:	Kalmar::HSADevice::CreateKernel(char const*) + 0x19
0x0000000000411770:	hc::completion_future hc::parallel_for_each<main::$_2>(hc::accelerator_view const&, hc::extent<1> const&, main::$_2 const&) + 0x410
0x00000000004092e0:	hc::completion_future hc::parallel_for_each<1, main::$_2>(hc::extent<1> const&, main::$_2 const&) + 0x40
0x0000000000408ee1:	main + 0xba1
0x00007fce6d798830:	__libc_start_main + 0xf0
0x0000000000404a29:	_start + 0x29


whchung:
@scchan very interesting usage of libunwind. I noticed there is only one place which dumps stack trace. is this intended, or would be improved later?

scchan:
I was planning to add to other places later but let's do it now

scchan:
@whchung Please hold off merging this PR.  I need to give folks managing the build farm a heads-up to install libunwind first otherwise this will break the build.

whchung:
@scchan I totally understand. in fact I wasn't really planning to take this PR until your green light because of additional dependency packages.

whchung:
@scchan please ping me when the CI infrastructure is ready to add `libunwind` as a new dependency.

docker/dockerfile-hcc-lc-ubuntu-16.04
```diff
@@ -6,7 +6,7 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install --no-instal
   curl -sL http://packages.amd.com/rocm/apt/debian/rocm.gpg.key | apt-key add - && \
   sh -c 'echo deb [arch=amd64] http://packages.amd.com/rocm/apt/debian/ xenial main > /etc/apt/sources.list.d/rocm.list' && \
   apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install --no-install-recommends -y \
-  hsa-rocr-dev g++-multilib gcc-multilib findutils libelf1 libpci3 file && \
+  hsa-rocr-dev g++-multilib gcc-multilib findutils libelf1 libpci3 file libunwind-dev && \
```
kknox: Instead of manually installing libunwind-dev to the docker install image, would it work better to make it a dependency on the debian package so that users benefit from package management installing the dependency?  I suppose same question for g++-multilib gcc-multilib.

scchan: make sense

PR #294, opened 2017-04-20
CodeXL Activity Logger profiling

ot-amd:
Added support for profiling using the CodeXL Activity Logger.
Profiling can be enabled and disabled using the environment variable HCC_CODEXL_PROFILING.
The profiler starts at the beginning of the HSAContext object and markers can be placed using a single line of code.
In an application compiled with HCC, only adding the hc_profile header file and the markers are required to do profiling.

As an example, a marker has been placed at the beginning of the HSAContext object. More markers will be placed in the future to profile other parts of HCC.


CMakeLists.txt
```diff
@@ -159,19 +159,19 @@ endif (HCC_INTEGRATE_ROCDL)
 ################
 # Detect CodeXL Activity Logger
 ################
-if (USE_CODEXL_ACTIVITY_LOGGER EQUAL 1)
-
 find_path(CODEXL_ACTIVITY_LOGGER_HEADER NAMES CXLActivityLogger.h PATHS ${CODEXL_ACTIVITY_LOGGER_HEADER_DIR} ${ROCM_ROOT}/profiler/CXLActivityLogger/include NO_DEFAULT_PATH)
 find_path(CODEXL_ACTIVITY_LOGGER_HEADER NAMES CXLActivityLogger.h)
 
 if (NOT CODEXL_ACTIVITY_LOGGER_HEADER)
+  MESSAGE("")
```
whchung: this line seems to be redundant.

CMakeLists.txt
```diff
@@ -159,19 +159,19 @@ endif (HCC_INTEGRATE_ROCDL)
...
   MESSAGE("CodeXL header not found. use -DCODEXL_ACTIVITY_LOGGER_HEADER_DIR=<path_to_CXLActivityLogger.h>.")
 endif (NOT CODEXL_ACTIVITY_LOGGER_HEADER)
 
 find_path(CODEXL_ACTIVITY_LOGGER_LIBRARY NAMES libCXLActivityLogger.so PATHS ${CODEXL_ACTIVITY_LOGGER_LIBRARY_DIR} ${ROCM_ROOT}/profiler/CXLActivityLogger/bin/x86_64 NO_DEFAULT_PATH)
 find_path(CODEXL_ACTIVITY_LOGGER_LIBRARY NAMES libCXLActivityLogger.so)
 
 if (NOT CODEXL_ACTIVITY_LOGGER_LIBRARY)
+  MESSAGE("")
```
whchung: this line seems to be redundant.

hcc_config/hcc_config.cpp
```diff
@@ -166,6 +166,8 @@ void ldflags(void) {
     std::cout << " -L" XSTR(CODEXL_ACTIVITY_LOGGER_LIBRARY);
     std::cout << " -lCXLActivityLogger";
 #endif
+
+    std::cout << " -lhc_profile";
```
whchung: @otauqir I'm not fond of the idea to add a mandatory library to be linked by every HCC application. Plus the implementation of `hc_profile` forces strong dependency to CodeXL library. I think it would be better to use dynamic shared library, and only load on demand.

whchung (changes requested):

lib/hsa/hc_profile.cpp
```diff
@@ -0,0 +1,29 @@
+#include "hc_profile.hpp"
+#include "CXLActivityLogger.h"
+
+cxlProfiler::cxlProfiler() {
+    if (getenv("HCC_CODEXL_PROFILING")) {
+        enabled = 1;
+        cxlMarker::enabled = 1;
+        amdtInitializeActivityLogger();
```
whchung: @otauqir static dependency to CodeXL is undesirable. It means every HCC application now would be linked with CodeXL library. Please consider improve the design to use dynamic libraries.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2799,6 +2800,9 @@ class HSAContext final : public KalmarContext
     */
     hsa_agent_t host;
 
+    cxlProfiler profiler;
+    cxlMarker contextMarker = CXL_MARKER_CLASS;
```
whchung, line 2804: @otauqir since the dependency to `hc_profile` is directly imposed here, why wouldn't `hc_profile` be simply placed inside `mcwamp_hsa` library, but to make it a standalone one, and has to be linked by `hcc-config --ldflags`? It could potentially make HCC applications more fragile, as *not* every HCC application are linked using `hcc-config`.

whchung (changes requested):

lib/hsa/mcwamp_hsa.cpp
scchan: @whchung we want to use this infrastructure to profile hcc RT itself.  The issue with putting it in mcwamp_hsa won't work well because then all the other libraries (i.e. libhc_am, libmcwamp) will  have to link against mcwamp_hsa.  
I think most of the code in hc_profile could be moved into a header; however, there's a class static  flag cxlMarker::enable that we need to put into a .cpp but it looks hcc doesn't have a .so that's being linked to by all the .so in hcc RT.   
We may need to create a new shared library for that.
If we build libhc_am, libmcwamp_hsa and friends and link them against hc_profile, then user won't need to worry about the new shared library or codexl's.

hcc_config/hcc_config.cpp
scchan: I'd suggest adding a cmake flag to control the enablement of hc_profile and let it be OFF by default for now.  Once the profiler package becomes available in a future release of rocm, we could consider turning it ON by default and make that a dependency for hcc.

bensander:
The profiler is a small user-mode library so not a significant dependency – think of it as an extension of HCC runtime lib but one that is shared with other language runtimes.

lib/hsa/mcwamp_hsa.cpp
whchung: @scchan If I understand you correctly, it seems:
- `hc_profile` is meant to be a proxy to encapsulate all interactions with CodeXL,
- and `hc_profile` can be used to provide `hc_am` and `mcwamp_hsa` themselves, plus HCC applications.

If that's the case I think it is fine to build `hc_profile` as s standalone library, *but* we should:
- make this a CMake option to be switched on/off at configure time
- in `hcc-config`, turn on/off `-lhc_profile` inside `--ldflags` based on the CMake option
- guard the dependency to `hc_profile` inside `hc_am` and `mcwamp_hsa` based on the CMake option
- in `hcc-config`, perhaps add another flag `--profile` to explicitly specify `-lhc_profile` for user applications

scchan:
@otauqir you also need to add a line in docker/dockerfile-build-ubuntu-16.04 to install the profiler package; otherwise the automated build will not be able this

lib/hsa/hc_profile.cpp
```diff
@@ -0,0 +1,40 @@
...
+}
+cxlProfiler::~cxlProfiler() {
+#ifdef CODEXL_PROFILING_ENABLED
+    if (enabled) {
+        amdtFinalizeActivityLogger();
+    }
+#endif
+}
+
+bool cxlMarker::enabled = false;
```
whchung, line 24: @otauqir can we try remove this field, so `hc_profile.cpp` is not needed at all, and we can place all the implementation inside `hc_profile.hpp`? for example, reading from an env var.

whchung: @otauqir this would further simply CMake and `hcc-config`. I'm really not in favor of introducing another link-time dependency.

whchung (changes requested):

whchung:
@otauqir ping

lib/hsa/hc_profile.cpp
ot-amd: A global enabled flag is used to not use getenv everything a marker is created.

The enabled flag is set when the profiler is initialized and every marker that is created after doesn't have to check if the env var is set or not.

We can remove the dependency of hc_profile from user applications by not adding it in hcc-config, but we can still profile the runtime if the env var is set.

PR #295, opened 2017-04-20
added option to rocdl to build hc lib & cleanup

dfukalov:
1. added option to rocdl to build hc lib
2. minor cleanup of unused include

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -414,6 +414,7 @@ if (HCC_INTEGRATE_ROCDL)
             -DLLVM_DIR=${CLANG_BIN_DIR}
             -DAMDHSACOD=${AMDHSACOD}
             -DGENERIC_IS_ZERO=ON
+            -DBUILD_HC_LIB=ON
     COMMAND make -j ${NUM_BUILD_THREADS} # not portable, but we need it this way
     WORKING_DIRECTORY ${ROCDL_BUILD_DIR}
     DEPENDS clang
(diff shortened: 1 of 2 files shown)


dfukalov:
the option is "on" by default in device library currently, but going to be changed to default value "off" in order to re-enable the library build with LC (not HCC)

CMakeLists.txt
```diff
@@ -414,6 +414,7 @@ if (HCC_INTEGRATE_ROCDL)
             -DLLVM_DIR=${CLANG_BIN_DIR}
             -DAMDHSACOD=${AMDHSACOD}
             -DGENERIC_IS_ZERO=ON
+            -DBUILD_HC_LIB=ON
```
whchung, line 417: @dfukalov there is no such flag `BUILD_HC_LIB` inside `ROCm-Device-Libs`, could you explain the rationale behind adding this line?

whchung (changes requested):

CMakeLists.txt
whchung: @dfukalov after checking your earlier comment I think I know the rationale now. I'll hold off this PR and merge it when `ROCm-Device-Libs` does carry this switch.

whchung (approved):

CMakeLists.txt
dfukalov: Yes, the switch was added in http://git.amd.com:8080/#/c/82189/ and is not arrived yet.
The will not brake build, however warning of unused variable will be produced.

dfukalov:
will merge this pull request once library change arrives to hcc tree

PR #296, opened 2017-04-20
Reenable printf

scchan:
- revert the patch that removes the hc_printf.hpp header
- replace the hsail atomic used in the implementation to work with the LC backend
- update the test to print from 2 different tiles

### DIFF
--- a/tests/Unit/HSA/printf.cpp
+++ b/tests/Unit/HSA/printf.cpp
@@ -1,0 +1,53 @@
+
+// RUN: %hc %s -lhc_am -o %t.out && %t.out | %FileCheck %s
+
+#include <hc.hpp>
+#include <hc_printf.hpp>
+
+#include <iostream>
+
+#define SIZE (32)
+#define PRINTF_BUFFER_SIZE (512)
+
+int main() {
+  using namespace hc;
+
+  accelerator acc = accelerator();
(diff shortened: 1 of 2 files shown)


tests/Unit/HSA/printf.cpp
```diff
@@ -0,0 +1,39 @@
...
+  processPrintfBuffer(printf_buf);
+
+  deletePrintfBuffer(printf_buf);
+
+  return 0;
+}
+
+// CHECK: Hello HC from thread: 0
+// CHECK: Hello again from thread: 0
+// CHECK: Hello HC from thread: 64
```
whchung: @scchan the revived printf functionality works, but this test fails on my box. from the output it seems the order of CHECK macro is incorrect. but I'm concerned it could be a non-deterministic behavior.

whchung (changes requested):

tests/Unit/HSA/printf.cpp
scchan: @whchung I changed the CHECK directives into CHECK-DAG so out-of-order outputs should work fine now

whchung (approved):

whchung:
@scchan this PR looks good now. will take it once the Jenkins build finishes.

PR #300, opened 2017-04-26
add tests for expf, fix bugs in exp test

scchan:
- we didn't cover expf in our math tests, adding them
- the cumulative threshold is too low even for one set of results in some cases for float


whchung (approved):
LGTM

PR #301, opened 2017-04-28
Remove pointer from tracker after unlock

aditya4d:
This change removes pointer from am tracker once the page-locked pointer is unlocked as it is allocated by user and pinned by runtime.


scchan:
LGTM

aditya4d:
Waiting for review from @bensander

PR #304, opened 2017-05-01
Remove entries from tracker on host_unlock.

bensander:
Fixes hip tests for host register functionality.
Aditya's fix + some extra error checking.

### DIFF
--- a/lib/hsa/hc_am.cpp
+++ b/lib/hsa/hc_am.cpp
@@ -518,10 +518,6 @@ am_status_t am_memory_host_unlock(hc::accelerator &ac, void *hostPtr)
             am_status = AM_ERROR_MISC;
         }
     }
-    am_status_t remove_status = am_memtracker_remove(hostPtr);
-    if(remove_status != AM_SUCCESS){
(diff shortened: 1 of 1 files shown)


lib/hsa/hc_am.cpp
```diff
@@ -512,6 +512,11 @@ am_status_t am_memory_host_unlock(hc::accelerator &ac, void *hostPtr)
     if(am_status == AM_SUCCESS)
     {
         hsa_status_t hsa_status = hsa_amd_memory_unlock(hostPtr);
+        if (hsa_status == HSA_STATUS_SUCCESS) {
+            am_status = am_memtracker_remove(hostPtr);
+        } else {
+            am_status = AM_ERROR_MISC;
+        }
     }
     am_status_t remove_status = am_memtracker_remove(hostPtr);
```
scchan: maybe this line should be removed since it has been moved to inside the if branch above?

bensander: Yes.  I updated pull request.

PR #306, opened 2017-05-08
Improve am_memtracker_print.

bensander:
If targetAddres=0 print entire table
else find&print entrty closest to targetAddr.

Print peer info from hsa_amd_pointer_info call.

Use smaller table-style format.

Add HC-level test


whchung (approved):
LGTM

whchung (approved):
LGTM

PR #316, opened 2017-05-15
Provide a cmake target for creating shared libraries

pfultz2:
I called it `hccshared` but perhaps someone can come up with a better name,


whchung:
@pfultz2 thanks. could you elaborate a bit more on how to use this `hccshared` from user perspective?

pfultz2:
Yea, so when I build a shared library in cmake I can do:

```cmake
find_package(hcc PATHS /opt/rocm)
add_library(foo SHARED foo.cpp)
target_link_libraries(foo hcc::hccrt hcc::hccshared)
```

PR #319, opened 2017-05-16
Fix create_marker to set dependency on previous asyncOp (if needed)

bensander:
Don't always rely on barrier bits for implicit ordering -
call detectStreamDeps to see if additional signal dependency is
required.

In particular if previous command was copy, creating a marker
or blocking_marker will now add the completion signal from the copy
to the dependendcy list.


whchung (approved):
LGTM.

PR #320, opened 2017-05-17
Fix finegrain

bensander:
Includes a couple bug fixes in fine-grain mem allocations


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -299,6 +299,12 @@ class HSAQueue;
 class HSADevice;
 } // namespace Kalmar
 
+
+// Global trackers for all HSA agents:
+// Possible these should move as member static in the future.
+static std::vector<hsa_agent_t > g_hsaAgents; // Array of the CPU + all GPU agents.
```
whchung: It seems `g_hsaAgents` is never read in any codes. Would future codes need this vector? If not we should consider remove it.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -299,6 +299,12 @@ class HSAQueue;
 class HSADevice;
 } // namespace Kalmar
 
+
+// Global trackers for all HSA agents:
+// Possible these should move as member static in the future.
+static std::vector<hsa_agent_t > g_hsaAgents; // Array of the CPU + all GPU agents. 
+static std::map<uint64_t, Kalmar::HSADevice *> g_agentToDeviceMap;
```
whchung: Could `g_agentToDeviceMap` be a member variable of `HSAContext`? It can be then queries by `HSADevice` via a member function.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2742,6 +2757,16 @@ class HSADevice final : public KalmarDevice
             executables[index] = new HSAExecutable(hsaExecutable, code_object);
         }
     }
+
+    static int get_seqnum_from_agent (hsa_agent_t hsaAgent)
```
whchung: This static member function is defined but no one really calls it.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2167,6 +2175,9 @@ class HSADevice final : public KalmarDevice
             throw Kalmar::runtime_exception("HCC_CHECK_COPY can only be used on machines where accelerator memory is visible to CPU (ie large-bar systems)", 0);
         }
 
+        g_agentToDeviceMap.insert(std::pair<uint64_t, HSADevice*> (agent.handle, this));
```
whchung: `g_agentToDeviceMap` is populated here, and would be queried by `HSADevice::get_seqnum_from_agent()`. But in the PR there is actually no one calling that static member function, thus making this map unused.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2167,6 +2175,9 @@ class HSADevice final : public KalmarDevice
...
+        g_hsaAgents.push_back(agent);
```
whchung: `g_hsaAgents` is populated here but is never used elsewhere in this PR.

whchung (changes requested):
The 2 new global variables shall be better reviewed & re-factored.
- `g_hsaAgents` seems can be abolished.
- `g_agentToDeviceMap` seems can be set as a member variables inside `HSAContext`.

lib/hsa/mcwamp_hsa.cpp
bensander: This is future functionality, will be used for some upcoming debug functionality to translate the hsaAgent handle into a human-readable device#.

bensander:
Updated pull request to address Jack feeedback

whchung (approved):

whchung:
@bensander The PR looks good now. Could you help resolve the merge conflicts so I can take it?

PR #322, opened 2017-05-18
Fix floating point exception due to empty kernel

yxsamliu:
Empty kernel has workitem_vgpr_count equal 0, which cause floating point exception in mcwamp_hsa.cpp when calculating
max_num_work_items_per_cu. The fix is let it be 1.


yxsamliu:
We need this after llvm merge with amd-common

PR #325, opened 2017-05-19
Improve info for number of VGPRS > HW limit error message.

bensander:
Show registers, kernel name (mangled), grid dims


whchung (approved):
LGTM

PR #327, opened 2017-05-19
[Jenkins] clone hcc with git-submodule

scchan:


Jenkinsfile
```diff
@@ -24,40 +24,11 @@ node ('rocmtest')
     deleteDir( )
```
pfultz2: You should probably remove this, so that cloning the entire repo doesn't happen on every build. Then you will need to remove the build directory before building, so the build artifacts won't cause problems.

whchung: @scchan could you assess if arguments from @pfultz2 is valid? I don't have way to measure time required on the jenkins job. only you guys have..

pfultz2: In Jenkins, it takes between 10 to 15 minutes to checkout the repo, thats longer than the build which is about 7 minutes.

scchan: It took 27 min in total and I think it's still acceptable.
The build currently fails when uploading the docker image to the artifact server.  It's unclear to me why though

pfultz2: It still could probably be faster by removing the `deleteDir`.

PR #332, opened 2017-05-23
Fix -gline-tables-only issue when building hcc RT

scchan:
Add -gline-tables-only only when building hcc RT in Release mode to prevent it from overriding other debug related compiler switches


whchung (approved):
LGTM

PR #333, opened 2017-05-24
Feature opt flush

bensander:
Passes all HC tests with HCC_OPT_FLUSH=0  or HCC_OPT_FLUSH=1

### DIFF
--- a/include/hc.hpp
+++ b/include/hc.hpp
@@ -1453,14 +1453,15 @@ accelerator_view::create_marker(memory_scope scope) const {
     std::shared_ptr<Kalmar::KalmarAsyncOp> deps[1]; 
     // If necessary create an explicit dependency on previous command
     // This is necessary for example if copy command is followed by marker - we need the marker to wait for the copy to complete.
-    std::shared_ptr<Kalmar::KalmarAsyncOp> depOp = pQueue->detectStreamDeps(hcCommandMarker, nullptr);
+    hc::memory_scope releaseScope = hc::no_scope;
+    std::shared_ptr<Kalmar::KalmarAsyncOp> depOp = pQueue->detectStreamDeps(hcCommandMarker, nullptr, &releaseScope);
 
     int cnt = 0;
     if (depOp) {
         deps[cnt++] = depOp; // retrieve async op associated with completion_future
     }
 
-    return completion_future(pQueue->EnqueueMarkerWithDependency(cnt, deps, scope));
+    return completion_future(pQueue->EnqueueMarkerWithDependency(cnt, deps, greater_scope(releaseScope, scope)));
 }
 
 inline unsigned int accelerator_view::get_version() const { return get_accelerator().get_version(); }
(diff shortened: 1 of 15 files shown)


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -41,6 +41,7 @@
 #include <time.h>
 #include <iomanip>
 
+#define KALMAR_DEBUG (0)
```
whchung: @bensander can you remove this line?

include/kalmar_runtime.h
```diff
@@ -245,7 +264,7 @@ class KalmarQueue
   /// enqueue marker with prior dependency
   virtual std::shared_ptr<KalmarAsyncOp> EnqueueMarkerWithDependency(int count, std::shared_ptr <KalmarAsyncOp> *depOps, memory_scope scope) { return nullptr; }
```
whchung, line 265: @bensander It seems the 3rd parameter, `scope` , is getting much more importance in this PR. Would you mind add more comments here to explain how `scope` works?

And looking at this PR I'm also wondering could this `scope` argument have a default value, say `hc::system_scope` ?

include/kalmar_runtime.h
```diff
@@ -245,7 +264,7 @@ class KalmarQueue
...
 
-  virtual std::shared_ptr<KalmarAsyncOp> detectStreamDeps(hcCommandKind commandKind, KalmarAsyncOp *newCopyOp) { return nullptr; };
+  virtual std::shared_ptr<KalmarAsyncOp> detectStreamDeps(hcCommandKind commandKind, KalmarAsyncOp *newCopyOp, memory_scope *requiredRelease) { return nullptr; };
```
whchung, line 267: @bensander It seems the 3rd parameter, `requiredRelease`, is getting much more importance in this PR. Would you mind add more comments here to explain how the function works?

And would it be possible to change `requiredRelease` have be optional, like `memory_scope *requiredRelease = nullptr`? Because it seems to me not all call sites of `detectStreamDeps` really cares about the `memory_scope` changed by it.

include/hc.hpp
```diff
@@ -1470,7 +1471,8 @@ inline completion_future accelerator_view::create_blocking_marker(completion_fut
 
     // If necessary create an explicit dependency on previous command
     // This is necessary for example if copy command is followed by marker - we need the marker to wait for the copy to complete.
-    std::shared_ptr<Kalmar::KalmarAsyncOp> depOp = pQueue->detectStreamDeps(hcCommandMarker, nullptr);
+    hc::memory_scope releaseScope = hc::no_scope;
+    std::shared_ptr<Kalmar::KalmarAsyncOp> depOp = pQueue->detectStreamDeps(hcCommandMarker, nullptr, &releaseScope);
```
whchung: @bensander this is a place where we can consider make the 3rd parameter of `detectStreamDeps` be optional.

include/hc.hpp
```diff
@@ -1493,7 +1495,8 @@ accelerator_view::create_blocking_marker(InputIterator first, InputIterator last
...
```
whchung: @bensander this is a place where we can consider make the 3rd parameter of detectStreamDeps be optional.

whchung (changes requested):
@bensander This is an important PR. I'd like to propose a few changes before getting it merged. Besides more comments on the crucial APIs `detectStreamDeps` and `EnqueueMarkerWithDependency`, some arguments could also be considered made to be optional IMHO.

bensander:
Jack – thanks for review!  I submitted an updated request.  The hc.hpp issues below should have called greater_scope -  I made that change.  Left releaseScope as a mandatory argument since this is internal function and we want called to explicitly and carefully choose their release scope.

whchung (approved):
Thanks for the updated commit and clarifications. The PR now looks good to me.

PR #336, opened 2017-05-24
Overhaul math functions

AlexVlx:


whchung:
@AlexVlx could you shed some lights on why some tests are marked with `DISABLED_PENDING_REMOVAL` macro?

scchan (changes requested):
could you update the rocdl submodule link?

AlexVlx:
@whchung the DISABLED_PENDING_REMOVAL bit is there because I am removing the min and max overloads for floating point types that we were previously (interestingly) exposing, and which we should not quite provide (they are non-standard and can yield quirky interactions with standard or user defined functionality). However, perhaps there exists a 1 million dollar use case for them (I have very serious doubts :) ), case in the removal would need to be reverted and the tests residing under that guard re-instated.
Edit: @scchan, seems like you have the same question.

AlexVlx:
@scchan please see commit 89cdf7f5e0f6ee951706feb3945ca19e33bfbf9c. Thank you.

scchan (approved):

PR #339, opened 2017-05-25
Remove delete dir from jenkins file

pfultz2:
This will delete the build directory instead. This speeds up checkout especially with submodules.


scchan:
I've experimented with removing deleteDir() but it was necessary to get cloning of submodule to work especially for PRs that updates one of the submodule links

PR #340, opened 2017-05-26
Adding a try-catch guard around artifactory pushes

kknox:
Changing cmake to generate deb & rpm packages by default

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -541,20 +541,16 @@ set(PACKAGE_DESCRIPTION "HCC: An Open Source, Optimizing C++ Compiler for Hetero
 
 set(OFFICIAL_RELEASE_BUILD 1)
 
-option(CPACK_GENERATOR "Specify the installer package generator")
-
-# If no CPACK generator was specified, then set the default generator according to the distro 
-if (NOT CPACK_GENERATOR)
-    if ( "${DISTRO_ID}" MATCHES "ubuntu" )
-    set(CPACK_GENERATOR "DEB")
-  elseif ("${DISTRO_ID}" MATCHES "fedora")
-    set(CPACK_GENERATOR "RPM")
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -541,20 +541,9 @@ set(PACKAGE_DESCRIPTION "HCC: An Open Source, Optimizing C++ Compiler for Hetero
...
-    set(CPACK_GENERATOR "RPM")
-  else ()
-    # generate a tarball for unknown DISTRO_ID
-    set(CPACK_GENERATOR "TGZ")
-  endif ()
-endif (NOT CPACK_GENERATOR)
-
+# Default to generate both package types; users can override cache settings on command line with -D
+# RPM packages require RPM package on debian disto's
+set( CPACK_GENERATOR "DEB;RPM" CACHE STRING "cpack list: 7Z, DEB, IFW, NSIS, NSIS64, RPM, STGZ, TBZ2, TGZ, TXZ, TZ, ZIP" )
```
scchan: We've done some experiment and found out that the rpm generated on Ubuntu doesn't work on Fedora due to library version conflicts so I'd say that the rpm is probably not that useful.
This would also force users building on Fedora to generate a .deb package as well.

scchan (changes requested):

CMakeLists.txt
kknox: that's too bad; i'd like to start automating the generation of RPM's, but maybe we need to create a fedora dockerfile to build in.  I'll revert this change, but it will still look a bit different because CPACK_GENERATOR should not be an `option` like the way it was before.

kknox:
Fixed

scchan (approved):

scchan:
Thanks for the contribution!

PR #341, opened 2017-05-26
Fix cross queue flush

bensander:


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4264,16 +4282,13 @@ static Kalmar::hcCommandKind resolveMemcpyDirection(bool srcInDeviceMem, bool ds
 }
 
 inline hsa_status_t
-HSACopy::enqueueAsyncCopyCommand(Kalmar::HSAQueue* hsaQueue, const Kalmar::HSADevice *copyDevice, const hc::AmPointerInfo &srcPtrInfo, const hc::AmPointerInfo &dstPtrInfo) {
+HSACopy::enqueueAsyncCopyCommand(Kalmar::HSAQueue* hsaQueue2, const Kalmar::HSADevice *copyDevice, const hc::AmPointerInfo &srcPtrInfo, const hc::AmPointerInfo &dstPtrInfo) {
```
scchan: It seems that the HSAQueue being passed in has become redundant, could we remove it from the signature?

Other classes derived from KalmarAsyncOps such as HSABarrier has similar issues

lib/mcwamp.cpp
```diff
@@ -357,7 +357,7 @@ void BuildProgram(KalmarQueue* pQueue) {
 
 // used in parallel_for_each.h
 void *CreateKernel(std::string s, KalmarQueue* pQueue) {
-  return pQueue->getDev()->CreateKernel(s.c_str());
+  return pQueue->getDev()->CreateKernel(s.c_str(), pQueue);
```
scchan: `CreateKernel` should probably be splitted into 2 functions:

1- the queue object should have a member function `CreateDispatch(s.c_str())` that creates and returns a dispatch
2- In `CreateDispatch`, it would call `getDev()->CreateKernel(s.c_str())`, which will only be responsible for loading a kernel from object.  

This line would be simplified into `pQueue->CreateDispatch(s.c_str())`

tests/Unit/AcceleratorViewCopy/copy_coherency.cpp
```diff
@@ -48,19 +49,53 @@ int main()
...
+        // Set B to -42, followed immediately by setting to expected value:
+        memsetIntKernel(av0, B, -42, numElements);
+        memsetIntKernel(av0, B, bexpected, numElements);
 
-    av.wait();
-    check(Bh, numElements, binit);
+        // Async copy back to host with no intervening code - 
+        // This should fail unless we properly issue a release fence after the memset:
+        av0.copy_async(B, Bh, sizeElements); 
+
```
scchan, line 68: Let's say if we change this example to use pinned memory and we remove the copy_async.  If we read directly  from the  memory on the host after memset, would it require a release fence?  If yes, would the runtime detects it?

bensander: If memory is pinned and read on host as you describe, then we would need a system-scope release to detect this.  The HCC runtime will insert a system-scope release on for accelerator_view::wait() or completion_future::wait().   This is tracked by _needs_sys_release (set if a kernel is dispatched to a qeueue which may dirty the L2) and the necessary release fence is determined by detectStreamDeps.  I will expand the test to include this case.

lib/hsa/mcwamp_hsa.cpp
bensander: I added a TODO do this as a future cleanup step, we can address on next rainy day.

lib/mcwamp.cpp
bensander: I added a TODO for this as future cleanup item.

lib/hsa/mcwamp_hsa.cpp
bensander: I knew someone was going to ask for this :)  I cleaned this up and updated the pull request, please take a look.

whchung:
@bensander there are build errors in this PR. could you help look into it?

PR #343, opened 2017-05-26
Add hcc-extra-libs flag to clamp-link

pfultz2:
This is so `clamp-link` can add extra libraries without needing an environment variable. Of course this PR [here](https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/51) is needed so this flag can be passed to the hcc driver.

### DIFF
--- a/lib/clamp-device.in
+++ b/lib/clamp-device.in
@@ -95,6 +95,10 @@ do
     AMDGPU_TARGET="${ARG#*=}"
     continue
     ;;
+    --hcc-extra-libs=*)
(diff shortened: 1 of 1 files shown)


whchung (changes requested):
@pfultz2 could you help NOT add submodule changes in this PR? I'll do that after I merge the PR in `hcc-clang-upgrade`

pfultz2:
So, I've updated it to not include the submodule.

whchung (approved):
LGTM

PR #344, opened 2017-05-26
Db and resources

bensander:
Move to runtime debug &&
Increase initial resource allocation sizes


whchung (approved):
Thanks for improving the runtime debugging facility!

PR #345, opened 2017-05-27
Enable HCC_OPT_FLUSH=1

bensander:
Modify am_alloc test to show how to use different synchronization
mechanisms for coherent and non-coherent host memory.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -97,7 +97,7 @@ long int HCC_D2H_PININPLACE_THRESHOLD = 1024;
 int HCC_SERIALIZE_KERNEL = 0;
 int HCC_SERIALIZE_COPY = 0;
 
-int HCC_OPT_FLUSH=0;
+int HCC_OPT_FLUSH=1;
 
 
 unsigned HCC_DB = 0;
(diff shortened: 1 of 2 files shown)


whchung:
@bensander could you elaborate a bit more on the potential impacts to existing HCC applications? I expect this change would mean HCC runtime would behave differently for *ALL* HCC apps, right?

bensander:
Today’s HCC adds a system-scope acquire and release around each kernel dispatch.  This works but generates a flush operation on the GPU L2 cache which can 3-5us to complete.   With this change, flushes are deferred and only performed “as-needed”.  The work over the last weeks has been to implement and test the different cases.  A brief summary of how the design works with HCC_OPT_FLUSH=1 :


-          Kernel commands perform agent-scope acquire and release, which on ROCM1.5 are very fast.

-          A system-scope release is performed before a D2H copy command. This ensures the GPU-side DMA engine sees the latest copy of the data.

-          accelerator_view::wait() performs a system-scope release.  Notably this allows the host to see any non-coherent pinned host memory written by the kernel.

-          HCC create_marker and create_blocking_marker support an optional release_scope parameter which allow the user to control the release fence:

o    Agent-scope for speed

o    System-scope if data needs to be made visible to other agents in the system

-          One caveat: There are 2 changes we need to make for multi-GPU P2P coherency which will be in an upcoming pull request.

Overall most application should benefit from the optimization with no impact to functionality.  Apps which are not correctly synchronized may need to be modified to ensure the system-scope release issues using one of the mechanism above.  For example it used to work to launch a PFE and wait on host for a while – now this requires use of one of the above synchronization commands.   We have done extensive testing and not seen issues, so believe it is time to enable.

-ben

whchung:
@bensander this PR fails `cmake-test.cpp` in `cmake-tests` directory. You can find the log on Jenkins CI. Could you help look into it?

whchung:
@bensander to my understanding you would amend this PR to address issues found in `array_view`, correct?

bensander:
Correct, and merge to tip as well.  Will be a new pull request.

whchung:
@bensander should i close this one and focus on #353 ?

bensander:
Yes

On Jun 2, 2017, at 11:05 PM, Wen-Heng (Jack) Chung <notifications@github.com<mailto:notifications@github.com>> wrote:


@bensander<https://github.com/bensander> should i close this one and focus on #353<https://github.com/RadeonOpenCompute/hcc/pull/353> ?

PR #347, opened 2017-05-31
Add pinned_vector<T>, a std::vector<T> with host pinned memory

rwvo:
This pull request adds a hc::pinned_vector<T>, which is an alias for
std::vector<T, am_allocator>. The purpose of hc::pinned_vector<T> is to speed
up data transfers between host and device automagically. On our test platform
(Intel Core i7-6700K, Radeon R9 Nano), speedup is ~30% compared to data transfers
from non-pinned memory.

The underlying memory for a pinned_vector<T> is allocated with
am_alloc() using the amHostPinned flag. This allows for fast DMA copies
from a pinned_vector<T> on the host to the device and vice versa
without the need for implicitly or explicitly creating intermediate
pinned buffers on the host. The destructor of pinned_vector<T> uses
am_allocator::deallocate() to properly release the pinned memory with
am_free().

### DIFF
--- a/include/pinned_vector.hpp
+++ b/include/pinned_vector.hpp
@@ -1,0 +1,52 @@
+#pragma once
+
+#ifndef _PINNED_VECTOR_H
+#define _PINNED_VECTOR_H
+
+#include <new>
+#include "hc.hpp"
+#include "hc_am.hpp"
+
+namespace hc
+{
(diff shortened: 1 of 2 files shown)


whchung (approved):
LGTM. @bensander ?

PR #348, opened 2017-06-01
Support for hc::abort through HSAQueue callback

gargrahul:
Added abort function. Mapped to __builtin_trap.

Please find attached the HCC directed test report.
[HCC_DirectedTest_Abort.txt](https://github.com/RadeonOpenCompute/hcc/files/1044480/HCC_DirectedTest_Abort.txt)


whchung (changes requested):
Please help provide a test case. In this PR you are introducing an new API in user visible interface. Please create a new test case which actually use it.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -787,13 +787,17 @@ namespace Kalmar {
...
+    }
+
     RocrQueue(hsa_agent_t agent, size_t queue_size, HSAQueue *hccQueue) 
     {
 
         assert(queue_size != 0);
 
         /// Create a queue using the maximum size.
-        hsa_status_t status = hsa_queue_create(agent, queue_size, HSA_QUEUE_TYPE_SINGLE, NULL, NULL,
+        hsa_status_t status = hsa_queue_create(agent, queue_size, HSA_QUEUE_TYPE_SINGLE, callbackQueue /*NULL*/, NULL,
```
whchung: @gargrahul Please remove `/*NULL*/` here.

whchung (changes requested):

include/hc.hpp
```diff
@@ -3202,7 +3202,9 @@ inline int __mad24(int x, int y, int z) [[hc]] {
   return __mul24(x,y) + z;
 }
 
-
+extern "C" inline void abort() __HC__ {
```
scchan: we don't need extern "C" here since this abort function is in the hc namespace

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -787,13 +787,17 @@ namespace Kalmar {
 // HSAQueue is the implementation of accelerator_view for HSA back-and.  HSAQueue
 // points to RocrQueue, or to nullptr if the HSAQueue is not currently attached to a RocrQueue.
 struct RocrQueue {
+    static void callbackQueue(hsa_status_t status, hsa_queue_t* queue, void *data) {
+        STATUS_CHECK(status, __LINE__);
```
scchan, line 791: Could you described how we could distinguish different abort conditions (i.e. vm fault, user abort(), etc..)?

scchan (changes requested):

gargrahul:
*@whchung *I have added a test case in 2nd commit.

tests/Unit/HC/abort.cpp
```diff
@@ -0,0 +1,93 @@
+
+// RUN: %hc %s -o %t.out && %t.out
+
```
scchan: This test is expected to abort during execution but it would be counted as a test failure if it's run as is.
The proper way would be to pipe stdout and stderr into the FileCheck tool to look for the expected error message.  You'll find examples with FileCheck in hcc/tests/Unit

scchan (changes requested):

tests/Unit/HC/abort.cpp
```diff
@@ -0,0 +1,93 @@
...
+// RUN: %hc %s -o %t.out && %t.out
```
whchung: @gargrahul the test is incorrect as @scchan pointed out. Once you create a test you should run it with `make test` to make sure it works.

whchung (changes requested):
the test is invalid.

scchan (approved):

scchan:
@whchung could you review and approve?

whchung (approved):

PR #349, opened 2017-06-02
[Jenkins]don't override hcc install path

scchan:


pfultz2:
This fails because it isn't installed into /opt/rocm.

PR #352, opened 2017-06-02
Use rocm cmake modules

pfultz2:
This is still a WIP. This will use the rocm cmake modules. The installation will now install the symlinks instead of doing it in the pre/post packaging script, which means the installation will be the same as installing with the package.


whchung:
@pfultz2 could you check the errors in Jenkins build? even though hcc is successfully built and packaged, but it seems to have something related to the changes in this PR so the package isn't usable in a Docker container. Also, could you help rebase the PR so it aligns with the tip of `clang_tot_upgrade` branch?

whchung:
@pfultz2 could you remove submodule changes from this PR? i believe only CMake files within `hcc` are really changed, and other submodules shall be as-is.

pfultz2:
I didn't change the submodules.

whchung:
@pfultz2 you may add `git add clang compiler lld` unknowingly. or could you rebase this PR with the current tip?

pfultz2:
I just needed to merge from upstream. Now there is no conflicts.

pfultz2:
But it shows that the submodules are changed.

pfultz2:
Ok, now they are synced with clang_tot_upgrade

PR #353, opened 2017-06-03
Enable opt flush3

bensander:
********************
Testing Time: 256.61s
********************
Failing Tests (4):
    HCC :: Unit/AsyncPFE/async_av_dependent7.cpp

  Expected Passes    : 721
  Expected Failures  : 30
  Unexpected Failures: 4

### DIFF
--- a/include/hc.hpp
+++ b/include/hc.hpp
@@ -1046,7 +1046,7 @@ public:
      * Check if @p other is peer of this accelerator.
      *
      * @return true if other can access this accelerator's device memory pool or false if not.
-     * the acceleratos is its own peer.
+     * The acceleratos is not its own peer.
      */
     bool get_is_peer(const accelerator& other) const {
         return pDev->is_peer(other.pDev);
--- a/include/hc.hpp
+++ b/include/hc.hpp
@@ -275,6 +275,9 @@ public:
      *   - accelerator_scope: Memory is released to the accelerator scope where the marker executes.
      *   - system_scope: Memory is released to system scope (all accelerators including CPUs)
      *
+     * dependent_futures may be recorded in another queue or another accelerator.  If in another accelerator,
+     * the runtime performs cross-accelerator sychronization.  
+     *
      * @return A future which can be waited on, and will block until the
      *         current batch of commands, plus the dependent event have
      *         been completed.
@@ -1453,15 +1456,14 @@ accelerator_view::create_marker(memory_scope scope) const {
     std::shared_ptr<Kalmar::KalmarAsyncOp> deps[1]; 
     // If necessary create an explicit dependency on previous command
     // This is necessary for example if copy command is followed by marker - we need the marker to wait for the copy to complete.
-    hc::memory_scope depReleaseScope = hc::no_scope;
-    std::shared_ptr<Kalmar::KalmarAsyncOp> depOp = pQueue->detectStreamDeps(hcCommandMarker, nullptr, &depReleaseScope);
+    std::shared_ptr<Kalmar::KalmarAsyncOp> depOp = pQueue->detectStreamDeps(hcCommandMarker, nullptr);
 
     int cnt = 0;
     if (depOp) {
         deps[cnt++] = depOp; // retrieve async op associated with completion_future
     }
 
-    return completion_future(pQueue->EnqueueMarkerWithDependency(cnt, deps, greater_scope(depReleaseScope, scope)));
+    return completion_future(pQueue->EnqueueMarkerWithDependency(cnt, deps, scope));
 }
 
 inline unsigned int accelerator_view::get_version() const { return get_accelerator().get_version(); }
@@ -1471,8 +1473,7 @@ inline completion_future accelerator_view::create_blocking_marker(completion_fut
 
     // If necessary create an explicit dependency on previous command
     // This is necessary for example if copy command is followed by marker - we need the marker to wait for the copy to complete.
-    hc::memory_scope depReleaseScope = hc::no_scope;
-    std::shared_ptr<Kalmar::KalmarAsyncOp> depOp = pQueue->detectStreamDeps(hcCommandMarker, nullptr, &depReleaseScope);
+    std::shared_ptr<Kalmar::KalmarAsyncOp> depOp = pQueue->detectStreamDeps(hcCommandMarker, nullptr);
 
     int cnt = 0;
     if (depOp) {
@@ -1483,7 +1484,7 @@ inline completion_future accelerator_view::create_blocking_marker(completion_fut
         deps[cnt++] = dependent_future.__asyncOp; // retrieve async op associated with completion_future
     } 
     
-    return completion_future(pQueue->EnqueueMarkerWithDependency(cnt, deps, greater_scope(depReleaseScope, scope)));
+    return completion_future(pQueue->EnqueueMarkerWithDependency(cnt, deps, scope));
 }
 
 template<typename InputIterator>
@@ -1495,8 +1496,7 @@ accelerator_view::create_blocking_marker(InputIterator first, InputIterator last
 
     // If necessary create an explicit dependency on previous command
     // This is necessary for example if copy command is followed by marker - we need the marker to wait for the copy to complete.
-    hc::memory_scope depReleaseScope = hc::no_scope;
-    std::shared_ptr<Kalmar::KalmarAsyncOp> depOp = pQueue->detectStreamDeps(hcCommandMarker, nullptr, &depReleaseScope);
+    std::shared_ptr<Kalmar::KalmarAsyncOp> depOp = pQueue->detectStreamDeps(hcCommandMarker, nullptr);
 
     int cnt = 0;
     if (depOp) {
@@ -1512,14 +1512,14 @@ accelerator_view::create_blocking_marker(InputIterator first, InputIterator last
         if (iter->__asyncOp) {
             deps[cnt++] = iter->__asyncOp; // retrieve async op associated with completion_future
             if (cnt == 5) {
-                lastMarker = completion_future(pQueue->EnqueueMarkerWithDependency(cnt, deps, greater_scope(depReleaseScope, scope)));
+                lastMarker = completion_future(pQueue->EnqueueMarkerWithDependency(cnt, deps, hc::no_scope));
                 cnt = 0;
             }
         }
     }
 
     if (cnt) {
-        lastMarker = completion_future(pQueue->EnqueueMarkerWithDependency(cnt, deps, greater_scope(depReleaseScope, scope)));
+        lastMarker = completion_future(pQueue->EnqueueMarkerWithDependency(cnt, deps, scope));
     }
 
     return lastMarker;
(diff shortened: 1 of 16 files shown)


include/hc.hpp
```diff
@@ -1453,15 +1456,14 @@ accelerator_view::create_marker(memory_scope scope) const {
     std::shared_ptr<Kalmar::KalmarAsyncOp> deps[1]; 
     // If necessary create an explicit dependency on previous command
     // This is necessary for example if copy command is followed by marker - we need the marker to wait for the copy to complete.
-    hc::memory_scope depReleaseScope = hc::no_scope;
-    std::shared_ptr<Kalmar::KalmarAsyncOp> depOp = pQueue->detectStreamDeps(hcCommandMarker, nullptr, &depReleaseScope);
+    std::shared_ptr<Kalmar::KalmarAsyncOp> depOp = pQueue->detectStreamDeps(hcCommandMarker, nullptr);
```
whchung, line 1459: @bensander this is one of the greatest departure from the previous implementation. could you help clarify why we don't need `depReleaseScope` here, and why we don't need `greater_scope()` later in this function?

include/kalmar_runtime.h
```diff
@@ -922,6 +923,10 @@ struct rw_info
         /// 2. Because the data pointer cannout be released, erase itself from devs
         if (HostPtr)
             synchronize(false);
+        if (curr) {
+            // Wait issues a system-scope release:
+            curr->wait();
```
whchung: @bensander this is the most important change to tame `array_view`-based tests, right?

tests/Unit/AcceleratorViewCopy/copy_coherency.cpp
```diff
@@ -94,23 +114,157 @@ int main()
...
+    int numElements = 20000000;
+
+    const size_t sizeElements = numElements * sizeof(int);
+    printf ("info: buffer size = %6.2f MB\n", sizeElements / 1024.0 / 1024.0);
+
+    if (1) {
+        singleAccelerator(numElements);
+    }
+
+    if (0) {
```
whchung: @bensander this is intended, right?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1040,10 +1067,9 @@ class HSAQueue final : public KalmarQueue
 
 
     void waitForStreamDeps (KalmarAsyncOp *newOp) {
-        hc::memory_scope releaseScope;
-        std::shared_ptr<KalmarAsyncOp> depOp = detectStreamDeps(newOp->getCommandKind(), newOp, &releaseScope);
+        std::shared_ptr<KalmarAsyncOp> depOp = detectStreamDeps(newOp->getCommandKind(), newOp);
         if (depOp != nullptr) {
-            EnqueueMarkerWithDependency(1, &depOp, HCC_OPT_FLUSH ? releaseScope : hc::system_scope);
+            EnqueueMarkerWithDependency(1, &depOp, HCC_OPT_FLUSH ? hc::no_scope : hc::system_scope);
```
whchung, line 1073: @bensander the greatest difference from the previous impl is here. as `HCC_OPT_FLUSH` is by default set as 1, marker with scope `hc::no_scope` would be used. but in previous version `releaseScope` is used. could you help clarify about the justifcation?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1088,30 +1117,13 @@ class HSAQueue final : public KalmarQueue
...
-                    // In the code below, this will be the first op waited on
-                    auto marker = EnqueueMarker();
-                    DBOUT("oldest AsyncOp has no signal - enqueue marker "<< marker<<"\n");
-                }
-                break;
-            }
-        }
-#endif
-        if (HCC_OPT_FLUSH && _needs_sys_release) {
+        if (HCC_OPT_FLUSH && _nextSyncNeedsSysRelease) {
```
whchung: @bensander This part seems to be the same as `releaseToSystemIfNeeded()`. Please try use the function istead of duplicated logic here.

tests/Unit/AsyncPFE/async_av_dependent7.cpp
```diff
@@ -8,14 +8,13 @@
 // loop to deliberately slow down kernel execution
 #define LOOP_COUNT (10240)
 
-#define TEST_DEBUG (0)
+#define TEST_DEBUG (1)
```
whchung: @bensander maybe we should turn this back to 0?

tests/Unit/AsyncPFE/async_av_dependent7.cpp
```diff
@@ -85,30 +84,28 @@ bool test1D() {
...
+  test1D<32, 16>();
+  assert(av.get_pending_async_ops() == 0);
+  test1D<64, 8>();
+  assert(av.get_pending_async_ops() == 0);
+  test1D<128, 32>();
+  assert(av.get_pending_async_ops() == 0);
+  test1D<256, 64>();
+  assert(av.get_pending_async_ops() == 0);
+  test1D<1024, 256>();
+  assert(av.get_pending_async_ops() == 0);
```
whchung, line 108: @bensander could you clarify the rationale to change from AND-comparison to assert() is to really check the pending async op counter only in debug mode?

i believe the most important reason is that we could now add more markers into the queue in different situations and that could change the number of async ops in the queue.

in other tests you added `HCC_OPT_FLUSH` and use the flag to determine the proper expected pending async ops when `HCC_OPT_FLUSH` is on, but not in this test.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -494,15 +499,18 @@ class HSABarrier : public Kalmar::KalmarAsyncOp {
...
 
     // prior dependencies
     // maximum up to 5 prior dependencies could be associated with one
     // HSABarrier instance
     int depCount;
 
-
 public:
+    uint16_t  header;
+    hc::memory_scope _acquire_scope;
```
whchung: @bensander we could use some documentation here for `_acquire_scope`, but later on in `enqueuemarker` we would take another parameter `releaseScope`. The difference between them deserve to be logged.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -494,15 +499,18 @@ class HSABarrier : public Kalmar::KalmarAsyncOp {
...
     std::shared_future<void>* future;
 
     // prior dependencies
     // maximum up to 5 prior dependencies could be associated with one
     // HSABarrier instance
     int depCount;
 
-
 public:
+    uint16_t  header;
```
whchung: @bensander comment should be added that this field holds the header bits AQL packet to be constructed

whchung (changes requested):
@bensander I'm largely fine with this important PR, but some clarifications are do required.

- why `greaterScope` isn't required anymore, and the change of `detectStreamDeps`
- differences between `_acquire_scope` and `releaseScope`
- changes to existing HCC tests

include/hc.hpp
bensander: Yes.  SDMA copies do not probe the L2 cache so we need to issues a system-scope release before a copy if any kernels have written L2.   The new design does this check before sending the copy.  The previous design allowed release-scope for markers to be "upgraded" to system-release - removing this simplifies the implementation and the create*marker APIs are free from the upgrade baggage and complexity.

include/kalmar_runtime.h
bensander: Yes.  The case was that an array_view destructor might deallocate the memory while it was still dirty in the L2 cache.  When L2 was later released, data was written to stale physical addresses and GPUVM faults could occur.  this check assures we flush L2 before deallocating the memory.  

At some point we should refactor the array_view copy routines so they share common code with the accelerator_view::copy routines appropriately.

tests/Unit/AcceleratorViewCopy/copy_coherency.cpp
bensander: Yes.  We need to review and later enable this test.

lib/hsa/mcwamp_hsa.cpp
whchung: Also it seems this field `header` is added mostly for debugging purpose, could we make it a private member variable and use accessor functions to get its value?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1580,16 +1611,32 @@ class HSAQueue final : public KalmarQueue
...
+                    if (depHSAQueue->nextSyncNeedsSysRelease()) {
+                        setNextSyncNeedsSysRelease(true);
                     }
+
+                    if (depHSAQueue->getHSADev() != this->getHSADev()) {
+                        // Cross-accelerator dependency case.
+                        // This requires system-scope acquire 
+                        // TODO - only needed if these are peer GPUs, could optimize with an extra check
+                        DBOUT(DB_WAIT, "  Adding cross-accelerator system-scope acquire\n");
+                        barrier->_acquire_scope = hc::system_scope;
```
whchung: @bensander this seems to be the only place which writes to `HSABarrier::_acquire_scope` outside `HSABarrier`. Perhaps we should change it to an accessor member function?

lib/hsa/mcwamp_hsa.cpp
whchung: also would it be better if `_acquire_scope` is a private member variable, and create accessor functions to get/set it?

lib/hsa/mcwamp_hsa.cpp
bensander: Correct, this is a key part of the change.   waitForStreamDeps is the HCC-internal function used to wait for commands in the same stream., specifically used for kernels or markers that need to wait for copy operations.    With HCC_OPT_FLUSH=1 now tracking and providing flushes in the appropriate places through other mechanisms, this marker is only needed for ordering and not cache flushing and thus can use the faster release synchronization.

tests/Unit/AsyncPFE/async_av_dependent7.cpp
bensander: assert fires in both release and debug mode. The intent of this change was so the test would exit at precisely the point where the fails (with an assertion and perhaps a back-trace if in debugger) rather than an error code when test exits.

lib/hsa/mcwamp_hsa.cpp
bensander: good point, will make this change.

tests/Unit/AsyncPFE/async_av_dependent7.cpp
bensander: Also I only added HCC_OPT_FLUSH define to tests where the HCC_OPT_FLUSH=1 changed the number of expected ops.

lib/hsa/mcwamp_hsa.cpp
bensander: Yes good point.

whchung (approved):
LGTM

PR #355, opened 2017-06-09
Fixes testcases for no peer accelerators

aaronenyeshi:
When checking size of peers, cannot count self as a peer, therefore must handle when size of peers is zero.


whchung:
@aaronenyeshi you are changing the tests , but are they failing before? could you provide the result of `make test` before and after your proposed changes?

aaronenyeshi:
Yes, this is for issue SWDEV-122926.

PR #358, opened 2017-06-14
HCC RT Header needs to trigger re-build

aaronenyeshi:
This resolves JIRA Ticket SWDEV-123229 - HCC RT header changes don't trigger a re-build.

There was an issue with HCC RT Headers not triggering a re-build when running make. This was due to dependencies not being properly set and checked for headers in hcc/include.


aaronenyeshi:
I have tested this on unit tests, and there are no new testcase failures with this. With this fix, developers can modify headers such as kalmar_runtime.h and run make which will be triggered from header file changes. 

Please let me know if I should also make this change for files include/coordinate, include/array_view, and subdirectory include/experimental/.

whchung:
@aaronenyeshi thanks for this PR! please help amend it for other headers within `include` directory. Thanks.

include/CMakeLists.txt
```diff
@@ -1,13 +1,60 @@
...
+  kalmar_math.h
+  kalmar_runtime.h
+  kalmar_serialize.h
+  pinned_vector.hpp)
+
+set(INL_HEADERS
+  hc_norm_unorm.inl
+  hc_short_vector.inl
+  kalmar_short_vectors.inl)
+
```
scchan: is there a way to automatically generate these lists instead of constructing them manually?

scchan (changes requested):

include/CMakeLists.txt
aaronenyeshi: I will take a look

aaronenyeshi (commented):
Just added another commit to generate the dependency without having to specify each header files explicitly, using regular expressions on the file names.

I still need to add another commit for other header files in the `include `directory.

aaronenyeshi:
Okay, I've made it generate lists automatically for the header files.
In addition, I've included the other headers inside `include` .
Also, now there is a target for hcc-headers, and pstl-headers.
 
These commits does not produce any new testcase failures. Please take a look. Thanks!

whchung (approved):
LGTM. @scchan and @pfultz2 ?

scchan:
looks like the build failed to generate the .deb package.
@aaronenyeshi could you do a "make package" and see if you could reproduce it?

aaronenyeshi:
@scchan Yes, I was able to find the bug, and fix it. Seems it dropped the sub-directory impl/ string in the cmake-install.cmake built file.

include/CMakeLists.txt
```diff
@@ -1,20 +1,43 @@
...
+    DEPENDS ${src}
+    COMMAND ${CMAKE_COMMAND} -E copy_if_different ${src} ${dst}
+    COMMENT "Copying HCC's ${f}...")
+  list(APPEND out_files ${dst})
+endforeach( f )
+
+# Create target for hcc-headers and set dependencies
+add_custom_target(hcc-headers ALL DEPENDS ${out_files})
+add_dependencies(world hcc-headers)
+set_target_properties(hcc-headers PROPERTIES FOLDER "HCC Misc")
```
aaronenyeshi: This line (35) might not be useful, as its meant for MS Visual Studio UI navigation. Should I remove it?

whchung: @aaronenyeshi thanks for spotting this. let's remove it for now.

aaronenyeshi: Great! I think it is ready to be merged.

scchan (approved):

PR #361, opened 2017-06-15
Checkout submodules using scm

pfultz2:
This should help save checkout time on jenkins,

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -25,14 +25,13 @@ node ('rocmtest')
   // The client workspace is shared with the docker container
   stage('HCC Checkout')
   {
-    deleteDir( )
-    checkout scm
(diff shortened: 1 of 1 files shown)


whchung (approved):
LGTM. @kknox ?

kknox:
This has the potential of significantly improving our clone/checkout times, by orders of magnitude.  The jenkinsfile code looks correct to me, and would be how I would write it.  

The PR does not yet show how well jenkins will be able to handle the updating of submodules, but we won't really know that until we try it and see.  I think it's worth the risk of a few failed builds.  On jenkins, after a full checkout, I'm seeing a subsequent checkout running as fast as 10s.

kknox (approved):

PR #362, opened 2017-06-15
Feature hcc profile

bensander:
Simple but useful GPU command profiler for HCC.  Enabled with env var, shows kernel and data  xfer times and bandwidth.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -41,7 +41,6 @@
 #include <time.h>
 #include <iomanip>
 
-#define KALMAR_DEBUG (0)
 #ifndef KALMAR_DEBUG
 #define KALMAR_DEBUG (0)
 #endif
@@ -106,8 +105,6 @@ unsigned HCC_DB = 0;
 int HCC_MAX_QUEUES = 20;
 
 
-static std::ofstream g_hccProfileFile; // if using a file open it here
-static std::ostream *g_hccProfileStream = nullptr; // point at file or default stream
 
 #define HCC_PROFILE_VERBOSE_BASIC                   (1 << 0)
 #define HCC_PROFILE_VERBOSE_TIMESTAMP               (1 << 1)
@@ -121,13 +118,13 @@ char * HCC_PROFILE_FILE=nullptr;
 // Profiler:
 #define LOG_PROFILE(start, end, type, tag, msg) \
 {\
-    *g_hccProfileStream  << "profile: " << std::setw(7) << type << ";\t" \
+    Kalmar::ctx.getHccProfileStream() << "profile: " << std::setw(7) << type << ";\t" \
                          << std::setw(25) << tag\
                          << ";\t" << std::fixed << std::setw(6) << std::setprecision(1) << (end-start)/1000.0 << " us;";\
     if (HCC_PROFILE_VERBOSE & (HCC_PROFILE_VERBOSE_TIMESTAMP)) {\
-            *g_hccProfileStream << "\t" << start << ";\t" << end << ";";\
+            Kalmar::ctx.getHccProfileStream() << "\t" << start << ";\t" << end << ";";\
     }\
-   *g_hccProfileStream <<  msg << "\n";\
+   Kalmar::ctx.getHccProfileStream() <<  msg << "\n";\
 }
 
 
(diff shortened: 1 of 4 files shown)


aaronenyeshi:
Hi, is there documentation on these debuggers and profilers? I think it would be very useful to add how to use this tool (such as which env var to set) into debugging documentations.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -41,6 +41,7 @@
 #include <time.h>
 #include <iomanip>
 
+#define KALMAR_DEBUG (0)
```
whchung: @bensander please help remove this line.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2203,9 +2276,68 @@ class HSADevice final : public KalmarDevice
     }
 
     void* CreateKernel(const char* fun, Kalmar::KalmarQueue *queue) override {
+        //fun = "_ZZN8hip_impl21grid_launch_hip_impl_IZ76Cijk_Alik_Bljk_SB_MT064x128x08_GSU06_MO10_PGR0_PLR0_TT04_08_WG16_16_01_WGM01PfPKfS3_ffjjjjjjjjjjjjjP12ihipStream_tjPP11ihipEvent_tS8_E4$_36JNS_17Empty_launch_parmERS1_RjSC_SC_SC_SC_SC_EEEvNS_12_GLOBAL__N_119New_grid_launch_tagE4dim3SF_iRKN2hc16accelerator_viewET_DpOT0_ENUlRKNSG_11tiled_indexILi3EEEE_19__cxxamp_trampolineES1_jjjjjj";
```
whchung: @bensander please help remove this line.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -105,6 +106,31 @@ unsigned HCC_DB = 0;
...
+static std::ostream *g_hccProfileStream = nullptr; // point at file or default stream
+
+#define HCC_PROFILE_VERBOSE_BASIC                   (1 << 0)
+#define HCC_PROFILE_VERBOSE_TIMESTAMP               (1 << 1)
+#define HCC_PROFILE_VERBOSE_BARRIER                 (1 << 2)
+
+
+int HCC_PROFILE=0;
+int HCC_PROFILE_VERBOSE=0;
+char * HCC_PROFILE_FILE=nullptr;
```
whchung, line 116: @bensander i'm seeing more file-scope global variables growing... how about we move them into `HSAContext`?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -395,6 +429,10 @@ class HSACopy : public Kalmar::KalmarAsyncOp {
     hsa_signal_t signal;
     int signalIndex;
     bool isSubmitted;
+    bool isAsync;;          // copy was performed asynchronously
```
whchung: @bensander double `;` in the code.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4509,8 +4751,6 @@ HSACopy::syncCopyExt(hc::hcCommandKind copyDir, const hc::AmPointerInfo &srcPtrI
             DBOUT(DB_COPY, "HSACopy::syncCopy(), hsa_amd_memory_async_copy() returns: 0x" << std::hex << hsa_status << std::dec <<"\n");
             throw Kalmar::runtime_exception("hsa_amd_memory_async_copy error", hsa_status);
         }
-        Kalmar::ctx.releaseSignal(signal, signalIndex);
```
whchung, line 4512: @bensander could you provide justification why these 2 lines be removed in this PR? I thought this PR is to add kernel profiling logic. but changing these 2 lines would essentially change the behavior of HSA signal reclaim process. I'm not alluding the code is wrong, but perhaps it should be made in a separate PR?

whchung (changes requested):

lib/hsa/mcwamp_hsa.cpp
bensander: I believe this is not the right place to deallocate a single - instead this should occur in the HSACopy::dispose() routine.  The new profile mode is implemented inside the dispose routine and uses the timestamps from the signal so requires this change.

lib/hsa/mcwamp_hsa.cpp
bensander: Yes - could be separate pull request?  these are environment variables which arguably are best as globals.  the file handle should be moved.

whchung: ok to make them in another PR.

lib/hsa/mcwamp_hsa.cpp
whchung: thanks for the clarification

whchung (approved):

PR #364, opened 2017-06-19
Add unit test for indirect function argument and SROA

yxsamliu:


whchung:
@yxsamliu Many thanks for contributing to this important fix!

PR #366, opened 2017-06-19
Fix implementation of generic address space to 0 for HCC in rocdl

aaronenyeshi:
HCC will perform transformation to the .ll files from the build directory, instead of the source. Therefore we do not need to git checkout to fix damages to source. SWDEV-123226.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -424,11 +424,12 @@ if (HCC_INTEGRATE_ROCDL)
     DEPENDS clang
   )
 
-  # FIXME: We have to reset the state of the ROCDL before building
-  add_custom_command(TARGET rocdl PRE_BUILD
-    COMMAND git checkout -- .
-    WORKING_DIRECTORY ${ROCDL_SRC_DIR}
-  )
+  # We will be performing addr space transformations in build directory
+  # Adopting generic address space as address space 0
+  file(COPY ${ROCDL_SRC_DIR}/utils/add_amdgiz.sed
+       DESTINATION ${ROCDL_BUILD_DIR}/utils)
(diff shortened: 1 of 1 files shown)


aaronenyeshi:
Also, the rest of this fix was submitted for code-review on gerrit repository of rocdl master branch. Here: http://git.amd.com:8080/#/c/93544/

aaronenyeshi:
Also, I'm not sure if it would be better to push these hcc/CMakeLists.txt changes to hcc/rocdl/CMakeLists.txt. This idea just came up, but haven't tested.

aaronenyeshi:
@b-sumner would it be better to put this change into rocdl/CMakeLists.txt instead?

whchung:
will hold this PR until `ROCm-Device-Libs` is changed.

b-sumner:
Is there any reason to copy these in, instead of just checking them in to rocdl?

aaronenyeshi:
@b-sumner I wanted to run a script "change-addr-space.sh" in the binary directory. However it references another file "add-amdgiz.sed" which is in the src directory. So I copied those files into the build directory under rocdl/utils. Is this okay?

b-sumner:
@aaronenyeshi can't we simply check add-amdgiz.sed into rocdl/utils once, instead of copying it each time cmake is run?

scchan:
It feels that they should go into rocdl's cmake.  Without the context of rocdl's cmake, it's unclear why hcc's cmake need to copy those scripts into the build directory but it "doesn't" use it.  I think it would be clearer if these steps, if required, are added to rocdl's build.

aaronenyeshi:
I've decided it would be better to find an alternative to copying the scripts to Binary directory. Closing this PR .

PR #368, opened 2017-06-19
add arch flag when compiling cmake_test

scchan:


scchan:
@whchung could you take a look when you got a chance?

whchung (approved):
sorry for late response. LGTM.

PR #371, opened 2017-06-21
Suppress readelf error output

whchung:
In case readelf is executed with an ASCII file as an input, suppress
undesirable warning message.

This case would happen when clamp-link script is invoked with text-format
inputs, such as GNU ld version string specification files.


scchan:
LGTM, will wait for the green light from the build for merging in

PR #373, opened 2017-06-23
Revert "Improve HCC linker script"

scchan:
Reverts RadeonOpenCompute/hcc#372


scchan:
Reverting due to regression

PR #374, opened 2017-06-26
[Driver] Inline everything before launching EraseNonkernels pass

whchung:
Teach LLVM inliner we do want all functions used by a kernel be inlined
by assigning a big iniling threshold value.

This allows applications which huge device functions be properly inlined on AMD platform.


scchan:
LGTM

PR #377, opened 2017-06-27
Optimize map lookups for kernel/buffer deps for array_view.

bensander:
First call "find", avoids insert (from []) and then removal
in cases where kernel does not have any array_view deps.

********************
Testing Time: 190.11s
********************
Failing Tests (4):
    HCC :: Unit/AsyncPFE/async_av_dependent7.cpp

  Expected Passes    : 721
  Expected Failures  : 30
  Unexpected Failures: 4


whchung (approved):
LGTM

PR #379, opened 2017-06-29
use llvm' cmake parameters to specify external projects paths

dfukalov:
intead of creating symlinks in source tree

note: possible already created symlinks should be removed manually (or clone clean source tree)


whchung:
LGTM. I'd like to ask for an extra pair of eyes though. @scchan ?

whchung (approved):

PR #381, opened 2017-07-07
Update CI docker scripts to use new repo server

mangupta:


whchung:
@kknox @mangupta CI server failed due to: `gpg: no valid OpenPGP data found.` I guess it could be a configuration issue. The PR itself looks fine.

pfultz2:
The issue is that the IP address for `repo.radeon.com` is not accessible from within AMD's network, and docker doesn't use the host network when building either(as the test nodes do have the correct IP address in the /etc/hosts files). In future version of docker, you can pass in additional hosts. 

For now, as a workaround you can pass in the host as a build arg. So in the dockerfile, something like this:

```
ARG REPO_RADEON=repo.radeon.com

RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends wget && \
    wget -qO - http://$REPO_RADEON/rocm/apt/debian/rocm.gpg.key | apt-key add -
```

Then update the Jenkinsfile to pass in the additional build arg with `--build-arg REPO_RADEON=10.255.8.5`.

pfultz2:
@mangupta The REPO_RADEON arg needs to be added to the 'dockerfile-hcc-lc-ubuntu-16.04' dockerfile as well.

kknox:
I believe all of our test machines have 1.6 kernel now, I can help with this PR tomorrow.

kknox:
With this PR merged, CI is working on clang_tot_upgrade again.  We also need this PR (commits) to be cherry picked into the 1.6.x branch, as that branch is running into the same build problems ctu had.

PR #383, opened 2017-07-10
cmake refactoring

dfukalov:
1. combined install dirs 'bin' and 'compiler/bin'
2. explicit llvm+clang compiler' cmake+make invoke replaced with add_subdirectory()
3. minor improvements


dfukalov:
I've merged in changes with fixes for jenkins and now it passes. Possible it should be better to re-create new "clean" branch with the changes?

whchung:
@dfukalov it'd be a good idea.

PR #395, opened 2017-07-22
support for RHEL7/CentOS7

scchan:
- choose to use libc++ due to older gcc version on rhel7
- link to libstdc++ due to dependency in libc++
- disable abi::__cxa_demangle to avoid conflicts between libcxx and libstdc++ headers


kknox:
Up to the HCC team, but since this PR is all about adding a new distro, a new dockerfile could be added to explicitly test building on that platform.  Downside is it will increase test time, but you could offset that increased time some by doing the build in parallel.  MIOpen does this.

scchan:
That's a good idea.   We don't have pre-built packages for other components on rhel7/centos7 at the moment but I'll add a new dockerfile once they are available

whchung (approved):
LGTM

PR #396, opened 2017-07-22
Fix queue empty check for commands w/o signals.

bensander:
If the command doesn't have a signal, assume it is not yet ready.
Commands without signals come through the
hcc_dispatch_hsa-kerninterface.

********************
Testing Time: 180.24s
********************
Failing Tests (9):
    HCC :: Unit/AsyncPFE/async_av_dependent8.cpp
    HCC :: Unit/AsyncPFE/async_av_independent4.cpp
    HCC :: Unit/Atomic/atomic_add_float_global.cpp
    HCC :: Unit/HC/multi_acc.cpp
    HCC :: Unit/HC/multi_acc_array2.cpp
    HCC :: Unit/ParallelSTL/is_partitioned_carray.cpp

  Expected Passes    : 717
  Expected Failures  : 30
  Unexpected Failures: 9
tests/CMakeFiles/test.dir/build.make:57: recipe for target 'tests/CMakeFiles/test' failed

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -1178,7 +1178,10 @@ public:
 					if (v != 0) {
 						return false;
 					}
-				}
+                } else {
+                    // no signal, have to assume the command is still running
+                    return false;
+                }
             }
         };
         return true;


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1178,7 +1178,10 @@ class HSAQueue final : public KalmarQueue
 					if (v != 0) {
 						return false;
 					}
-				}
+                } else {
+                    // no signal, have to assume the command is still running
+                    return false;
```
scchan, line 1183: Is this fix mostly for efficiency rather than for correctness?
Normally I'd think that eventually it will encounter a command with a signal?  I guess I'm not sure in which situation that the last command would not have a signal?

 I found one usage of isEmpty() in this file, which is in createOrstealRocrQueue.  Note there's a lock_guard before the isEmpty() call; however, there's no lock around pushAsyncOp() so other threads may call this to add new commands to the queue so this could be a race condition?

Also, the usage of lock_guard probably has a bug because it doesn't have a variable name so it would get destroyed right after it's constructed.

scchan (changes requested):

bensander:
This is for correctness.
Commands are not required to have completion signals.  For example, the dispatch path used by MIOpen that directly launches code modules does not utilize completion signals.  PFE_v2 will provide an async PFE with no completion signal flavor of PFE as well.
At any given time, the queue can contain a series of commands with none of them actually writing completion signals.  This still needs to be considered a “non-empty” queue for the purposes of this discussion.

bensander:
We should review the lock_guard, this sounds like a potential bug.
Stealing a queue requires that the queue be empty which might help.  Worth some more deep thought here though too.  V2 uses different hw queue management scheme so can focus future energy there.

scchan:
I'm fine with merging this and fix the lock_guard in a different PR

PR #401, opened 2017-07-27
second part of cmake refactoring

dfukalov:
1. add_custom_target() changed to add_subdirectory() to build rocdl subdir
2. changes related to the transdormation (cmakt-test dependencies, include(AddLLVM) etc)
3. removed transformed_src subdirs from rocdl installation
4. removed NUM_BUILD_THREADS - it's not needed now
5. minor cleanups

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -11,11 +11,6 @@ IF (NOT CMAKE_BUILD_TYPE)
   SET(CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build, options are: Release Debug" FORCE)
 ENDIF(NOT CMAKE_BUILD_TYPE)
 
-IF (NOT NUM_BUILD_THREADS)
-  include(ProcessorCount)
-  ProcessorCount(NUM_BUILD_THREADS)
-ENDIF(NOT NUM_BUILD_THREADS)
-
 # Use two different methods to determine host distribution: lsb_release and if that fails grep /etc/os-release
 find_program( lsb_executable lsb_release )
 
@@ -33,7 +28,7 @@ endif( )
 
 # Accepted values for DISTRO_ID: trusty (Ubuntu 14.04), xenial (Ubuntu 16.06), fd23 (Fedora 23)
 string(TOLOWER "${DISTRO_ID}" DISTRO_ID )
-if( DISTRO_ID MATCHES "ubuntu" OR DISTRO_ID MATCHES "fedora" 
+if( DISTRO_ID MATCHES "ubuntu" OR DISTRO_ID MATCHES "fedora"
    OR DISTRO_ID MATCHES "centos" OR DISTRO_ID MATCHES "redhatenterpriseserver")
   message( STATUS "Detected distribution: ${DISTRO_ID}:${DISTRO_RELEASE}" )
 else()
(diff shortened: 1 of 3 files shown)


cmake-tests/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ if(NOT CMAKE_CXX_COMPILER MATCHES ".*hcc")
 endif()
 
 add_executable(cmake-test cmake-test.cpp)
+add_dependencies(cmake-test clang_links rocdl_links)
```
pfultz2: This won't work in a standalone build, and this is designed to work both standalone and in the project.

dfukalov: thanks! fix submitted

cmake-tests/CMakeLists.txt
```diff
@@ -17,6 +17,7 @@ endif()
 set_target_properties(cmake-test PROPERTIES LINK_FLAGS ${new_cmake_test_link_flags})
 
 if(TARGET hccrt)
+    add_dependencies(cmake-test clang_links rocdl_links)
```
pfultz2, line 20: Shouldn't these dependencies be expressed in hccrt target?

dfukalov: No, this add_dependencies is about 'cmake-test' target that is described in the file. It was moved into 'if(TARGET hccrt)' since the dependencies should not affect standalone build of 'cmake-test'

CMakeLists.txt
```diff
@@ -372,20 +365,24 @@ if (HCC_INTEGRATE_ROCDL)
...
+  include(AddLLVM)
+
+  add_definitions(${LLVM_DEFINITIONS})
+
+  # build rocdl
+  add_subdirectory(${ROCDL_SRC_DIR})
 
   file(MAKE_DIRECTORY ${ROCDL_BUILD_DIR}/lib)
-  add_custom_command(TARGET rocdl POST_BUILD
+  add_custom_target(rocdl_links DEPENDS ${AMDGCN_LIB_TARGETS})
```
scchan, line 384: where AMDGCN_LIB_TARGETS was defined?

scchan (changes requested):

CMakeLists.txt
dfukalov: the library we include in hcc has parameter `AMDGCN_TARGETS_LIB_LIST` that contains variable name to return libs targets list: `set(AMDGCN_TARGETS_LIB_LIST "AMDGCN_LIB_TARGETS")` so after `add_subdirectory(${ROCDL_SRC_DIR})` we have all target names built by library in `AMDGCN_LIB_TARGETS` variable

scchan (approved):

CMakeLists.txt
```diff
@@ -372,20 +365,24 @@ if (HCC_INTEGRATE_ROCDL)
...
-            -DAMDHSACOD=${AMDHSACOD}
-            -DGENERIC_IS_ZERO=ON
-            -DBUILD_HC_LIB=ON
-    COMMAND make -j ${NUM_BUILD_THREADS} # not portable, but we need it this way
-    WORKING_DIRECTORY ${ROCDL_BUILD_DIR}
-    DEPENDS clang_links llvm-link
-  )
+  set(LLVM_DIR ${CLANG_BIN_DIR})
+  set(AMDHSACOD ${AMDHSACOD})
+  set(GENERIC_IS_ZERO ON)
```
whchung, line 370: Turns out this PR is bad, and will render ROCDL not working.

whchung: Turns out `set(GENERIC_IS_ZERO ON CACHE BOOL ON FORCE)` is the proper way to ensure the flag is passed down to `ROCDL`.

Also this PR breaks `change-addr-space.sh` within `ROCDL`. I've posted another commit to fix it temporarily and would submit for formal review in `ROCDL`:
https://github.com/whchung/ROCm-Device-Libs/commit/96375f63f45383ced9328221e6cbaccf0e06b13d

dfukalov: Yes, you're right it was a bug of the change. I've double-checked other rocdl' pararamaters set in the filem and they're ok

PR #408, opened 2017-07-31
Remove Bsymbolic

AlexVlx:
This changes the procedure for retrieving bundled code objects to follow 
the one developed for HC2. The symbols which point into the .kernel
section are internalised so as to not be affected by interposition. At
runtime, alongside the local .kernel section, all other such sections
present in linked in shared objects are obtained via ELF introspection.
This removes the need for using the Bsymbolic hack. Unit tests are
corrected to reflect this. The ELFIO library is used for the
introspection, and it is added into the folder which is to host HC2 in
the near future.


AlexVlx:
This is not sufficient for all cases, hence closing it pending resubmission.

whchung:
@AlexVlx could you briefly explain what cases are missing in this PR? I was just planning to review this one and it'd be very helpful to let me gain some insights faster.

AlexVlx:
@whchung I've sorted things out, if you'd care to review. Thank you.

whchung:
Thanks. Let me do this first thing on Monday.

hc2/headers/functions/hsa_interfaces.hpp
```diff
@@ -0,0 +1,299 @@
...
+#include <hsa/hsa.h>
+
+#include <algorithm>
+#include <cstddef>
+#include <functional>
+#include <memory>
+#include <utility>
+
+inline
+bool operator==(const hsa_agent_t& x, const hsa_agent_t& y)
```
scchan: can we add `static` to all the free functions?

AlexVlx: @scchan we could but in this context I think it may not be very useful. Are you concerned about potential ODR violations if someone else decides to make hsa_agent_t EqualityComparable? The downside is that now you technically may get an operator== code drop in all TUs which include that header (an academic issue, agreed). I do not feel strongly either way, and these are temporary...

scchan: It's mainly about potential ODR violations.  We have other headers in hcc containing inlined functions  without `static` and they are included in multiple .cpp files.  That is causing some portability issues with some compilers.  We are looking into fixing them.

AlexVlx: @scchan done:)

PR #411, opened 2017-08-03

weixingzhang:
aligned.

original allocated address.


whchung (approved):
LGTM.

whchung (changes requested):
@weixingzhang per inputs from @bensander and @scchan would you mind revise this PR so we retain existing `am_alloc` API, but add another `am_aligned_alloc` API?

weixingzhang:
The change is updated to reflect the feedback. Please review PR #415. Thanks!

scchan:
superseded by https://github.com/RadeonOpenCompute/hcc/pull/415

PR #414, opened 2017-08-04
Fix - mcwamp.cpp would reference outdated headers

aaronenyeshi:
mcwamp.cpp may reference outdated headers files in binary directory.
So we must install the headers before we start building rocdl or lib.

Case: if you change hc_rt_debug.h and mcwamp.cpp then compile, you cannot reference new definitions you've created because mcwamp.cpp is built first, and will reference hc_rt_debug.h from previous build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -364,6 +364,9 @@ install( FILES ${CMAKE_CURRENT_SOURCE_DIR}/scripts/cmake/ImportedTargets.cmake
 
 add_custom_target(world DEPENDS clang_links)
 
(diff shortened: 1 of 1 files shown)


whchung (approved):
LGTM. Thanks for contributing this quiet important improvement!

bensander:
Thx!    I certainly appreciate these too. :)

On Aug 4, 2017, at 2:04 PM, Wen-Heng (Jack) Chung <notifications@github.com<mailto:notifications@github.com>> wrote:


Merged #414<https://github.com/RadeonOpenCompute/hcc/pull/414>.

PR #415, opened 2017-08-08

weixingzhang:
aligned.

original allocated address.

### DIFF
--- a/lib/hsa/hc_am.cpp
+++ b/lib/hsa/hc_am.cpp
@@ -26,6 +26,7 @@ AmPointerInfo & AmPointerInfo::operator= (const AmPointerInfo &other)
 {
     _hostPointer = other._hostPointer;
     _devicePointer = other._devicePointer;
+    _unalignedDevicePointer = other._unalignedDevicePointer;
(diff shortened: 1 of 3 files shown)


lib/hsa/hc_am.cpp
```diff
@@ -238,9 +239,8 @@ AmPointerTracker g_amPointerTracker;  // Track all am pointer allocations.
 namespace hc {
 
 // Allocate accelerator memory, return NULL if memory could not be allocated:
-auto_voidp am_alloc(size_t sizeBytes, hc::accelerator &acc, unsigned flags) 
+auto_voidp am_alloc(size_t sizeBytes, hc::accelerator &acc, unsigned flags)
```
scchan, line 242: I think am_alloc should just call am_aligned_alloc underneath..

scchan (changes requested):

lib/hsa/hc_am.cpp
weixingzhang: updated it in PR#419. Please review it.

PR #418, opened 2017-08-09
[tests] fix unit test

vsytch:
asin and acos were not getting proper input, resulting in them returning NAN. This change sets their input domain to be [-1, 1].


scchan:
Could you create a fresh new branch from clang_tot_upgrade, cherry-pick e3fee96 and then re-submit your PR?  That will remove the commits for change .gitmodules.  Thanks!

PR #419, opened 2017-08-09

weixingzhang:
aligned.

original allocated address.

### DIFF
--- a/include/hc_am.hpp
+++ b/include/hc_am.hpp
@@ -19,7 +19,8 @@ namespace hc {
 class AmPointerInfo {
 public:
     void *      _hostPointer;   ///< Host pointer.  If host access is not allowed, NULL.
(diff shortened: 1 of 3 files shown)


scchan (changes requested):
Could you create a unit test for am_aligned_alloc?

whchung (changes requested):
@weixingzhang Besides asking for test cases for this PR, I've got one question:

In `AmPointerTracker::reset()`, you changed arguments for `hsa_amd_memory_pool_free()` from `iter->first._basePointer` to `iter->second._unalignedDevicePointer`. This is completely understandable.

However, in `AmPointerTracker::update_peers()`, similar logic were applied to `hsa_amd_agents_allow_access()`. Is this intended design, or an omission?

PR #420, opened 2017-08-09
[tests] fix unit test

vsytch:
asin and acos were not getting proper input, resulting in them returning NAN. This change sets their input domain to be [-1, 1].


tests/Unit/AmpMath/amp_math_acos.cpp
```diff
@@ -33,9 +39,18 @@ int main(void) {
     gb[i] = fast_math::acos(ga[i]);
   }
 
-  float sum = 0;
+  _Tp sum = 0.0;
   for(unsigned i = 0; i < vecSize; i++) {
     sum += fast_math::fabs(fast_math::fabs(gc[i]) - fast_math::fabs(gb[i]));
```
scchan: could you take a look at this line?  it seems that it should only apply fabs() on the difference rather than on the values of acos()?

tests/Unit/AmpMath/amp_math_acosf.cpp
```diff
@@ -33,9 +39,18 @@ int main(void) {
     gb[i] = fast_math::acosf(ga[i]);
   }
 
-  float sum = 0;
+  _Tp sum = 0.0;
   for(unsigned i = 0; i < vecSize; i++) {
     sum += fast_math::fabs(fast_math::fabs(gc[i]) - fast_math::fabs(gb[i]));
```
scchan: see previous comment on fabs

tests/Unit/AmpMath/amp_math_asin.cpp
```diff
@@ -31,9 +39,17 @@ int main(void) {
     gb[i] = fast_math::asin(ga[i]);
   }
 
-  float sum = 0;
+  _Tp sum = 0.0;
   for(unsigned i = 0; i < vecSize; i++) {
     sum += fast_math::fabs(fast_math::fabs(gc[i]) - fast_math::fabs(gb[i]));
```
scchan: see previous comment on fabs()

tests/Unit/AmpMath/amp_math_asinf.cpp
```diff
@@ -30,9 +39,17 @@ int main(void) {
     gb[i] = fast_math::asinf(ga[i]);
   }
 
-  float sum = 0;
+  _Tp sum = 0.0;
   for(unsigned i = 0; i < vecSize; i++) {
     sum += fast_math::fabs(fast_math::fabs(gc[i]) - fast_math::fabs(gb[i]));
```
scchan: see previous comment on fabs

scchan (changes requested):

tests/Unit/AmpMath/amp_math_acos.cpp
vsytch: You're right, but we're doing this in every AmpMath test. I'll go through them and fix this.

PR #423, opened 2017-08-10
Refactored test cases

vsytch:
This change refactors the code of the AmpMath test cases so it's consistent between them, adds a check if a math function returns NaN, and fixes absolute error not being calculated correctly.


whchung (approved):
LGTM

PR #428, opened 2017-08-15
hc::accelerator_view::copy() enhancement

whchung:
improve copy implementation logic to take offsets into consideration

Old logic simply uses pointers returned by AM trackers, but it doesn't
consider the case where user codes could add offsets from pointers returned by
am_alloc().

One example would be:

int* buffer = am_alloc(size);
int* src = &buffer[offset];
av.copy(src, dst, size-offset);

In this PR the logic of hc::acclerator_view::copy() is improved to be
aware of this kind of use cases, and 2 test cases are provided.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -4454,30 +4454,61 @@ hsa_status_t HSACopy::hcc_memory_async_copy(Kalmar::hcCommandKind copyKind, cons
 
     hsa_agent_t hostAgent = const_cast<Kalmar::HSADevice *> (copyDeviceArg)->getHostAgent();
 
-    void *dstPtr = dstPtrInfo._devicePointer;
-    void *srcPtr = srcPtrInfo._devicePointer;
+    /* Determine src and dst pointer passed to ROCR runtime.
+     *
+     * Pre-condition:
+     * - this->dst and this->src must be tracked by AM API implemantation.
+     * - dstPtrInfo and srcPtrInfo must be valid.
+     */
+    void *dstPtr = nullptr;
+    void *srcPtr = nullptr;
 
     hsa_agent_t srcAgent, dstAgent;
     switch (copyKind) {
         case Kalmar::hcMemcpyHostToHost: 
             srcAgent=hostAgent; dstAgent=hostAgent;
 
-            // Use host pointers since this copy will be performed with CPU.
-            // If pointers are registered, then devicePointer may not match host pointer.
-            dstPtr = dstPtrInfo._hostPointer;
-            srcPtr = srcPtrInfo._hostPointer;
-            
-            //throw Kalmar::runtime_exception("HCC should not use hsa_memory_async_copy for host-to-host copy");
+            /* H2H case
+             * Simply pass this->src and this->dst to ROCR runtime.
+             */
+            dstPtr = this->dst;
+            srcPtr = const_cast<void*>(this->src);
             break;
         case Kalmar::hcMemcpyHostToDevice: 
             srcAgent=hostAgent; dstAgent=copyAgent;
+
+            /* H2D case
+             * Destination is simply this->dst.
+             * Source has to be calculated by adding mapped GPU device pointer
+             * with potential offsets specified in user codes.
+             */
+            dstPtr = this->dst;
+            srcPtr = reinterpret_cast<unsigned char*>(srcPtrInfo._devicePointer) +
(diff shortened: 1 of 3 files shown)


whchung:
HIP test results: 
```100% tests passed, 0 tests failed out of 74```

File: lib/hsa/mcwamp_hsa.cpp [hunk 1]
```diff
@@ -4454,30 +4454,61 @@ hsa_status_t HSACopy::hcc_memory_async_copy(Kalmar::hcCommandKind copyKind, cons
...
+             */
+            dstPtr = this->dst;
+            srcPtr = const_cast<void*>(this->src);
             break;
         case Kalmar::hcMemcpyHostToDevice: 
             srcAgent=hostAgent; dstAgent=copyAgent;
+
+            /* H2D case
+             * Destination is simply this->dst.
+             * Source has to be calculated by adding mapped GPU device pointer
```
scchan: Instead of "mapped GPU device pointer", I think it would be clearer to say "by adding the offset to the pinned host pointer"?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4454,30 +4454,61 @@ hsa_status_t HSACopy::hcc_memory_async_copy(Kalmar::hcCommandKind copyKind, cons
...
+            srcPtr = reinterpret_cast<unsigned char*>(srcPtrInfo._devicePointer) +
+                     (reinterpret_cast<unsigned char*>(const_cast<void*>(this->src)) -
+                      reinterpret_cast<unsigned char*>(srcPtrInfo._hostPointer));
             break;
         case Kalmar::hcMemcpyDeviceToHost: 
             srcAgent=copyAgent; dstAgent=hostAgent;
+
+            /* H2D case
+             * Source is simply this->src.
+             * Desination has to be calculated by adding mapped GPU device
```
scchan: see previous comment on "mapped GPU device pointer"

scchan (commented):
I only have some suggestions on the wording of some code comments; otherwise LGTM

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4454,30 +4454,61 @@ hsa_status_t HSACopy::hcc_memory_async_copy(Kalmar::hcCommandKind copyKind, cons
...
+    void *dstPtr = nullptr;
+    void *srcPtr = nullptr;
 
     hsa_agent_t srcAgent, dstAgent;
     switch (copyKind) {
         case Kalmar::hcMemcpyHostToHost: 
             srcAgent=hostAgent; dstAgent=hostAgent;
 
-            // Use host pointers since this copy will be performed with CPU.
-            // If pointers are registered, then devicePointer may not match host pointer.
```
bensander, line 4466: I think it is important to retain the comment that we are expecting ROCR runtime to continue to use host pointers for host to host copies.  (rather than using GPU copy engine which would use device pointers)

whchung: @bensander I'd like to ask for your clarification on whether to keep the comment only, or should we continue to use `dstPtrInfo._hostPointer` in H2H case? My interpretation is the former one, not the later one. But I'd like to get your consent. Thanks.

lib/hsa/mcwamp_hsa.cpp
whchung: @scchan thanks. let me rephrase.

lib/hsa/mcwamp_hsa.cpp
whchung: @bensander I'd like to ask for your clarification on whether to keep the comment only, or should we continue to use dstPtrInfo._hostPointer in H2H case? My interpretation is the former one, not the later one. But I'd like to get your consent. Thanks.

lib/hsa/mcwamp_hsa.cpp
whchung: @scchan thanks. let me rephrase.

lib/hsa/mcwamp_hsa.cpp
bensander: Let's take the opportunity to refactor, say:
"We expect ROCR runtime to continue to use the CPU for host to host copies, and thus must pass host pointers here."

whchung:
@scchan @bensander I've revised the comments. Please kindly review once again. Thanks.

scchan (approved):
LGTM

PR #432, opened 2017-08-18
HCC Runtime optimizations

bensander:
Collection of HCC runtime optimizations and improvements - see commit history for details.

Passes HIP dtests and tested in larger MI framework.

Not sure on failure to automatically merge?

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -1250,7 +1250,7 @@ public:
                 if (!foundFirstValidOp) {
                     hsa_signal_t sig =  *(static_cast <hsa_signal_t*> (asyncOp->getNativeHandle()));
                     assert(sig->handle != 0);
-                    foundFirstValidOp = false;
+                    foundFirstValidOp = true;
                 }
                 // wait on valid futures only
                 std::shared_future<void>* future = asyncOp->getFuture();
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -221,18 +221,18 @@ static const char* getHSAErrorString(hsa_status_t s) {
 #define STATUS_CHECK(s,line) if (s != HSA_STATUS_SUCCESS && s != HSA_STATUS_INFO_BREAK) {\
     hc::print_backtrace(); \
     const char* error_string = getHSAErrorString(s);\
-		printf("### HCC STATUS_CHECK Error: %s (0x%x) at file:%s line:%d\n", error_string, s, __FILENAME__, line);\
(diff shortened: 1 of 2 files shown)


whchung:
@bensander your branch doesn't seem to track the tip of HCC. could you rebase and submit once again?

bensander:
Done!

bensander:
@whchung can you review?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1230,12 +1240,21 @@ class HSAQueue final : public KalmarQueue
...
+
+        bool foundFirstValidOp = false;
+
         for (int i = asyncOps.size()-1; i >= 0;  i--) {
             if (asyncOps[i] != nullptr) {
                 auto asyncOp = asyncOps[i];
+                if (!foundFirstValidOp) {
+                    hsa_signal_t sig =  *(static_cast <hsa_signal_t*> (asyncOp->getNativeHandle()));
+                    assert(sig->handle != 0);
+                    foundFirstValidOp = false;
```
scchan: should set to "true" here?

scchan (changes requested):

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1084,21 +1088,25 @@ class HSAQueue final : public KalmarQueue
...
                     << std::endl);
 
 
 
-        if (asyncOps.size() >= MAX_INFLIGHT_COMMANDS_PER_QUEUE) {
+        if (!drainingQueue_ && (asyncOps.size() >= MAX_INFLIGHT_COMMANDS_PER_QUEUE-1)) {
             DBOUT(DB_WAIT, "*** Hit max inflight ops asyncOps.size=" << asyncOps.size() << ". op#" << op->getSeqNum() << " force sync\n");
             DBOUT(DB_RESOURCE, "*** Hit max inflight ops asyncOps.size=" << asyncOps.size() << ". op#" << op->getSeqNum() << " force sync\n");
 
+            drainingQueue_ = true;
```
whchung, line 1100: do we need a mutex here?

lib/hsa/mcwamp_hsa.cpp
whchung: this line indeed seems to be incorrect.

whchung (changes requested):

lib/hsa/mcwamp_hsa.cpp
bensander: Yep.  Pushed a fix.

lib/hsa/mcwamp_hsa.cpp
bensander: I believe HCC queues are not thread-safe by design, so no mutex is required here.

whchung (approved):

PR #435, opened 2017-08-18
Enable functionality to build downstream hip

kknox:
*  If a branch build in clang_tot_upgrade, this will just fire off a build of
HIP to keep HIP up to date with hcc

*  If a PR request, this will fire off a test build of HIP and wait on the results
to determine that the PR does not break HIP build or directed tests

This PR should wait for ROCm-Developer-Tools/HIP#161 to be merged; it won't do what it's supposed to do without the above PR being merged first.


whchung (approved):
LGTM

PR #437, opened 2017-08-21
Improve Printf implementation to use char arrays on GPU

aaronenyeshi:
Allocated a new buffer to store format strings used in printf calls.
Strings created on gpu will be copied into the string buffer on
host coherent memory, and will be printed after kernels. Users can
also write strings directly into arguments of printf call, as shown
in testcase.

TODO: We want to hide the printf implementation from user. Also we
want the printf buffer to be cyclical, so that we can reuse space
that is already printed. There is also a memory access fault that we
run into when the space used is close to buffer size, which needs
further investigation.

### DIFF
--- a/include/hc_printf.hpp
+++ b/include/hc_printf.hpp
@@ -35,6 +35,8 @@ enum PrintfPacketDataType {
   ,PRINTF_CONST_VOID_PTR
   ,PRINTF_BUFFER_CURSOR
   ,PRINTF_BUFFER_SIZE
+  ,PRINTF_FORMAT_BUFFER_SIZE
+  ,PRINTF_FORMAT_BUFFER_CURSOR
 };
 
 class PrintfPacket {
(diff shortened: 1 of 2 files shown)


include/hc_printf.hpp
```diff
@@ -82,30 +93,69 @@ static inline void countArg(unsigned int& count, const T& t, const Rest&... rest
...
+  queue[offset].set(&fbuffer[fb_offset]);
+  return PRINTF_SUCCESS;
+}
+
 template <typename T>
-static inline void set_batch(PrintfPacket* queue, int offset, const T t) [[hc,cpu]] {
-  queue[offset].set(t);
+static inline PrintfError set_batch(PrintfPacket* queue, char* fbuffer, int offset, const T t) [[hc,cpu]] {
+  if (std::is_pointer<T>::value){
+    if (process_str_batch(queue, fbuffer, offset, (char*)t) == PRINTF_FORMAT_BUFFER_OVERFLOW)
```
scchan: This could be an issue here.  It assumes that any pointer passing to printf is a string, which is not the case since printf supports the %p format specifier.

include/hc_printf.hpp
```diff
@@ -82,30 +93,69 @@ static inline void countArg(unsigned int& count, const T& t, const Rest&... rest
...
+
+  return PRINTF_SUCCESS;
 }
 template <typename T, typename... Rest>
-static inline void set_batch(PrintfPacket* queue, int offset, const T t, Rest... rest) [[hc,cpu]] {
-  queue[offset].set(t);
-  set_batch(queue, offset + 1, rest...);
+static inline PrintfError set_batch(PrintfPacket* queue, char* fbuffer, int offset, const T t, Rest... rest) [[hc,cpu]] {
+  if (std::is_pointer<T>::value){
+    if (process_str_batch(queue, fbuffer, offset, (char*)t) == PRINTF_FORMAT_BUFFER_OVERFLOW)
```
scchan: The check is too specific, it would be better to check if the return code != PRINTF_SUCCESS

include/hc_printf.hpp
```diff
@@ -82,30 +93,69 @@ static inline void countArg(unsigned int& count, const T& t, const Rest&... rest
...
+      return PRINTF_FORMAT_BUFFER_OVERFLOW;
```
scchan: This is too specific, it should just return the status from process_str_batch

include/hc_printf.hpp
```diff
@@ -82,30 +93,69 @@ static inline void countArg(unsigned int& count, const T& t, const Rest&... rest
...
+}
+
+static inline PrintfError process_str_batch(PrintfPacket* queue, char* fbuffer, int offset, const char* fstring) [[hc,cpu]] {
+  unsigned int fstr_len = string_length(fstring);
+  unsigned int fb_offset = queue[3].data.ai.fetch_add(fstr_len + 1);
+  if (queue[3].data.ui > queue[2].data.ui){
+    return PRINTF_FORMAT_BUFFER_OVERFLOW;
+  }
+  copy_n(&fbuffer[fb_offset], fstring, fstr_len);
+  fbuffer[fb_offset + fstr_len] = '\0';
```
scchan: It would be betterto do 
  copy_n(&fbuffer[fb_offset], fstring, **fstr_len + 1**);
then explicitly adding an extra '\0' at the end

scchan (changes requested):

aaronenyeshi:
@scchan I've added changes which address your comments, now we will check for char* and const char*. However for users who want to use %p specifier for their char arrays must first cast to void*.

scchan (approved):

PR #443, opened 2017-08-25
Printf - Ordering in Buffers using 1 atomic inst

aaronenyeshi:
Previously, the packets were being allocated into the two buffers in different orders with respect to each other. Now using a single atomic compare_exchange_weak, we can maintain ordering in both buffers with respect with each other. Also, string arguments in a single printf, will be allocated consecutively in-order as well. Offsets will be managed in a single ullong, which can be divided into 2 4byte uints. Also, we are now performing a look-ahead to check for the space required to allocation into the string buffer, so we can perform space allocation in printf command chunks.

Also small update to test.

### DIFF
--- a/include/hc_printf.hpp
+++ b/include/hc_printf.hpp
@@ -33,10 +33,11 @@ union PrintfPacketData {
 
 enum PrintfPacketDataType {
   // Header types
-  PRINTF_BUFFER_SIZE
-  ,PRINTF_STRING_BUFFER
-  ,PRINTF_STRING_BUFFER_SIZE
-  ,PRINTF_OFFSETS
+  PRINTF_BUFFER_SIZE = 0
+  ,PRINTF_STRING_BUFFER = 1
+  ,PRINTF_STRING_BUFFER_SIZE = 2
+  ,PRINTF_OFFSETS = 3
+  ,PRINTF_HEADER_SIZE = 4
 
   // Packet Data types
   ,PRINTF_UNUSED
@@ -75,25 +76,26 @@ static inline PrintfPacket* createPrintfBuffer(hc::accelerator& a, const unsigne
     printfBuffer = hc::am_alloc(sizeof(PrintfPacket) * numElements, a, amHostCoherent);
 
     // Initialize the Header elements of the Printf Buffer
-    printfBuffer[0].type = PRINTF_BUFFER_SIZE;
-    printfBuffer[0].data.ui = numElements;
+    printfBuffer[PRINTF_BUFFER_SIZE].type = PRINTF_BUFFER_SIZE;
+    printfBuffer[PRINTF_BUFFER_SIZE].data.ui = numElements;
 
     // Header includes a helper string buffer which holds all char* args
-    printfBuffer[1].type = PRINTF_STRING_BUFFER;
-    printfBuffer[1].data.ptr = hc::am_alloc(sizeof(char) * numElements * 12, a, amHostCoherent);
-    printfBuffer[2].type = PRINTF_STRING_BUFFER_SIZE;
-    printfBuffer[2].data.ui = numElements * 12;
+    // PrintfPacket is 12 bytes, equivalent string buffer size used
+    printfBuffer[PRINTF_STRING_BUFFER].type = PRINTF_STRING_BUFFER;
+    printfBuffer[PRINTF_STRING_BUFFER].data.ptr = hc::am_alloc(sizeof(char) * numElements * 12, a, amHostCoherent);
+    printfBuffer[PRINTF_STRING_BUFFER_SIZE].type = PRINTF_STRING_BUFFER_SIZE;
+    printfBuffer[PRINTF_STRING_BUFFER_SIZE].data.ui = numElements * 12;
 
     // Using one atomic offset to maintain order and atomicity
-    printfBuffer[3].type = PRINTF_OFFSETS;
-    printfBuffer[3].data.ai[0] = 4;
-    printfBuffer[3].data.ai[1] = 0;
+    printfBuffer[PRINTF_OFFSETS].type = PRINTF_OFFSETS;
+    printfBuffer[PRINTF_OFFSETS].data.ai[0] = PRINTF_HEADER_SIZE;
+    printfBuffer[PRINTF_OFFSETS].data.ai[1] = 0;
   }
   return printfBuffer;
 }
 
 void deletePrintfBuffer(PrintfPacket*& buffer) {
-  hc::am_free(buffer[1].data.ptr);
+  hc::am_free(buffer[PRINTF_STRING_BUFFER].data.ptr);
   hc::am_free(buffer);
   buffer = NULL;
 }
@@ -132,8 +134,8 @@ static inline void countArg(unsigned int& count_arg, unsigned int& count_char, c
 static inline PrintfError process_str_batch(PrintfPacket* queue, int poffset, unsigned int& soffset, const char* string) [[hc,cpu]] {
   unsigned int str_len = string_length(string);
   unsigned int sb_offset = soffset;
-  char* string_buffer = (char*) queue[1].data.ptr;
-  if (!string_buffer || soffset + str_len + 1 > queue[2].data.ui){
+  char* string_buffer = (char*) queue[PRINTF_STRING_BUFFER].data.ptr;
+  if (!string_buffer || soffset + str_len + 1 > queue[PRINTF_STRING_BUFFER_SIZE].data.ui){
     return PRINTF_STRING_BUFFER_OVERFLOW;
   }
   copy_n(&string_buffer[sb_offset], string, str_len + 1);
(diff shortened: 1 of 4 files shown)


include/hc_printf.hpp
```diff
@@ -111,56 +102,93 @@ static inline void copy_n(char* dest, const char* src, unsigned int len) [[hc,cp
...
+  else if (!queue[1].data.ptr || count_char + queue[3].data.ai[1] > queue[2].data.ui){
+    error = PRINTF_STRING_BUFFER_OVERFLOW;
+  }
+  else {
+    do {
+	  old_off.al = queue[3].data.al.load();
+	  old_long = old_off.al.load();
+	  try_off.ai[0] = old_off.ai[0] + count_arg + 1;
+	  try_off.ai[1] = old_off.ai[1] + count_char;
+    } while(!(queue[3].data.al.compare_exchange_weak(old_long, (unsigned long long)try_off.al.load())));
```
scchan: old_off and try_off are local to the thread and therefore they don't need to be defined with an atomic type

include/hc_printf.hpp
```diff
@@ -64,40 +64,31 @@ enum PrintfError {
...
-    printfBuffer[4].data.ui = 0;
+    printfBuffer[1].type = PRINTF_STRING_BUFFER;
+    printfBuffer[1].data.ptr = hc::am_alloc(sizeof(char) * numElements * 12, a, amHostCoherent);
+    printfBuffer[2].type = PRINTF_STRING_BUFFER_SIZE;
+    printfBuffer[2].data.ui = numElements * 12;
+
+    // Combine offsets into 1 atomic element to maintain order and atomicity
+    printfBuffer[3].type = PRINTF_OFFSETS;
+    printfBuffer[3].data.ai[0] = 4;
+    printfBuffer[3].data.ai[1] = 0;
```
scchan: Could we create names for different header indices?

include/hc_printf.hpp
```diff
@@ -111,56 +102,93 @@ static inline void copy_n(char* dest, const char* src, unsigned int len) [[hc,cp
...
+	  try_off.ai[1] = old_off.ai[1] + count_char;
+    } while(!(queue[3].data.al.compare_exchange_weak(old_long, (unsigned long long)try_off.al.load())));
+
+	unsigned int poffset = (unsigned int)old_off.ai[0];
+	unsigned int soffset = (unsigned int)old_off.ai[1];
+
+    if (poffset + count_arg + 1 > queue[0].data.ui) {
+      error = PRINTF_BUFFER_OVERFLOW;
+    }
+	else if (soffset + count_char > queue[2].data.ui){
```
scchan: indent issue...

scchan (changes requested):

include/hc_printf.hpp
aaronenyeshi: Sorry, what did you mean create names? Right now, different header indices have a type member that aligns with enum PrintfPacketDataType.

scchan: right... could we re-arrange the PrintPacketDataType enum to make the code more readable:

```
enum PrintfPacketDataType {
  PRINTF_BUFFER_SIZE = 0,
  PRINTF_STRING_BUFFER,
...
};

...

printfBuffer[PRINTF_BUFFER_SIZE].type = PRINTF_BUFFER_SIZE;
printfBuffer[PRINTF_BUFFER_SIZE].data.ui = numElements;
```

aaronenyeshi:
I've fixed the implementation due to comments. Runs correctly on the testcases. @whchung can you take a look while @scchan is away

whchung:
@aaronenyeshi will do.

include/hc_printf.hpp
```diff
@@ -64,40 +71,33 @@ enum PrintfError {
...
-    printfBuffer[3].type = PRINTF_STRING_BUFFER_SIZE;
-    printfBuffer[3].data.ui = numElements * 12;
-    printfBuffer[4].type = PRINTF_STRING_BUFFER_CURSOR;
-    printfBuffer[4].data.ui = 0;
+
+    // Header includes a helper string buffer which holds all char* args
+    printfBuffer[1].type = PRINTF_STRING_BUFFER;
+    printfBuffer[1].data.ptr = hc::am_alloc(sizeof(char) * numElements * 12, a, amHostCoherent);
+    printfBuffer[2].type = PRINTF_STRING_BUFFER_SIZE;
+    printfBuffer[2].data.ui = numElements * 12;
```
whchung: @aaronenyeshi  i'm a bit afraid of all these magic numbers. could you try change them into enums?

include/hc_printf.hpp
```diff
@@ -241,14 +276,14 @@ static inline void processPrintfBuffer(PrintfPacket* gpuBuffer) {
 
   if (gpuBuffer == NULL) return;
   unsigned int bufferSize = gpuBuffer[0].data.ui;
-  unsigned int cursor = gpuBuffer[1].data.ui;
-  unsigned int numPackets = ((bufferSize<cursor)?bufferSize:cursor) - 5;
+  unsigned int cursor = gpuBuffer[3].data.ai[0];
+  unsigned int numPackets = ((bufferSize<cursor)?bufferSize:cursor) - 4;
```
whchung: @aaronenyeshi  i'm a bit afraid of all these magic numbers. could you try change them into enums?

whchung (changes requested):
Could you change those magic values into enums?

aaronenyeshi:
@whchung I've removed the magic values and used enums instead. I didn't know this was what SiuChi meant initially. Thanks for reviewing!

include/hc_printf.hpp
```diff
@@ -23,23 +23,31 @@ union PrintfPacketData {
   float           f;
   void*           ptr;
   const void*     cptr;
-  std::atomic_int ai;
+
+  // Used by headers for atomic offset insts
+  std::atomic_uint ai[2];
```
whchung: @aaronenyeshi for other types they are relatively easy to understand, but perhaps it's better to document what these fields are for. especially what `ai[0]` `ai[1]` `uia[0]` `uia[1]` means.

whchung (changes requested):
@aaronenyeshi looks much better now, if you could add some more comments on `PrintfPacketData`.

aaronenyeshi:
@whchung I've changed the atomic data members a little bit. Seems that we didn't need atomic uint anymore, since we only use a single atomic compare_exchange on atomic ullong al. I've added some comments as well.

include/hc_printf.hpp
```diff
@@ -64,40 +76,34 @@ enum PrintfError {
 
 static inline PrintfPacket* createPrintfBuffer(hc::accelerator& a, const unsigned int numElements) {
   PrintfPacket* printfBuffer = NULL;
-  if (numElements > 6) {
+  if (numElements > 5) {
```
whchung: 5 looks to be a magic value worthwhile be enumed.

also the logic would return a `nullptr`, but other APIs such as `deletePrintBuffer` doesn't guard `nullptr`. I would recommend you improve the test case to check `nullptr` cases as well.

whchung (changes requested):

include/hc_printf.hpp
```diff
@@ -23,23 +23,35 @@ union PrintfPacketData {
...
   void*           ptr;
   const void*     cptr;
-  std::atomic_int ai;
+
+  // Members al, uia, ull used by headers for buffer offset insts.
+  // There are two offsets, the PrintfPacket buffer offset uia[0] and
+  // the Printf string buffer offset uia[1]. We want to update a
+  // single atomic offset al, so that both buffers will increment
+  // in order with respect to each other. This way, offsets can be
+  // managed by single ullong 8B, which is divided into 2 4B uints.
```
whchung: @aaronenyeshi description here doesn't seem to clearly explain the relationship between `al`, `uia[]`, and `ull`. there is no explaination about what `ull` would be used.

include/hc_printf.hpp
```diff
@@ -111,56 +117,91 @@ static inline void copy_n(char* dest, const char* src, unsigned int len) [[hc,cp
...
+  }
+  else if (!queue[PRINTF_STRING_BUFFER].data.ptr || count_char + queue[PRINTF_OFFSETS].data.uia[1] > queue[PRINTF_STRING_BUFFER_SIZE].data.ui){
+    error = PRINTF_STRING_BUFFER_OVERFLOW;
+  }
+  else {
+    do {
+      old_off.ull = queue[PRINTF_OFFSETS].data.al.load();
+      try_off.uia[0] = old_off.uia[0] + count_arg + 1;
+      try_off.uia[1] = old_off.uia[1] + count_char;
+    } while(!(queue[PRINTF_OFFSETS].data.al.compare_exchange_weak(old_off.ull, try_off.ull)));
```
whchung: @aaronenyeshi this relates to my comments earlier in the file. I think the relationship between `ull`, `uia[]`, `al` deserved to be explained a bit better.

whchung (changes requested):

include/hc_printf.hpp
aaronenyeshi: Should test case also check for error cases from printf call on gpu?

whchung: since the test case use FileCheck to check its output perhaps it's fine just to add nullptr check to make sure the tests won't crash, but could capture the case when the implementation returns a nullptr.

aaronenyeshi: Do you mean use an assertion for the case where create returns a nullptr? Or another sort of capture?

whchung: since you use FileCheck to check the output i think it's not good to let the test case crash. so you can catch nullptr and output a different message, and use `CHECK-NOT`

aaronenyeshi: I've committed a change for the test cases to check for nullptr. Just a note, printf, processprintf, and deleteprintf should be guarding against nullptr buffer. So, test case will fail by `CHECK-DAG`, since the test won't crash if createprintf fails.

whchung (approved):

PR #445, opened 2017-08-28
Hcc prof4

bensander:


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1914,7 +1926,7 @@ class HSADevice final : public KalmarDevice
 
     size_t queue_size;
     std::mutex queues_mutex; // protects access to the queues vector:
-    std::vector< std::weak_ptr<KalmarQueue> > queues;
+    std::vector< std::shared_ptr<KalmarQueue> > queues;
```
whchung: @bensander could you help clarify a bit more why we move from `weak_ptr` to `shared_ptr`? the purpose of `weak_ptr` here is to make life-cycle of `KalmarQueue` instances be somewhat independent from `KalmarDevice` but it appears you want to make them be more tightly-coupled by switch to `shared_ptr`?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -3960,12 +3968,12 @@ HSADispatch::dispatchKernel(hsa_queue_t* lockedHsaQueue, const void *hostKernarg
     q_aql->header = header;
 
     hsa_queue_store_write_index_relaxed(lockedHsaQueue, index + 1);
-    DBOUTL(DB_AQL, " dispatch_aql op#" << this->getSeqNum() <<" into " << *hsaQueue() << "(" << lockedHsaQueue << ") " << *q_aql );
+    DBOUTL(DB_AQL, " dispatch_aql op#" << this << "(hwq=" << lockedHsaQueue << ") kernargs=" << hostKernargSize << " " << *q_aql );
     DBOUTL(DB_AQL2, rawAql(*q_aql));
 
     if (DBFLAG(DB_KERNARG)) { 
         // TODO, perhaps someday we could determine size of kernarg block here:
```
whchung: @bensander this comment line should be removed.

whchung (changes requested):

lib/hsa/mcwamp_hsa.cpp
bensander: This is related to profiler which dumps timestamp info when ops are disposed.  Ops have references to queues and this is used in some of the debug information (to print unique op identification issues).   before this change the queue memory would sometimes be deleted before the ops that referenced them.

whchung: @bensander that's the intended behavior of `weak_ptr` because the vector here doesn't own the life-cycle of enclosing items. Consider a case where we allocate 1000 `array_view` on stack. When the function goes out of scope, their corresponding `KalmarQueue` instance would now not be released because they are associated with `KalmarDevice` which only gets destructed at the end of the process.

I'm not feeling comfortable alter this design, just for the sake of profiling. It's possible to used `expired()` to check if the entry within the vector is still valid.

whchung: @bensander could you help change `KalmarAsyncOp` or `HSAOp` to use `weak_ptr` to point the queue?

bensander:
@whchung - please take a fresh look.

whchung (approved):
LGTM

PR #460, opened 2017-09-13
[RUNTIME] simplify kernel name demangling logic

whchung:
Logic in HCC runtime which tries to make demangled kernel names readable would actually obscure the actual string in case the kernel name is complicated. Simplify them in this PR.


scchan (approved):
LGTM

bensander (commented):
I believe the other formats are useful for different use cases.  Mangled for compiler developers, demangled for correlation to source code, and the shortened version for display with profiling tools or if the HIP kernel naming feature is used.  The env var HCC_DB_SYMBOL_FORMAT can be used to choose.  This change removes some of the useful formats.

whchung:
Use `HCC_DB_SYMBOL_FORMAT=2` would retain original demangled kernel name string. Close this PR hereby.

PR #463, opened 2017-09-15
Improve Link Perf by disabling slp-vectorize

aaronenyeshi:
SLP Vectorization is already enabled in Clang frontend ToolChains, so we do not need it for Link time optimizations. This improvement will cut link-time of rocBLAS from 8 mins to 6 mins. Also there is performance increase in run-time of rocblas-test as well.

In detail for compilation time, rocBLAS takes 8 mins for linking, 4 mins for each target. Of the 4 mins, 2 mins spent on opt and 2 mins spent on llc. By removing the slp-vectorize in opt, we will improve opt to run in 1 min.

### DIFF
--- a/lib/clamp-device.in
+++ b/lib/clamp-device.in
@@ -194,7 +194,7 @@ fi
 #  -disable-simplify-libcalls:  prevents transforming loops into library calls such as memset, memcopy on GPU
 # Teach LLVM inliner we do want all functions used by a kernel be inlined
 # by assigning a big iniling threshold value
-$OPT -inline -inline-threshold=1048576 -mtriple amdgcn--amdhsa-amdgiz -mcpu=$AMDGPU_TARGET -infer-address-spaces -amdgpu-internalize-symbols -disable-simplify-libcalls $KMOPTOPT -verify $2.linked.bc -o $2.opt.bc
+$OPT -inline -inline-threshold=1048576 -mtriple amdgcn--amdhsa-amdgiz -mcpu=$AMDGPU_TARGET -infer-address-spaces -amdgpu-internalize-symbols -disable-slp-vectorization -vectorize-slp=false -disable-simplify-libcalls $KMOPTOPT -verify $2.linked.bc -o $2.opt.bc
 
 # error handling for opt
 RETVAL=$?


scchan (changes requested):
Have we had any discussion with the backend folks on whether slp vectorization is useful for the AMDGPU backend?  If it's an important optimization on amdgpu then it would be critical to have after linking the LLVM IRs because that's when more opportunities may be exposed.

whchung:
@scchan actually we already run `SLP vectorizer` when the frontend emits IR in kernel path. Please refer to:
- How Clang determines whether `SLP vectorizer` is needed:
https://github.com/RadeonOpenCompute/hcc-clang-upgrade/blob/clang_tot_upgrade/lib/Driver/ToolChains/Clang.cpp#L4385

- It's enabled when `-O` is larger than 1.
https://github.com/RadeonOpenCompute/hcc-clang-upgrade/blob/clang_tot_upgrade/lib/Driver/ToolChains/Clang.cpp#L648

- We pass `-O2` in kernel path by default.
https://github.com/RadeonOpenCompute/hcc-clang-upgrade/blob/clang_tot_upgrade/lib/Driver/ToolChains/Clang.cpp#L3071

So effectively this PR means we remove one additional `SLP vectorizer` invocation, after we link with ROCDL and inline everything.

scchan:
@whchung I understand we already run SLP before linking.

I think it could be important to run SLP vectorization again after linking and inlining because the code being inlined into a loop may expose more vectorization opportunities.  Also, function calls usually hinder vectorizations therefore it's usually more useful to run that pass after inlining.

whchung:
@aaronenyeshi As another experiment, could you try disable `SLP vectorizer` in the frontend, and enable `SLP vectorizer` at link-time in `clamp-device`?

aaronenyeshi:
Yes, I will give it a try

aaronenyeshi:
@scchan @whchung seems that removing slp-vectorizer in Clang toolchains and using it at link-time works fine. Seems that all tests pass:
```
[----------] 256 tests from rocblas_trsm_scalar_transpose/trsm_gtest (136537 ms total)

[----------] Global test environment tear-down
[==========] 3585 tests from 18 test cases ran. (1421837 ms total)
[  PASSED  ] 3585 tests.```

whchung:
@aaronenyeshi what about the wall clock time of compilation + linking?

aaronenyeshi:
@whchung We have great improvement on the `compilation + linking`. Seems that total `time make` is 15min 10s now. Before it was 18 mins (after duplicate target fix). The 3 mins that was lost was not in `link-time`, `link-time` stays about the same as before. However the `compilation-time` saves almost 3 mins!

whchung:
@aaronenyeshi in such case i think we shall close this PR?

aaronenyeshi:
@whchung Okay, lets close this. Do you think I should open one for the disabling in compile-time and enable in link-time?

PR #470, opened 2017-09-26

vsytch:
This changes removes the need to use hcc-config when compiling with hcc. All of the options (except for 'hc' and 'std=c++amp') added by --cxxflags and --ldflags are now being included inside the hcc tool chain. To invoke hcc in build mode, instead of setting the --build flag in hcc-config, now it is required to set the HCC_BUILD environmental variable.


whchung:
@vsytch could you break this commit into 2 parts:
- 1 for `hcc-clang-upgrade`, file one PR there
- 1 for `hcc`, file one PR here

PR #471, opened 2017-09-26

vsytch:
This change removes the need to use hcc-config when compiling with hcc. All of the options (except for 'hc' and 'std=c++amp') added by --cxxflags and --ldflags are now being included inside the hcc tool chain. To invoke hcc in build mode, instead of setting the --build flag in hcc-config, now it is required to set the HCC_BUILD environmental variable.


scchan (changes requested):
You probably have to add an extra commit to update clang once https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/86 has been merged

PR #473, opened 2017-09-28
[CMAKE] force all hcc libraries to be installed in the lib directory

scchan:
Force all hcc libraries to be installed in the lib directoryon all distros

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -349,6 +349,9 @@ install(PROGRAMS $<TARGET_FILE:LLVMAMDGPUDesc>
         COMPONENT compiler
 )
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -349,6 +349,9 @@ install(PROGRAMS $<TARGET_FILE:LLVMAMDGPUDesc>
         COMPONENT compiler
 )
 
+# force library install path to lid
```
whchung: typo: `lib`

whchung (changes requested):
typo

scchan:
typo fixed

whchung (approved):

PR #477, opened 2017-10-01
Fix memory fence handling for barriers

bensander:
- release_scope now fence_scope and controls both acquire and release fence.
- override aquire fence on dispatch_hsa_kernel path, if needed.


whchung (approved):
LGTM

PR #485, opened 2017-10-05

vsytch:
This change removes the need to use hcc-config when compiling with hcc. All of the options (except for 'hc' and 'std=c++amp') added by --cxxflags and --ldflags are now being included inside the hcc tool chain. To invoke hcc in build mode, instead of setting the --build flag in hcc-config, now it is required to set the HCC_BUILD environmental variable.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -316,6 +316,7 @@ if(EXISTS ${PROJECT_SOURCE_DIR}/clang-tools-extra/CMakeLists.txt)
   set(LLVM_EXTERNAL_CLANG_TOOLS_EXTRA_SOURCE_DIR "${PROJECT_SOURCE_DIR}/clang-tools-extra")
 endif()
 
+include_directories("${CMAKE_CURRENT_BINARY_DIR}/hcc_config")
 add_subdirectory(${CLANG_SRC_DIR})
 
 install(PROGRAMS $<TARGET_FILE:llvm-as>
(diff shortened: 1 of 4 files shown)


whchung:
@vsytch thanks for this revised PR. One thing i'd like to understand: `make test` would originally fetch include path and lib path from `--build`, but in this PR `HCC_BUILD` would have to be used instead, so `lit.cfg` is also changed in this PR.

I know probably you are working on a different platform, but would you mind post `make test` result on Linux?

whchung:
just for recording purpose this PR should be merged alongside with https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/90

vsytch:
Here are the results https://pastebin.com/h7DYmA7Z. I also ran `make conformance` today with and without my changes and failed the same amount of tests.

scchan:
Have you tried "make package", install the .deb package and then test with the hcc from its installed location?

scchan:
@whchung  perhaps this line: https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/90/files#diff-24e9f257506077abe4b6698eb50248dcR344
which gets the relative path to the libraries/headers from clang's path.  
@vsytch can you confirm?

vsytch:
@scchan yes C.getDriver().Dir gets the path to the directory from which Clang was invoked.

vsytch:
@scchan I ran hcc from the install location and got an error. I think https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/90/files#diff-24e9f257506077abe4b6698eb50248dcR344 should just be `path_hcc_lib = C.getDriver().Dir + "/../" + CMAKE_INSTALL_LIB;`

scchan:
@vsytch I merged https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/95, so you should update `clang` in this PR

whchung:
@scchan I believe there is no `clang` submodule in this PR?

scchan:
@whchung   right,  @vsytch needs to add the `clang` module to this PR

PR #487, opened 2017-10-06
Fix cross-queue flush synchronization.

bensander:
- Barriers capture the flushes required after the barrier executes.
  A barrier enqueued in another stream which is dependent on this
  barrier will pick up the flush state from the barrier (not the queue
  since other older ops may have been enqueued, including for example
  an op which changes the flush state).

- Add HCC_FORCE_CROSS_QUEUE_FLUSH
- Add debug information on DB_CMD2 for flush operations.
- Retab and clean up formatting (growKernargBuffer, others)


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -617,17 +618,26 @@ class HSABarrier : public HSAOp {
...
+    bool                                            _barrierNextSyncNeedsSysRelease;
+    bool                                            _barrierNextKernelNeedsSysAcquire;
+
 public:
     uint16_t  header;  // stores header of AQL packet.  Preserve so we can see flushes associated with this barrier.
 
     // array of all operations that this op depends on.
     // This array keeps a reference which prevents those ops from being deleted until this op is deleted.
-    std::shared_ptr<HSAOp> depAsyncOps [HSA_BARRIER_DEP_SIGNAL_CNT];
+    std::shared_ptr<HSABarrier> depAsyncOps [HSA_BARRIER_DEP_SIGNAL_CNT];
```
whchung, line 632: @bensander this means all ops (kernel/copy/barrier) would only depend on barrier ops, correct?

whchung:
working on application-level validation.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -104,6 +104,7 @@ int HCC_SERIALIZE_COPY = 0;
 int HCC_FORCE_COMPLETION_FUTURE = 0;
 
 int HCC_OPT_FLUSH=1;
+int HCC_FORCE_CROSS_QUEUE_FLUSH=1;
```
whchung: this value should be 0 by default?

whchung (changes requested):

whchung (approved):

PR #491, opened 2017-10-09
[RUNTIME] revise kernarg pool release algorithm

whchung:


bensander (commented):
LGTM, commit definitely makes sense given how kerning pool is allocated.

scchan (approved):
The name KERNARG_POOL_SIZE is confusing and I think it would be clearer to call it to something like NUM_CHUNKS_PER_ALLOCATION.

scchan:
@whchung do we need this for 1.6?

PR #492, opened 2017-10-09
Check asyncOpsIndex is in-bounds of asyncOps.

bensander:


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1899,38 +1901,44 @@ class HSAQueue final : public KalmarQueue
...
-            } else {
-                // The queue is retired in-order, and ops only inserted at "top", and ops can only be removed at two defined points:
-                //   - Draining the entire queue in HSAQueue::wait() - this calls asyncOps.clear()
-                //   - Events in the middle of the queue can be removed, but will call this function which removes all older ops.
-                //   So once we remove the asyncOps, there is no way for an older async op to be come non-null and we can stop search here:
-                //
-
-                break; // stop searching if we find null, there cannot be any more valid pointers below.
-            }
+		int nullifiedCount = 0;
```
whchung: could you re-indent the file?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1899,38 +1901,44 @@ class HSAQueue final : public KalmarQueue
...
+			// All older ops are known to be done and we can reclaim their resources here:
+			// Both execute_in_order and execute_any_order flags always remove ops in-order at the end of the pipe.
+			// Note if not found above targetIndex=-1 and we skip the loop:
+			for (int i = targetIndex; i>=0; i--) {
+				Kalmar::KalmarAsyncOp *op = asyncOps[i].get();
+				if (op) {
+					nullifiedCount++;
+					asyncOps[i] = nullptr;
+
+	#if CHECK_OLDER_COMPLETE
```
whchung: since `CHECK_OLDER_COMPLETE` is 0 by default, would we still have performance gain with this PR? I know using `asyncOp->asyncOpsIndex()` in the beginning of the function itself should already give us some perf boost.

whchung (changes requested):

lib/hsa/mcwamp_hsa.cpp
bensander: yes.  The big win is eliminating the loop that finds the targetIndex, which was in the previous checkin and is preserved here.  We might see a small gain here since we don't query the signal in the normal "fast" mode

lib/hsa/mcwamp_hsa.cpp
bensander: Thx Done!

whchung (approved):

PR #494, opened 2017-10-10
update hc::printf

scchan:
- pick up fixes for pfe lambda function


scchan:
replaced by https://github.com/RadeonOpenCompute/hcc/pull/495

PR #495, opened 2017-10-10
update hc::printf

scchan:
- pick up fixes for pfe lambda function

### DIFF
--- a/lib/hsa/hc_am.cpp
+++ b/lib/hsa/hc_am.cpp
@@ -1,5 +1,9 @@
+
+#include "hc.hpp"
 #include "hc_am.hpp"
 
+#include <cstddef>
+#include <mutex>
 #include <cstdint>
 #include <iomanip>
 #include <hsa/hsa.h>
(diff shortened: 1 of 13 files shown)


lib/hsa/hc_am.cpp
```diff
@@ -524,8 +528,9 @@ am_status_t am_map_to_peers(void* ptr, size_t num_peer, const hc::accelerator* p
         else
             return AM_ERROR_MISC;
     }
-    
-    std::vector<hsa_agent_t> agents{hc::accelerator::get_all().size()};
+
+    const std::size_t max_agent = hc::accelerator::get_all().size();
+    hsa_agent_t agents[max_agent];
```
whchung, line 533: please resync with `clang_tot_upgrade` branch. there is a recent PR which fixed this VLA.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -271,6 +271,22 @@ static unsigned extractBits(unsigned v, unsigned pos, unsigned w)
...
 
+
+
+namespace hc {
+
+// printf buffer size (in number of packets, see hc_printf.hpp)
+constexpr unsigned int default_printf_buffer_size = 2048;
+
+// address of the global printf buffer in host coherent system memory
+PrintfPacket* printf_buffer = nullptr;
```
whchung, line 282: @scchan would it be possible to make it a member variable of `HSAContext`, as its life-cycle is tied to it?

whchung (changes requested):

scchan:
closing for now due to a couple of changes needed based on feedback

lib/hsa/mcwamp_hsa.cpp
scchan: I'd prefer to leave it here.  One of the reasons is that the device function hc::printf has a reference to this global object and if I put this inside HSAContext then we will have to include the HSAContext definition when we compile device code, which I'd try to avoid.

PR #496, opened 2017-10-10
Partial support for namespace scope variables.

AlexVlx:
This adds incomplete support for generalised C++ language namespace scope variables, relying on ELF introspection and relevant HSART interfaces. At the moment there is an issue with observing accelerator writes to a global correctly, and support for shared objects using a global defined in main is unimplemented. The latter would require quite a bit of backporting from HC2, which unless proven to be utterly necessary can be eschewed.


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -271,6 +271,143 @@ static unsigned extractBits(unsigned v, unsigned pos, unsigned w)
...
+    {
+        using namespace ELFIO;
+        using namespace std;
+
+        if (!code_object_dynsym || !process_symtab) return;
+
+        const auto undefined_symbols = copy_names_of_undefined_symbols(
+            symbol_section_accessor{reader, code_object_dynsym});
+
+        for (auto&& x : undefined_symbols) {
```
scchan, line 417: Maybe we should check whether `x` has already been defined for this `agent` and if it has already been defined, then skip to the next undefined symbols?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -271,6 +271,143 @@ static unsigned extractBits(unsigned v, unsigned pos, unsigned w)
...
+            const auto tmp = find_symbol_address(
+                symbol_section_accessor{self_reader, process_symtab}, x);
+
+            assert(tmp.first);
+
+            void* p = nullptr;
+            hsa_amd_memory_lock(
+                reinterpret_cast<void*>(tmp.first), tmp.second, &agent, 1, &p);
```
scchan: Should we check whether the host variable has already been "pinned" previously?

tests/Unit/NamespaceScopeVariables/global.cpp
```diff
@@ -0,0 +1,53 @@
...
+        read_scalar[0] = global_scalar;
+        for (auto i = 0u; i != array_size; ++i) read_array[i] = global_array[i];
+    });
+
+    if (read_scalar[0] != global_scalar) return EXIT_FAILURE;
+    if (!equal(global_array, global_array + array_size, read_array.data())) {
+        return EXIT_FAILURE;
+    }
+
+    #if false // 10/09/2017 - writes to globals are not correctly observed.
```
scchan, line 34: any idea why this doesn't work

scchan (changes requested):

tests/Unit/NamespaceScopeVariables/global.cpp
AlexVlx: @scchan not at this moment, and unless there is a strong need for that now (I do not believe that is the case, but I may be wrong) I would punt on it. Mainly due to time constraints and it working in HC2. It is your call though.

lib/hsa/mcwamp_hsa.cpp
AlexVlx: We could, but this would only be beneficial if we had multiple code objects per agent (otherwise having multiple unmerged symbols with the same name in the same code object / TU would be an ODR violation / not possible), which I do not believe we do at this point in time. The underlying assumption here is that the mapping from code-object undefined symbol to loader allocated host symbol is bijective. However, I may be missing something about how "executables" is managed at the moment...

lib/hsa/mcwamp_hsa.cpp
AlexVlx: We could, but this would require either using a slightly more complex data structure or O(n) searching on each lock. If the mapping is bijective, this appears unnecessary, for there can be no scenario in which multiple different symbols try to lock the same loader allocated variable in the executable's image. If we were to have multiple code-objects containing code which references the same global, and which are loaded during the same execution, this may be an optimisation worth pursuing. At the moment we do not support it, but a scenario for it would be both the executable and shared object(s) referencing a global that is defined in a shared object.

lib/hsa/mcwamp_hsa.cpp
scchan: If an application is linked to a shared library with GPU code then it's possible that an agent will have more than one code objects (i.e. one from the application's objects and another from the shared library).

lib/hsa/mcwamp_hsa.cpp
scchan: Any reason why we don't support multiple objects referencing the same global?  It seems to be that your algorithm should just work and we just need verify whether ROCR is able to handle double locking/unlocking the same global.  If ROCR couldn't handle it properly than hcc RT would need to manage the memory lock/unlock more carefully.

whchung (dismissed):
to begin with, I think this PR deserves a better 1-line summary? "Add missing files to prior commit." is misleading.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -271,6 +272,207 @@ static unsigned extractBits(unsigned v, unsigned pos, unsigned w)
...
+
+            if (globals.find(x) == globals.cend()) {
+                const auto it1 = symbol_addresses().find(x);
+
+                if (it1 == symbol_addresses().cend()) {
+                    throw runtime_error{
+                        "Global symbol: " + x + " is undefined."};
+                }
+
+                if (globals.find(x) == globals.cend()) {
```
scchan: I don't see `x` being added to `globals` or maybe I'm missing something?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -271,6 +272,207 @@ static unsigned extractBits(unsigned v, unsigned pos, unsigned w)
...
+                    static mutex mtx;
+                    lock_guard<mutex> lck{mtx};
```
scchan: I presume this mutex is guarding `globals` for the insertion of `x` so shouldn't the lock comes before the 2nd `find(x)`?

scchan (changes requested):

lib/hsa/mcwamp_hsa.cpp
AlexVlx: I guess I was mentally transporting the electrons:) Thank you, fixed.

lib/hsa/mcwamp_hsa.cpp
AlexVlx: Not quite; `find` is `const`, therefore thread safe, doing a const read under a lock would be a pessimisation. If we really did not find it (i.e. nobody added it in between our second search) then we actually go and lock mtx. Since we already reserved the maximum possible amount of memory, the insertion would not invalidate any iterators so the return value of find or cend would not be rendered bogus by another thread doing an insertion. Bad things will happen if some detached thread somewhere actually tries to do this, since it may be running after main and static destruction have returned, but that's a different story altogether and a programmer error.

scchan: Don't you need to prevent another thread inserting the same `x` between the 2nd `find(x)` and the lock?

AlexVlx: I was about to say that that is benign since it would only yield multiple calls to hsa_amd_memory_lock, but that's incorrect since the destruction of the temporary would call hsa_amd_memory_unlock on possibly the same, already inserted p. Thus you are correct; fixed it, thank you.

scchan (approved):
LGTM

scchan:
@whchung do you have other concerns regarding this PR?

whchung:
@scchan please go ahead and merge this PR

PR #498, opened 2017-10-10
update hc::printf

scchan:
- making hc::printf an opt-in feature; user will have to compile with  `-DHCC_ENABLE_ACCELERATOR_PRINTF` to enable it

### DIFF
--- a/include/hc_printf.hpp
+++ b/include/hc_printf.hpp
@@ -341,16 +341,21 @@ static inline void processPrintfPackets(PrintfPacket* packets, const unsigned in
 
 static inline void processPrintfBuffer(PrintfPacket* gpuBuffer) {
 
-  if (gpuBuffer == NULL) return;
-  unsigned int bufferSize = gpuBuffer[PRINTF_BUFFER_SIZE].data.ui;
+  if (gpuBuffer == nullptr) return;
+
   unsigned int cursor = gpuBuffer[PRINTF_OFFSETS].data.uia[0];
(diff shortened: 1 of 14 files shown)


include/hc_printf.hpp
```diff
@@ -286,6 +336,7 @@ static inline void processPrintfPackets(PrintfPacket* packets, const unsigned in
     formatString = std::regex_replace(formatString,doubleAmpersandPattern,"%");
     std::printf("%s",formatString.c_str());
   }
+  std::flush(std::cout);
 }
 
 static inline void processPrintfBuffer(PrintfPacket* gpuBuffer) {
```
whchung, line 342: @scchan if i understand the logic correctly, `processPrintBuffer` would still be invoked with a non-`nullptr` `gpuBuffer`. would it be possible to further optimize this?

whchung (changes requested):
@scchan with the new flag now the feature is optional, which would definitely speed up kernels when `printf()` is not used, but it seems print buffer flushing would still take place on the host side.  is it possible to also optimize this?

include/hc_printf.hpp
scchan: If you refer to the following line, the host would first check whether the printf buffer contains any new data before proceeding to do further processing.  If kernels in the applicaton doesn't print anything the overhead should be minimal

https://github.com/RadeonOpenCompute/hcc/pull/498/files#diff-531a057446b7d96d716399f005b31ad8R349

whchung: got it. with the revised logic it's more clear.

whchung (approved):

PR #508, opened 2017-10-18
Create a function to re-initialize the printf buffer 

scchan:
- HIP's hipDeviceReset() may destroy the printf buffer. Create a backdoor for HIP to re-initialize the printf buffer after a hipDeviceReset()

- pinned the pointer to the printf buffer instead of allocating another piece of coherent system memory of it, this removes one more item that could be destroyed by hipDeviceReset()


whchung (approved):
LGTM

PR #515, opened 2017-10-25
[Jenkins] wait for HIP integration

scchan:


Jenkinsfile
```diff
@@ -199,7 +199,7 @@ node( 'rocmtest' )
     // If this a clang_tot_upgrade build, kick off downstream hip build so that the two projects are in sync
     if( env.BRANCH_NAME.toLowerCase( ).startsWith( 'clang_tot_upgrade' ) )
     {
-      build( job: 'ROCm-Developer-Tools/HIP/master', wait: false )
+      build( job: 'ROCm-Developer-Tools/HIP/master', wait: true )
```
kknox, line 202: The only comment here is that without the boolean parameter, the hip job will launch all 3 build configurations (hcc-tot,hcc-rocm,nvcc).  I think this is what we want when clang_tot_upgrade branch changes.

kknox (approved):
I left a comment, but no need for an action.  When clang_tot_upgrade changes, all 3 build variations of HIP will fire off.  This might take slightly longer, but keeps all the configurations in sync for version numbers.

bensander:
If clang-tot-upgrade changes, we would only want to re-run the hcc-tot HIP config correct?
If anything in HIP changes, we want to run all 3.
Not sure if we have the flexibility to support both of these?

kknox:
@bensander 
The primary reason to kick off a full build of HiP is because of the stringent version checks that HiP uses.  If clang_tot_upgrade changed, the version number of hcc might change, and we want to build hip packages that can pass the version checks.  

It is true, that **really** the only thing that you need to test is hip-ctu build, and that is what we do for PR branches (controlled through the build parameter).  For the main branch ctu build, I think we want to build everything so that the install packages are all in sync.

PR #517, opened 2017-10-26
Add ptr field to am_memtracker_update

bensander:
Also add hcc_features.hpp file.

### DIFF
--- a/include/hc.hpp
+++ b/include/hc.hpp
@@ -23,6 +23,7 @@
 
 #include "hsa_atomic.h"
 #include "kalmar_cpu_launch.h"
+#include "hcc_features.hpp"
 
 #ifndef __HC__
 #   define __HC__ [[hc]]
--- a/include/hc_am.hpp
+++ b/include/hc_am.hpp
@@ -167,7 +167,7 @@ am_status_t am_memtracker_add(void* ptr, hc::AmPointerInfo &info);
  *
  * @see am_memtracker_getinfo, am_memtracker_add
  */
-am_status_t am_memtracker_update(const void* ptr, int appId, unsigned allocationFlags);
+am_status_t am_memtracker_update(const void* ptr, int appId, unsigned allocationFlags, void *appPtr=nullptr);
 
 
 /** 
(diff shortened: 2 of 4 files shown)


whchung (changes requested):
@bensander could you explain the reason to use "build"? I'm not particular fond of adding extra public header, let alone making it potentially be convoluted with build directories. Also I think the value inside the header seem to be better defined at config-time, so a cmake template file makes more sense

bensander:
Intent is to replace use of workweek for compiler feature detection.  Hip is example client but could be other apps that want to condition compilation based on features of the compiler.

On Oct 25, 2017, at 9:29 PM, Wen-Heng (Jack) Chung <notifications@github.com<mailto:notifications@github.com>> wrote:


@whchung requested changes on this pull request.

@bensander<https://github.com/bensander> could you explain the reason to use "build"? I'm not particular fond of adding extra public header, let alone making it potentially be convoluted with build directories. Also I think the value inside the header seem to be better defined at config-time, so a cmake template file makes more sense

scchan:
@bensander I think I understand the intent of having hcc_features.hpp
but is there a reason for putting it in build/include/hcc_features.hpp instead of include/hcc_features.hpp?

Also, I think hc.hpp should include hcc_features.hpp so that users don't have to include this header explicitly?

scchan (changes requested):
need to resolve placement of hcc_features.hpp

scchan:
@bensander this commit shows the author as Aditya perhaps your git environment has a config issue?

bensander:
It should be in include/hcc_features.hpp.  I will update the PR.  May still leave Aditya as the author. ☺

bensander:
Updated!

whchung:
@bensander now the location of `hcc_features.hpp` have been moved. as mentioned by @scchan earlier in the thread, would you consider adding `#include <hcc_features.hpp>` inside `hc.hpp`, so user programs don't to explicitly include it?

whchung:
@bensander per our discussion:

- use numerical values in `hcc_features.hpp`. and let client programs such as `hip` to use `#if defined() + #if` to retrieve the value
- add `#ifndef` guards in `hcc_features.hpp`.
- rename flag values with `__` prefix.
- include `hcc_features.hpp` inside `hc.hpp`.

scchan (approved):

whchung (approved):

PR #518, opened 2017-10-26
Fix hcc not being created in the correct directory

vsytch:


CMakeLists.txt
```diff
@@ -331,6 +331,9 @@ install(PROGRAMS $<TARGET_FILE:llvm-as>
 add_custom_target(clang_links DEPENDS clang)
 add_custom_command(TARGET clang_links POST_BUILD
   COMMAND ${CMAKE_COMMAND} -E create_symlink ${CLANG_BIN_DIR}/lib/clang/6.0.0/lib/linux/libclang_rt.builtins-x86_64.a ${PROJECT_BINARY_DIR}/lib/libclang_rt.builtins-x86_64.a
+  COMMAND ${CMAKE_COMMAND} -E create_symlink ${CLANG_BIN_DIR}/bin/clang-6.0 ${PROJECT_BINARY_DIR}/bin/clang
```
scchan: should we create a link to ${CLANG_BIN_DIR}/bin/clang instead?  otherwise we'll have to update this whenever the version gets updated.

scchan (changes requested):

CMakeLists.txt
vsytch: We have ${CLANG_BIN_DIR}/lib/clang/6.0.0/lib/linux/libclang_rt.builtins-x86_64.a hardcoded, so we would have to update that manually too.

scchan: yes, that would be nice

whchung (approved):
LGTM

whchung:
@scchan I believe this could be merged? #522 addresses hard-coded `x86_64` dependency.

scchan (approved):

PR #519, opened 2017-10-27
create a new feature flag for printf in hcc_features.hpp

scchan:


whchung (approved):
LGTM. Though I believe @AlexVlx may have other opinion.

PR #521, opened 2017-10-28
Correctly handle non-Ubuntu platforms in os-release

ChrisKitching:
Without quotes, this expands to nothing and the functions are
deemed to be being called with too few arguments. With quotes,
it's empty-string. Which is safe.

In general, quotes should always be used like this in cmake...


whchung (approved):
LGTM. Thanks for the contribution!

PR #523, opened 2017-10-30
Fix unit tests for gfx900

whchung:
Fix unit tests on test machines equipped with `gfx900`.

### DIFF
--- a/tests/Unit/AMDGPU/register-control.cpp
+++ b/tests/Unit/AMDGPU/register-control.cpp
@@ -1,5 +1,6 @@
-// RUN: %hc %s -o %t.out -Xlinker -dump-llvm -Xlinker -dump-dir=%T --amdgpu-target=gfx803
+// RUN: %hc %s -o %t.out -Xlinker -dump-llvm -Xlinker -dump-dir=%T %target_all_gpus
 // RUN: %llvm-dis %T/dump-gfx803.opt.bc -f -o - | %FileCheck %s
+// RUN: %t.out
 
 #include <hc.hpp>
 #include <vector>
--- a/tests/Unit/AMDGPU/register-control.cpp
+++ b/tests/Unit/AMDGPU/register-control.cpp
@@ -1,6 +1,5 @@
 // RUN: %hc %s -o %t.out -Xlinker -dump-llvm -Xlinker -dump-dir=%T --amdgpu-target=gfx803
 // RUN: %llvm-dis %T/dump-gfx803.opt.bc -f -o - | %FileCheck %s
-// RUN: %t.out
 
 #include <hc.hpp>
 #include <vector>
--- a/tests/Unit/HC/multi_acc2.cpp
+++ b/tests/Unit/HC/multi_acc2.cpp
@@ -1,4 +1,4 @@
-// RUN: %hc --amdgpu-target=gfx701 --amdgpu-target=gfx801 --amdgpu-target=gfx802 --amdgpu-target=gfx803 --amdgpu-target=gfx900 %s -o %t.out && %t.out
+// RUN: %hc %target_all_gpus %s -o %t.out && %t.out
 #include <random>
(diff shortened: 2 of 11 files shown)


tests/Unit/AMDGPU/register-control.cpp
```diff
@@ -1,6 +1,5 @@
 // RUN: %hc %s -o %t.out -Xlinker -dump-llvm -Xlinker -dump-dir=%T --amdgpu-target=gfx803
 // RUN: %llvm-dis %T/dump-gfx803.opt.bc -f -o - | %FileCheck %s
-// RUN: %t.out
```
scchan: should we remove `--amdgpu-target=gfx803` and rely on the enumerator to decide on the arch?

tests/Unit/HC/multi_acc2.cpp
```diff
@@ -1,4 +1,4 @@
-// RUN: %hc --amdgpu-target=gfx701 --amdgpu-target=gfx801 --amdgpu-target=gfx802 --amdgpu-target=gfx803 %s -o %t.out && %t.out
```
scchan, line 1: maybe we should remove the static list of gpu and just use the enumerator?

scchan (changes requested):

tests/Unit/AMDGPU/register-control.cpp
whchung: @scchan I don't like to make `hcc` tests depend on `rocm_agent_enumerator`, as it's another tool not owned/built alongside with `hcc`.  Plus it appears the source code of `rocm_agent_enumerator` isn't posted online so it makes porting it to another architecture hard.

tests/Unit/HC/multi_acc2.cpp
whchung: As mentioned in my previous comment I'd like to keep `hcc` be less dependent to `rocm_agent_enumerator`.

tests/Unit/AMDGPU/register-control.cpp
scchan: maybe we should define a new substitution pattern in lit.cfg such that it would replace it with a string including all the GPU architectures?  When we have to add a new architecture in the future, we'll just have to put that in one file.

https://github.com/RadeonOpenCompute/hcc/blob/clang_tot_upgrade/tests/lit.cfg#L142

whchung: @scchan I've thought that, but problem with `rocm_agent_enumerator` is very peculiar:

- unlike most other tools specified in `lit.cfg`, it is NOT owned by `hcc`, so we don't usually know its existence or whereabout.
- unlike `c++filt` which is normally placed under `$PATH`, `rocm_agent_enumerator` is usually installed at `/opt/rocm/bin`, which may usually not be in `$PATH`
- and on a new architecture, `rocm_agent_enumerator` may not even install on `/opt/rocm/bin` and could be altered by `ROCM_ROOT` env var

due to these peculiarities of `rocm_agent_enumerator`, I don't feel like adding an extra dependency to it in `hcc` lit tests.

scchan: my previous comment was about creating something like:

`%target_all_gpus == "--amdgpu-target=gfx701 --amdgpu-target=gfx801 --amdgpu-target=gfx802   --amdgpu-target=gfx900" `

in lit.cfg so that someone could use %target_all_gpus in the unit tests.
It's not about using the enumerator.

whchung: makes sense. let me revise this PR.

bensander:
perhaps could support a file with the same format as the agent-enumerator, so it could be generated externally by "rocm_agent_enumerator > targets" or manually create target for the alternate archs ?

scchan (approved):

PR #533, opened 2017-11-03
Fix headers not being correctly copied to the build directory

vsytch:
Currently headers under `hcc/include/experimental/impl` are being copied to `build/include/experimental/impl/impl`. With this change they are copied to `build/include/experimental/impl`.


whchung (approved):
LGTM. thanks for spotting this.

PR #535, opened 2017-11-06
Add ThinLTO implementation of Cross-module importing

aaronenyeshi:
Here is the Phase 1 of implementing ThinLTO into HCC. Currently this implementation uses ThinLTO by maintaining kernels in their separate .bcs files, creates their module-summaries, performs llvm-lto’s cross-module function importing and then performs clamp-device (opt and llc tools) on each of those kernel files. Then links with lld into single .hsaco per target.

This implementation will greatly speed up build times of some user projects.


aaronenyeshi:
also, this PR requires https://github.com/RadeonOpenCompute/llvm/pull/110, so it needs to update the submodule configuration for compiler/ submodule.

whchung (changes requested):
since the feature needs more validation, could you revise the PR so the feature is optional by default, and be enabled via a compile-time flag, or an environment variable?

also since submodule configuration was changed please help rebase.

whchung:
@aaronenyeshi the if logic based on `KMTHINLTO` begins to make the code complex and hard to understand. please help:

1) use functions to separate logic for `ThinLTO` path and conventional path.
2) add comments to describe the differences between `ThinLTO` path and conventional route. For example, why `ld.lld` is now not needed in `clamp-device`.

scchan (changes requested):
This PR failed to build, could you take a look?


[100%] Linking CXX static library ../lib/libamptest.a
[100%] Built target amptest
[100%] Linking CXX executable ../bin/cmake-test
clang-6.0: error: unable to execute command: Executable "clamp-link" doesn't exist!
clang-6.0: error: linker command failed with exit code 1 (use -v to see invocation)
cmake-tests/CMakeFiles/cmake-test.dir/build.make:98: recipe for target 'bin/cmake-test' failed
make[2]: *** [bin/cmake-test] Error 1

aaronenyeshi:
Sorry, seems like clamp-link.in file got converted from `lib/clamp-link.in 100755 → 100644` . Which means it changed unix file permission modes and dropped the +x (executable) flag.

whchung (approved):
LGTM

PR #538, opened 2017-11-08
Don't use hcc-config for unit tests

vsytch:
This is a follow up change to https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/100.

### DIFF
--- a/tests/lit.cfg
+++ b/tests/lit.cfg
@@ -76,22 +76,12 @@ def inferClang(PATH):
 
 cxx_options = ' ' + ' '.join([
   "-I%s" % config.gtest_src_dir,
-  "-DGTEST_HAS_TR1_TUPLE=0",
-  subprocess.Popen([
-      os.path.join(config.executable_output_path, 'clamp-config'),
-      "--build",
-      "--cxxflags"],
-      stdout=subprocess.PIPE).communicate()[0].rstrip('\n'),
+  "-DGTEST_HAS_TR1_TUPLE=0"
 ]) + ' '
 
 link_options = ' ' + ' '.join([
-  subprocess.Popen([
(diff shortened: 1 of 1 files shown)


scchan (commented):
LGTM but before I approve this, you'll have to submit a PR to pick up the clang driver change.

scchan (changes requested):
- there's a conflict with clang so please resolve it
- don't you need to update cmakefile for compiling the hcc RT files?

vsytch:
I checked our cmake files, we don't add -hc when linking mcwamp and hc_am, so this change doesn't affect them.

scchan:
then where is fno-hcc-lib-link needed?

vsytch:
It's not needed anywhere in the build files as of right now.

scchan:
I'm confused so we don't need the fno-hcc-lib-link option at all?

vsytch:
I guess we can get rid of it for now them? But then we don't give the user an option if they don't want to link to libmcwamp/libhc_am for whatever reason.

scchan:
yes, please submit another PR to remove it

scchan:
some of the unit tests failed, a few of them are probably regressions from other commits, could you take a look at these ones?

    HCC :: Unit/RawGenericPointer/array_add_am14.cpp
    HCC :: Unit/RawGenericPointer/array_add_am15.cpp
    HCC :: Unit/RawGenericPointer/array_add_am16.cpp
    HCC :: Unit/RawGenericPointer/array_add_am9.cpp

scchan:
ignore my previous comment, I don't think this PR is causing those regressions

scchan (approved):

PR #539, opened 2017-11-08
Fix hybrid arch code object loading

whchung:
This PR addresses a peculiar issue on machines with hybrid GPU targets (ex: `gfx803` + `gfx900`):

When a program is built to support only a particular GPU target, but not other available ones on the system, then it would fail to launch. This hinders development / testing for HCC/HIP applications which use assembly codes only available on certain targets.

The current behavior inside HCC runtime is:
 
- HCC runtime tries to eagerly do:
  - Initialize ROCR runtime at bootstrap
  - Load code objects for each agent
  - Aborts when there is no compatible code object found for an agent
- `HCC_LAZYINIT=ON` is broken right now as no code objects would then be ever loaded.

 
This PR changes the behavior of HCC runtime to:
- HCC runtime tries to eagerly:
  - initialize ROCR runtime at bootstrap
  - Load code objects for each agent, but don’t abort when there is no compatible code object for an agent
  - `HCC_LAZYINIT` defers above initialization until 1st kernel dispatch
- Try load code object at 1st kernel dispatch, on an agent in case it wasn’t done yet
- The program would still abort in `CreateKernel` in case incorrect agent is used to dispatch a kernel (ex: try run a binary with `gfx900`-only kernels on a `gfx803` agent).

No test case is given in this PR because although it's easy to come up one like: https://github.com/adityaatluri/gemm-vega64 , but it's hard to create a test environment for this configuration with hybrid targets for now.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -341,6 +341,10 @@ extern "C" void PushArgPtrImpl(void *ker, int idx, size_t sz, const void *v);
 namespace Kalmar {
 class HSAQueue;
 class HSADevice;
+
+namespace CLAMP {
+  void LoadInMemoryProgram(KalmarQueue*);
+} // namespace CLAMP
 } // namespace Kalmar
 
 ///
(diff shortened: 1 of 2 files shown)


AlexVlx:
@whchung LGTM, thanks.

scchan (approved):
LGTM

PR #550, opened 2017-11-15
added new docker file for upstream merge

david-salinas:
Adding a new docker file to be used by upstream merge tool.  Based on the HCC-build docker file, but we've added rocm-utils, and the Hub (git wrapper) tool.


whchung (approved):
LGTM.

PR #552, opened 2017-11-18
Add pointer info check for D2H path.

bensander:
- Similar to check which already existed on H2D path.  If the pointer
  is already locked, we can't use pin-in-place.
- Add GetPointerInfo and refactor calls to hsa_amd_pointer_info.
- Add debug to unpinned copy engine calls.

### DIFF
--- a/lib/hsa/unpinned_copy_engine.cpp
+++ b/lib/hsa/unpinned_copy_engine.cpp
@@ -218,7 +218,7 @@ void UnpinnedCopyEngine::CopyHostToDevice(UnpinnedCopyEngine::CopyMode copyMode,
 {
     bool isLocked = false;
     if((copyMode == ChooseBest) || (copyMode == UsePinInPlace)) {
-        isLocked = GetPointerInfo(src);
+        isLocked = IsLockedPointer(src);
     }
     if (copyMode == ChooseBest) {
         if (_isLargeBar && (sizeBytes < _hipH2DTransferThresholdDirectOrStaging)) {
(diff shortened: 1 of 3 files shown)


lib/hsa/unpinned_copy_engine.cpp
```diff
@@ -525,3 +522,23 @@ void UnpinnedCopyEngine::CopyPeerToPeer(void* dst, hsa_agent_t dstAgent, const v
         hsa_signal_wait_acquire(_completionSignal2[i], HSA_SIGNAL_CONDITION_LT, 1, UINT64_MAX, HSA_WAIT_STATE_ACTIVE);
     }
 }
+
+
+bool UnpinnedCopyEngine::GetPointerInfo(const void *ptr)
```
scchan: This function should be renamed because it's not getting the pointer info but instead checking whether a pointer is locked.

lib/hsa/unpinned_copy_engine.cpp
```diff
@@ -358,16 +350,21 @@ void UnpinnedCopyEngine::CopyDeviceToHostPinInPlace(void* dst, const void* src,
...
+        if (sizeBytes > _hipD2HTransferThreshold && !isLocked) {
             copyMode = UsePinInPlace;
         } else {
             copyMode = UseStaging;
         }
     }
 
 
-	if (copyMode == UsePinInPlace) {
+	if (copyMode == UsePinInPlace && !isLocked) {
```
scchan: indentation

scchan (changes requested):

scchan:
@bensander ping...

lib/hsa/unpinned_copy_engine.cpp
bensander: I'm Ok with the proposed changes, could use some help if possible to actually implement them.

PR #553, opened 2017-11-18
Add env var HCC_SIGNAL_POOL_SIZE.

bensander:
- Env var allows runtime control over the size of the initial signal
  pool.
- Remove fallback path - assume we are using signal pool.
- Remove nullify DB printf that was on DB_RESOURCE


scchan (approved):
LGTM

PR #558, opened 2017-11-20
Use hcc in scripts instead of clang

tstellar:
hcc is a symlink to clang, so this is effectively the same.  It also
makes it possible to install an hcc executable in the same bindir
as system clang.


scchan (approved):
LGTM

PR #559, opened 2017-11-20
Smarter handling if queue contains only "no-signal" ops.

bensander:
Commands can be submitted w/o a completion signal.
Now if queue contains only these "no-signal" ops and isEmpty()
is called, then we add a barrier and return false.  Subsequent
calls to isEmpty() will test the signal associated with the barrier
to see if the queue is empty.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -691,18 +691,21 @@ public:
 
     void asyncOpsIndex(int asyncOpsIndex) { _asyncOpsIndex = asyncOpsIndex; };
 
+    void* getNativeHandle() override { return &signal; }
+
 protected:
     uint64_t   apiStartTick;
     HSAOpCoord _opCoord;
     int        _asyncOpsIndex;
+
+    hsa_signal_t signal;
+    int signalIndex;
 };
 std::ostream& operator<<(std::ostream& os, const HSAOp & op);
 
 
 class HSACopy : public HSAOp {
 private:
-    hsa_signal_t signal;
-    int signalIndex;
     bool isSubmitted;
     bool isAsync;          // copy was performed asynchronously
     bool isSingleStepCopy;; // copy was performed on fast-path via a single call to the HSA copy routine
@@ -735,7 +738,6 @@ public:
     std::shared_future<void>* getFuture() override { return future; }
     const Kalmar::HSADevice* getCopyDevice() { return copyDevice; } ;  // Which device did the copy.
 
-    void* getNativeHandle() override { return &signal; }
 
     void setWaitMode(Kalmar::hcWaitMode mode) override {
         switch (mode) {
@@ -835,8 +837,6 @@ private:
 
 class HSABarrier : public HSAOp {
 private:
-    hsa_signal_t signal;
-    int signalIndex;
     bool isDispatched;
     hsa_wait_state_t waitMode;
 
@@ -869,7 +869,6 @@ public:
     bool barrierNextSyncNeedsSysRelease() const { return _barrierNextSyncNeedsSysRelease; };
     bool barrierNextKernelNeedsSysAcquire() const { return _barrierNextKernelNeedsSysAcquire; };
 
-    void* getNativeHandle() override { return &signal; }
 
     void setWaitMode(Kalmar::hcWaitMode mode) override {
         switch (mode) {
@@ -908,8 +907,6 @@ public:
         } else {
             depCount = 0;
         }
-        signal.handle = 0;
-        signalIndex = -1;
     }
 
     // constructor with at most 5 prior dependencies
@@ -936,8 +933,6 @@ public:
             // throw an exception
             throw Kalmar::runtime_exception("Incorrect number of dependent signals passed to HSABarrier constructor", count);
         }
-        signal.handle = 0;
-        signalIndex = -1;
     }
 
     ~HSABarrier() {
@@ -985,8 +980,6 @@ private:
     int kernargMemoryIndex;
 
 
-    hsa_signal_t signal;
-    int signalIndex;
     hsa_kernel_dispatch_packet_t aql;
     bool isDispatched;
     hsa_wait_state_t waitMode;
@@ -1002,7 +995,6 @@ public:
     const char *getKernelName() { return kernel_name ? kernel_name : (kernel ? kernel->shortKernelName.c_str() : "<unknown_kernel>"); };
     const char *getLongKernelName() { return (kernel ? kernel->getLongKernelName().c_str() : "<unknown_kernel>"); };
 
-    void* getNativeHandle() override { return &signal; }
 
     void setWaitMode(Kalmar::hcWaitMode mode) override {
         switch (mode) {
@@ -1478,7 +1470,7 @@ public:
             if (asyncOp != nullptr) {
                 hsa_signal_t signal = *(static_cast <hsa_signal_t*> (asyncOp->getNativeHandle()));
                 if (signal.handle) {
-                    hsa_signal_value_t v = hsa_signal_load_relaxed(signal);
+                    hsa_signal_value_t v = hsa_signal_load_acquire(signal);
                     if (v != 0) {
                         ++count;
                     }
@@ -1494,31 +1486,29 @@ public:
     bool isEmpty() override {
         // Have to walk asyncOps since it can contain null pointers (if event is waited on and removed)
         // Also not all commands contain signals.
-        bool oldestNoSignal=false;
-        for (int i = asyncOps.size()-1; i >= 0;  i--) {
-            auto &asyncOp = asyncOps[i];
-            if (asyncOp != nullptr) {
-                hsa_signal_t signal = *(static_cast <hsa_signal_t*> (asyncOp->getNativeHandle()));
-                if (signal.handle) {
-                    oldestNoSignal=false;
-                    hsa_signal_value_t v = hsa_signal_load_relaxed(signal);
-                    if (v != 0) {
-                        return false;
-                    }
-                } else {
-                    // no signal, have to assume the command is still running
-                    oldestNoSignal=true;
-                }
-            }
-        };
+        
+        bool isEmpty = true;
 
-        if (oldestNoSignal) {
-            auto marker = EnqueueMarker(hc::system_scope);
-            DBOUTL(DB_CMD2, "Inside HSAQueue::isEmpty and queue contained only no-signal ops, enqueued marker " << marker << " into " << *this);
-            return false;
-        } else {
-            return true;
+        const auto& oldest = find_if(
+                    asyncOps.crbegin(), asyncOps.crend(), [](const std::shared_ptr<HSAOp> &asyncOp) { return asyncOp != nullptr; });
+
+
+        if (oldest != asyncOps.crend()) {
+            hsa_signal_t signal = *(static_cast <hsa_signal_t*> ((*oldest)->getNativeHandle()));
+            if (signal.handle) {
+                hsa_signal_value_t v = hsa_signal_load_acquire(signal);
+                if (v != 0) {
+                    isEmpty=false;
+                }
+            } else {
+                // oldest has no signal - enqueue a new one:
+                auto marker = EnqueueMarker(hc::system_scope);
+                DBOUTL(DB_CMD2, "Inside HSAQueue::isEmpty and queue contained only no-signal ops, enqueued marker " << marker << " into " << *this);
+                isEmpty=false;
+            }
         }
+
+        return isEmpty;
     };
 
 
@@ -4222,8 +4212,6 @@ HSADispatch::HSADispatch(Kalmar::HSADevice* _device, Kalmar::KalmarQueue *queue,
     future(nullptr),
     kernargMemory(nullptr)
 {
-    signal.handle = 0;
-    signalIndex = -1;
     if (aql) {
         this->aql = *aql;
     }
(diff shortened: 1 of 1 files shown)


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1494,22 +1494,31 @@ class HSAQueue final : public KalmarQueue
     bool isEmpty() override {
         // Have to walk asyncOps since it can contain null pointers (if event is waited on and removed)
         // Also not all commands contain signals.
-        for (int i = 0; i < asyncOps.size(); ++i) {
-            if (asyncOps[i] != nullptr) {
-                auto &asyncOp = asyncOps[i];
+        bool oldestNoSignal=false;
+        for (int i = asyncOps.size()-1; i >= 0;  i--) {
```
AlexVlx: @bensander couldn't this be just be find_if with reverse iterators + the associated processing? Which is to say something along the lines of: 
```cpp
const auto oldest = find_if(
    asyncOps.crbegin(), asyncOps.crend(), [](... asyncOp) { return asyncOp != nullptr; });

if (oldest != asyncOps.crend()) { /* do the check for signal */ }
```
The loop is somewhat unnatural as is (it appears more customary to use while loops when counting down).

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1494,22 +1494,31 @@ class HSAQueue final : public KalmarQueue
...
+            auto &asyncOp = asyncOps[i];
+            if (asyncOp != nullptr) {
                 hsa_signal_t signal = *(static_cast <hsa_signal_t*> (asyncOp->getNativeHandle()));
                 if (signal.handle) {
+                    oldestNoSignal=false;
                     hsa_signal_value_t v = hsa_signal_load_relaxed(signal);
```
AlexVlx: Are we sure relaxed with no fence is what we want here? I believe that this is concurrently written by the CP, is it not?

AlexVlx (changes requested):
In general LGTM, but could you please peek at the 2 items I mentioned? Thank you @bensander.

lib/hsa/mcwamp_hsa.cpp
bensander: This is unchanged from code before, but probably should be changed.  It comes down to what effect is implied on the memory model if a user submits a command and then checks if the associated accelerator view is empty - probably would assume that the results of that command are visible.

lib/hsa/mcwamp_hsa.cpp
bensander: yep, I think this can be refactored as you suggest, will submit a new one.  It started as a forward loop ; now that it runs backward all we really have to do is find the oldest op and see if it has a signal or not.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4970,14 +4969,14 @@ HSACopy::waitComplete() {
     // Wait on completion signal until the async copy is finishedS
     if (DBFLAG(DB_WAIT)) {
         hsa_signal_value_t v = -1000;
-        if (signal.handle) {
-            hsa_signal_load_acquire(signal);
+        if (_signal.handle) {
+            hsa_signal_load_acquire(_signal);
```
AlexVlx: Unless I am mistaken, this is deprecated, with `hsa_signal_load_scacquire` being the current / preferred form.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4970,14 +4969,14 @@ HSACopy::waitComplete() {
...
         }
-        DBOUT(DB_WAIT, "  wait for copy op#" << getSeqNum() << " completion with wait flag: " << waitMode << "signal="<< std::hex  << signal.handle << std::dec <<" currentVal=" << v << "...\n");
+        DBOUT(DB_WAIT, "  wait for copy op#" << getSeqNum() << " completion with wait flag: " << waitMode << "signal="<< std::hex  << _signal.handle << std::dec <<" currentVal=" << v << "...\n");
     }
 
     // Wait on completion signal until the async copy is finished
-    hsa_signal_wait_acquire(signal, HSA_SIGNAL_CONDITION_LT, 1, UINT64_MAX, waitMode);
+    hsa_signal_wait_acquire(_signal, HSA_SIGNAL_CONDITION_LT, 1, UINT64_MAX, waitMode);
```
AlexVlx, line 4979: Unless I am mistaken, this is deprecated, with `hsa_signal_wait_scacquire` being the current / preferred form.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4847,7 +4844,7 @@ static std::string fenceToString(int fenceBits)
 
 
 bool HSABarrier::isReady() override {
-    bool ready = (hsa_signal_load_acquire(signal) == 0);
+    bool ready = (hsa_signal_load_acquire(_signal) == 0);
```
AlexVlx: Unless I am mistaken, this is deprecated, with `hsa_signal_load_scacquire` being the current / preferred form.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4856,7 +4853,7 @@ bool HSABarrier::isReady() override {
 }
 
 bool HSACopy::isReady() override {
-    bool ready = (hsa_signal_load_acquire(signal) == 0);
+    bool ready = (hsa_signal_load_acquire(_signal) == 0);
```
AlexVlx: Unless I am mistaken, this is deprecated, with `hsa_signal_load_scacquire` being the current / preferred form.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4699,10 +4696,10 @@ HSABarrier::waitComplete() {
         return HSA_STATUS_ERROR_INVALID_ARGUMENT;
     }
 
-    DBOUT(DB_WAIT,  "  wait for barrier " << *this << " completion with wait flag: " << waitMode << "  signal="<< std::hex  << signal.handle << std::dec <<"...\n");
+    DBOUT(DB_WAIT,  "  wait for barrier " << *this << " completion with wait flag: " << waitMode << "  signal="<< std::hex  << _signal.handle << std::dec <<"...\n");
 
     // Wait on completion signal until the barrier is finished
-    hsa_signal_wait_acquire(signal, HSA_SIGNAL_CONDITION_EQ, 0, UINT64_MAX, waitMode);
+    hsa_signal_wait_acquire(_signal, HSA_SIGNAL_CONDITION_EQ, 0, UINT64_MAX, waitMode);
```
AlexVlx, line 4702: Unless I am mistaken, this is deprecated, with `hsa_signal_load_scacquire` being the current / preferred form.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4865,7 +4862,7 @@ bool HSACopy::isReady() override {
 }
 
 bool HSADispatch::isReady() override {
-    bool ready = (hsa_signal_load_acquire(signal) == 0);
+    bool ready = (hsa_signal_load_acquire(_signal) == 0);
```
AlexVlx: Unless I am mistaken, this is deprecated, with `hsa_signal_load_scacquire` being the current / preferred form.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -5388,11 +5387,11 @@ HSACopy::syncCopyExt(hc::hcCommandKind copyDir, const hc::AmPointerInfo &srcPtrI
         }
 
 
-        hsa_status_t hsa_status = hcc_memory_async_copy(copyDir, copyDevice, dstPtrInfo, srcPtrInfo, sizeBytes, depSignalCnt, depSignalCnt ? &depSignal:NULL, signal);
+        hsa_status_t hsa_status = hcc_memory_async_copy(copyDir, copyDevice, dstPtrInfo, srcPtrInfo, sizeBytes, depSignalCnt, depSignalCnt ? &depSignal:NULL, _signal);
 
         if (hsa_status == HSA_STATUS_SUCCESS) {
             DBOUT(DB_COPY, "HSACopy::syncCopyExt(), wait for completion...");
-            hsa_signal_wait_relaxed(signal, HSA_SIGNAL_CONDITION_LT, 1, UINT64_MAX, waitMode);
+            hsa_signal_wait_relaxed(_signal, HSA_SIGNAL_CONDITION_LT, 1, UINT64_MAX, waitMode);
```
AlexVlx, line 5394: Should we put this in a loop, since it's both a relaxed wait and it is legal for a signal to spuriously return?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -5376,10 +5375,10 @@ HSACopy::syncCopyExt(hc::hcCommandKind copyDir, const hc::AmPointerInfo &srcPtrI
 
         // Get a signal and initialize it:
         std::pair<hsa_signal_t, int> ret = Kalmar::ctx.getSignal();
-        signal = ret.first;
-        signalIndex = ret.second;
+        _signal = ret.first;
+        _signalIndex = ret.second;
 
-        hsa_signal_store_relaxed(signal, 1);
+        hsa_signal_store_relaxed(_signal, 1);
```
AlexVlx, line 5381: Should either consider using screlease or a release fence here, to ensure no reordering / fully observable precursors as far as the CP is concerned?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4434,11 +4431,11 @@ HSADispatch::waitComplete() {
...
 
 
-    if (signal.handle) {
-        DBOUT(DB_MISC, "wait for kernel dispatch op#" << *this  << " completion with wait flag: " << waitMode << "  signal="<< std::hex  << signal.handle << std::dec << "\n");
+    if (_signal.handle) {
+        DBOUT(DB_MISC, "wait for kernel dispatch op#" << *this  << " completion with wait flag: " << waitMode << "  signal="<< std::hex  << _signal.handle << std::dec << "\n");
 
         // wait for completion
-        if (hsa_signal_wait_acquire(signal, HSA_SIGNAL_CONDITION_LT, 1, uint64_t(-1), waitMode)!=0) {
+        if (hsa_signal_wait_acquire(_signal, HSA_SIGNAL_CONDITION_LT, 1, uint64_t(-1), waitMode)!=0) {
```
AlexVlx, line 4438: Unless I am mistaken, this is deprecated, with `hsa_signal_load_scacquire` being the current / preferred form.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1494,22 +1486,29 @@ class HSAQueue final : public KalmarQueue
...
+        bool isEmpty = true;
+
+        const auto& oldest = find_if(
+                    asyncOps.crbegin(), asyncOps.crend(), [](const std::shared_ptr<HSAOp> &asyncOp) { return asyncOp != nullptr; });
+
+
+        if (oldest != asyncOps.crend()) {
+            hsa_signal_t signal = *(static_cast <hsa_signal_t*> ((*oldest)->getNativeHandle()));
+            if (signal.handle) {
+                hsa_signal_value_t v = hsa_signal_load_acquire(signal);
```
AlexVlx: Unless I am mistaken, this is deprecated, with `hsa_signal_load_scacquire` being the current / preferred form.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1478,7 +1470,7 @@ class HSAQueue final : public KalmarQueue
             if (asyncOp != nullptr) {
                 hsa_signal_t signal = *(static_cast <hsa_signal_t*> (asyncOp->getNativeHandle()));
                 if (signal.handle) {
-                    hsa_signal_value_t v = hsa_signal_load_relaxed(signal);
+                    hsa_signal_value_t v = hsa_signal_load_acquire(signal);
```
AlexVlx: Unless I am mistaken, this is deprecated, with `hsa_signal_load_scacquire` being the current / preferred form.

AlexVlx (commented):
Thank you for the changes, LGTM. Now, since you went beyond the call of duty, and touched a few other bits, there are a few other (possibly pedantic :) ) concerns, but it is your call if you want to address them or not.

lib/hsa/mcwamp_hsa.cpp
bensander: Most of these are already present in the code and unrelated to the new functionality, lets treat as a new PR.

lib/hsa/mcwamp_hsa.cpp
bensander: These are resetting signal value so relaxed seems fine.     CP will store-release to set to 0 ; any consumers wait for the 0 value so that is important point to release.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1494,22 +1486,29 @@ class HSAQueue final : public KalmarQueue
...
+        if (oldest != asyncOps.crend()) {
+            hsa_signal_t signal = *(static_cast <hsa_signal_t*> ((*oldest)->getNativeHandle()));
+            if (signal.handle) {
+                hsa_signal_value_t v = hsa_signal_load_scacquire(signal);
+                if (v != 0) {
+                    isEmpty=false;
                 }
+            } else {
+                // oldest has no signal - enqueue a new one:
+                auto marker = EnqueueMarker(hc::system_scope);
```
scchan, line 1505: what's the reason for enqueueing a marker?

AlexVlx (approved):

lib/hsa/mcwamp_hsa.cpp
bensander: If last command had no completion signal then we have no way to tell if he queue is empty or not.  So we enque a marker here.  If app calls is empty again the marker will be oldest and we can get a valid read on the status.

PR #561, opened 2017-11-21
in dispatch_hsa_kernel, detect deps before adding kernel to asyncOps

bensander:
- This is required since detectStreamDeps compares new op against
  oldest op in the asyncOps, and may also need to add dependencies
  BEFORE the new op.
- Also add additional debug to detectStreamDeps.
-


scchan (approved):
LGTM

PR #566, opened 2017-11-28
Allow markers to be dependent on any command (not just markers)

bensander:
if last command was copy we can make the marker depend hcc
will make the marker wait on the signal written by the copy.
So the HSABarrier::depAsyncOps[] now contains HSAOps not
just HSABarrier.  Refactor some other code to support this.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -693,6 +693,9 @@ public:
 
     void* getNativeHandle() override { return &_signal; }
 
+    virtual bool barrierNextSyncNeedsSysRelease() const { return 0; };
+    virtual bool barrierNextKernelNeedsSysAcquire() const { return 0; };
+
 protected:
     uint64_t     apiStartTick;
     HSAOpCoord   _opCoord;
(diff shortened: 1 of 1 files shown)


bensander:
jack, alex - can you give this a once-over?  We need it for PR246 in HIP land.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -693,6 +693,9 @@ class HSAOp : public Kalmar::KalmarAsyncOp {
 
     void* getNativeHandle() override { return &_signal; }
 
+    virtual bool barrierNextSyncNeedsSysRelease() const { return 0; };
+    virtual bool barrierNextKernelNeedsSysAcquire() const { return 0; };
```
whchung, line 697: @bensander since these 2 are now virtual methods in `HSAOp`, perhaps ones in `HSABarrier` shall be changed to carry `override`. the reset of the PR LGTM.

whchung (changes requested):

lib/hsa/mcwamp_hsa.cpp
bensander: good point, I will update

whchung (approved):

PR #567, opened 2017-11-29
fix rpm dependency list

scchan:
remove libc++ from the dep list if it's not needed


aaronenyeshi:
Hi SiuChi, I believe that `libc++` and `libc++abi` is not needed for the fedora build. But I believe it is needed on CentOS 7 and RHEL environments. So those builds would probably need to enable `USE_LIBCXX` cmake option.

scchan:
It will and it would be enabled based on the gcc version check

aaronenyeshi (approved):

PR #568, opened 2017-11-30
Vgrps limit

smithakihide:
hello. 
I am japanese. so if there was any impolite words or poor English, please excuse me.
this pull request is very important. please consider.
 
I changed the fixing way of workgroup_size.

On the original code, we cannot run a kernel which use many VGPRs per work item.
because, although the original code have considered fixing workitem_size so that it will be in the limit number of workitem_max_dim, this have never made workgroup_size small.
As a result, it have made workitem_total_size 1024 in most cases.
For example, when workitem_total_size was set to 1024, we cannot use more than 64 VGPRs despite of the actual maximum VGPRs number of 256 on GCN architecture.

Most application such like Deep Leaning, FFT or Matrix operations would not need so many VGPRs. I supposed this is why no one mentioned about this problem. but in the Computational Fluid Dynamics, its code will be very complex and need many VGPRs. to run such a code, we have to reduce workgroup_size rather than VGPRs.

I would be the first person who run the Computational Fluid Dynamic code on ROCm.
best regards.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -4681,7 +4681,7 @@ HSADispatch::setLaunchConfiguration(int dims, size_t *globalDims, size_t *localD
 		}
 		workgroup_total_size = workgroup_size[0] * workgroup_size[1] * workgroup_size[2];
 	}
-	std::cout << "fixed workgroup_size = " << workgroup_size[0] << " " << workgroup_size[1] <<" " << workgroup_size[2] << std::endl;
+	std::cout << "cfixed workgroup_size = " << workgroup_size[0] << " " << workgroup_size[1] <<" " << workgroup_size[2] << std::endl;
 
 
     aql.workgroup_size_x = workgroup_size[0];
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -4658,7 +4658,7 @@ HSADispatch::setLaunchConfiguration(int dims, size_t *globalDims, size_t *localD
 	if (max_num_vgprs_per_work_item < workitem_vgpr_count)
 	{
 		std::string msg = "The number of VGPRs (" + std::to_string(kernel->workitem_vgpr_count) + ") needed by this launch (" +
-			std::to_string(kernel->getKernelName()) + ") exceeds HW (" +
+			kernel->getKernelName() + ") exceeds HW (" +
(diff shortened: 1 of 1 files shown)


scchan:
@smithakihide It seems that some of the logic in HSADispatch::setLaunchConfiguration is incorrect; I'll take a deeper look at it.
Meanwhile, to address the problem you are encountering, we have a function attribute called `hc_max_workgroup_dim(x,y,z)` for users to communicate to the compiler what would be the maximum workgroup size and the compiler would try to allocate registers to allow that.

For example, if you intent to launch a kernel with a 1-D workgroup of size 1024, then you could write

```
parallel_for_each(t_extent, [=](tiled_index<1> tidx) [[hc]] 
                                                     [[hc_max_workgroup_dim(1024,1,1)]] {
...
});
```
Note that a large workgroup size would reduce the number of registers available per workitem and therefore, has a negative impact on performance.  
Please give this a try and let us know.

smithakihide:
I am sorry to have kept you waiting. I had created the simple test code to make the problem. 

thank you for answering, and as I also supposed, when I designate the proper tile index it worked.
this is the test code. the 3D forth precision discretized diffusion equation.
[https://github.com/smithakihide/amp_eq_test.git](https://github.com/smithakihide/amp_eq_test.git)
to do test, please clone this repository and execute make command in the directory of sources.
the commands would be like below.

```
git clone https://github.com/smithakihide/amp_eq_test.git
cd amp_eq_test/sources
make
./exe
```

and you can change the work group size by setting TDS definition as below.

```
make  FLAGS="-DTDS=4"
```

note that the default value of TDS is 8. And each work group dimension size of wx,wy,wz will be calculated as below.

`wx = 2*TDS, wy = TDS, wz = TDS`
so if we set TDS = 8, it makes workgroup size the maximum limit.


and when I use the HCC forked by me ([https://github.com/smithakihide/hcc.git](https://github.com/smithakihide/hcc.git)), this code would be work without tile index.

smithakihide:
sorry, I added the two functions and comment output before execute each functions.
```
void Acc_fp64_tiled();
void Acc_fp64_normal();
```
former one is tiled version. and the other is normal version. the current HCC is unable to manage such like this normal version's kernel which will use many VGPRs more than 64.

smithakihide:
I found that my implementation is not correct. so please find another proper way to fix this problem.
we have to throw error when the compiler is given tile-indexed kernel and it will exceed the maximum number of vgprs. we cannot change the workgroup dimension size in such this case.

scchan:
closing this PR since it's not working properly

PR #570, opened 2017-12-01
Update docker file for auto-upstream-merge.  Adding sudoer id "jenkins"

david-salinas:
Added some lines to the bottom of this docker file that is being used by the auto-upstream-merge tool, to create and add a sudoer "jenkins".  This user id will exist/operate only within the Docker container.


whchung (approved):
LGTM

PR #571, opened 2017-12-01
Clear op->queue field when op is removed from AsyncOps.

bensander:
Avoid stale pointer issue if queue is later deleted before the op.
Also fix the profiler so it uses opCoord directly rather than relying
on queue pointer.
Refactor to move isReady into HSAOp, eliminating some duplicated code


scchan (approved):
LGTM

PR #572, opened 2017-12-01
insert to a non-const iterator

scchan:
the range string::insert takes an iterator rather than a const_iterator as 1st argument

### DIFF
--- a/hc2/headers/types/code_object_bundle.hpp
+++ b/hc2/headers/types/code_object_bundle.hpp
@@ -69,7 +69,7 @@ namespace hc2
                     std::copy_n(it, sizeof(y.cbuf), y.cbuf);
                     it += sizeof(y.cbuf);
 
-                    y.triple.insert(y.triple.cend(), it, it + y.triple_sz);
+                    y.triple.insert(y.triple.end(), it, it + y.triple_sz);
 
                     std::copy_n(
                         f + y.offset, y.bundle_sz, std::back_inserter(y.blob));


pfultz2:
C++11 allows inserting with a const_iterator.

scchan:
with the signature used in the code?

pfultz2:
Yes, see overload 8, [here](http://en.cppreference.com/w/cpp/string/basic_string/insert). In C++11, it was changed to use `const_iterator`. Is this needing to compile with a old C++ standard library?

scchan:
@pfultz2 You are right and yes this is needed for an older libstdc++ RHEL.  I was using g++ 6.3 from RHEL's Developer Toolset 6 and thinking that it was using a newer libstdc++.  I was confused by the fact it's still using the old libstdc++ from g++ 4.8

hc2/headers/types/code_object_bundle.hpp
```diff
@@ -69,7 +69,7 @@ namespace hc2
                     std::copy_n(it, sizeof(y.cbuf), y.cbuf);
                     it += sizeof(y.cbuf);
 
-                    y.triple.insert(y.triple.cend(), it, it + y.triple_sz);
+                    y.triple.insert(y.triple.end(), it, it + y.triple_sz);
```
AlexVlx, line 72: @scchan if we really want to work around such things on our end, it is probably preferable to change the call to `y.triple.assign(it, it + y.triple_sz)`. For reference, the interface of assign is as follows: http://en.cppreference.com/w/cpp/container/vector/assign.

scchan: but the insert is on `triple`, which is a `string` not a `vector`.  I can construct a new `string` and assign to triple or `append` `it` to `triple`

AlexVlx: That is OK @scchan, containers (string is confusingly sort of a container) have the same computational basis. See: http://en.cppreference.com/w/cpp/string/basic_string/assign.

scchan: thanks, will submit a fix through a different branch/PR

PR #573, opened 2017-12-04
workaround bash issue on RHEL with declare -n

scchan:
bash on RHEL doesn't seem to support "declare -n" (generated an error), implemented a workaround for that


aaronenyeshi (approved):
Yes this workaround should work. Seems that declare -n which uses namerefs was introduced in Bash 4.3 +. Your commit should support older versions since bash supports dynamic scoping, thus we can access variables in last scope. Thanks SiuChi!

PR #581, opened 2017-12-08
add precheckin dependencies to auto-merge docker file

david-salinas:
I've added install dependencies for the precheckin tests (rocBLAS, rocFFT, rocRAND) to the docker file for the auto-upstream-merge tool.  This should allow us to avoid using "./install.sh -d" and just use "install.sh -ic".   We want to ensure that the tool, when it runs the precheckin tests, that the tests are using the "development" HCC/HIP it just built from the result of the upstream merge, and not the default installed hcc/hip


whchung (approved):
LGTM

PR #582, opened 2017-12-11
Fix workgroup size handling logic in hcc RT

scchan:
The runtime currently has some logic to reduce workgroup sizes specified by a user if they exceed the device or the register limit.  This doesn't seem correct since it changes the local IDs expected by the user.  This PR will remove this logic and make the runtime to through an exception instead.

This PR also changed how group sizes are selected if they are not specified by the user.

Additional tests are created to test the new behavior.


smithakihide:
May I help you to test with my code?

scchan:
@smithakihide yes, please give this a try.  Also, have you tried with the hc_max_workgroup_dim attribute I suggested in https://github.com/RadeonOpenCompute/hcc/pull/568

scchan:
@whchung could you take a look at this when you get a chance?  Thanks.

smithakihide:
yes, I will try. And thank you for trying fix this problem.

I haven't tried 
[https://github.com/RadeonOpenCompute/hcc/pull/568](https://github.com/RadeonOpenCompute/hcc/pull/568) because I have used C ++AMP. but now I will try to use hc_max_workgroup_dim with hcc mode. please wait a few days.

Although I tested fix tile_extent with C++ amp as I mentioned in [https://github.com/RadeonOpenCompute/hcc/pull/568](https://github.com/RadeonOpenCompute/hcc/pull/568).

smithakihide:
I tested your branch of fix_workgroup_size with [https://github.com/smithakihide/amp_eq_test](https://github.com/smithakihide/amp_eq_test) and it worked well.

here is the commands to test this my repository.
```
git clone https://github.com/smithakihide/amp_eq_test
cd amp_eq_test/sources
make FLAGS="-DTDS=4"
./exe
```

note that this code dosen't work on current hcc.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4587,10 +4577,124 @@ void HSADispatch::overrideAcquireFenceIfNeeded()
...
+        }
+      }
+    }
+
+    auto kernel = this->kernel;
+
+    auto calculate_kernel_max_flat_workgroup_size = [&] {
+      constexpr unsigned int max_num_vgprs_per_work_item = 256;
+      constexpr unsigned int num_work_items_per_simd = 64;
+      constexpr unsigned int num_simds_per_cu = 4;
```
aaronenyeshi, line 4669: Are there any constant global variables that hold these constants? I am wondering if other functions also need to use these values, they can't reference this lambda here.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4600,57 +4704,11 @@ HSADispatch::setLaunchConfiguration(int dims, size_t *globalDims, size_t *localD
     aql.group_segment_size   = kernel->static_group_segment_size + dynamicGroupSize;;
```
aaronenyeshi: extra semi colon, also would it be useful to print `kernel->private_segment_size` in debugging mode too?

aaronenyeshi: extra semi colon, also would it be useful to print `kernel->private_segment_size` in debugging mode too?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4587,10 +4577,124 @@ void HSADispatch::overrideAcquireFenceIfNeeded()
...
+
+    auto calculate_kernel_max_flat_workgroup_size = [&] {
+      constexpr unsigned int max_num_vgprs_per_work_item = 256;
+      constexpr unsigned int num_work_items_per_simd = 64;
+      constexpr unsigned int num_simds_per_cu = 4;
+      const unsigned int workitem_vgpr_count = std::max((unsigned int)kernel->workitem_vgpr_count, 1u);
+      unsigned int max_flat_group_size = (max_num_vgprs_per_work_item / workitem_vgpr_count) 
+                                           * num_work_items_per_simd * num_simds_per_cu;
+      // round it to the closest wave size
+      max_flat_group_size -= max_flat_group_size%num_work_items_per_simd;
```
aaronenyeshi: I think since we multiplied by `num_work_items_per_simd`, then `max_flat_group_size` should already be a multiple of it. So we might not need to round to closest wave, unless this `(max_num_vgprs_per_work_item / workitem_vgpr_count)` produces a fraction, but we could guard it with a cast to unsigned int.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4587,10 +4577,124 @@ void HSADispatch::overrideAcquireFenceIfNeeded()
...
+      if (dims == 1) {
+        workgroup_size[0] = recommended_flat_workgroup_size;
+      }
+      else if (dims == 2) {
+
+        // compute the group size for the 1st dimension
+        for (unsigned int i = 1; ; i<<=1) {
+          if (i == recommended_flat_workgroup_size
+              || i >= globalDims[0]) {
+            workgroup_size[0] = i;
```
aaronenyeshi: Should we be choosing the minimum between `i`, `globalDims[0]`, and `recommended_flat_workgroup_size` here?

aaronenyeshi (changes requested):

lib/hsa/mcwamp_hsa.cpp
scchan: removed

lib/hsa/mcwamp_hsa.cpp
scchan: I don't see a need in this PR but it could be refactored in the future

lib/hsa/mcwamp_hsa.cpp
scchan: yes, we could take a min between i and globalDims[0] which would lead to a more finer grain group size, I also updated the logic for other dimensions due to this change.

lib/hsa/mcwamp_hsa.cpp
scchan: makes sense, the rounding was removed in the newer version

aaronenyeshi (approved):

aaronenyeshi:
LGTM, but should wait for jenkins

PR #585, opened 2017-12-13
Update rpt to support "older" traces without enqueue time

parallelo:


bensander (approved):
LGTM, seems HCC does not yet write the enq time so we still need this code path in RPT.

PR #586, opened 2017-12-15
added more dependencies for precheckin tests (hiptensorflow)

david-salinas:
Just for the auto-upstream-merge tool.

Added some more packages to be installed, needed for hiptensorflow, as well as some environment arguments.


whchung (approved):
LGTM

PR #594, opened 2018-01-04
Make amp flag private

pfultz2:
This is so users can set their own C++ version.

### DIFF
--- a/scripts/cmake/MCWAMP.cmake
+++ b/scripts/cmake/MCWAMP.cmake
@@ -39,7 +39,6 @@ macro(amp_target name )
   target_include_directories(${name} SYSTEM INTERFACE $<INSTALL_INTERFACE:$<INSTALL_PREFIX>/include>)
   add_libcxx_option_if_needed(${name})
   target_compile_options(${name} PUBLIC -hc -fPIC)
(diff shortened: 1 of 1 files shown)


scripts/cmake/MCWAMP.cmake
```diff
@@ -38,7 +38,8 @@ macro(amp_target name )
   target_include_directories(${name} PRIVATE ${MCWAMP_INC_DIR})
   target_include_directories(${name} SYSTEM INTERFACE $<INSTALL_INTERFACE:$<INSTALL_PREFIX>/include>)
   add_libcxx_option_if_needed(${name})
-  target_compile_options(${name} PUBLIC -hc -std=c++amp -fPIC)
+  target_compile_options(${name} PUBLIC -hc -fPIC)
+  target_compile_options(${name} PRIVATE -std=c++amp)
```
scchan: I believe the c++amp is not needed anymore so I'd suggest that we just remove it.

scchan (changes requested):

scripts/cmake/MCWAMP.cmake
pfultz2: Got it, I removed it.

scchan (approved):

PR #595, opened 2018-01-04
Add ubsan test

pfultz2:
This adds a test for using ubsan with hcc. It requires [this](https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/124) PR to be merged first.


pfultz2:
It looks like jenkins ignores when tests fail, which is not good.

pfultz2:
@scchan I think this PR can be safely merged in now.

scchan (approved):

PR #596, opened 2018-01-08
[Jenkins] check the unit test return code

scchan:


pfultz2:
Are we still not checking the return code? It seems like this should be merged.

scchan:
@pfultz2 could you review and approve?

pfultz2 (approved):

PR #605, opened 2018-01-17
[Config] revise submodule configuration from upstream merge

rocm-hcc:
Upstream merge with correction for Lamdba expressions in Clang.


rocm-hcc:
There is a Merge conflict with this PR.  But as far as I can tell, it is in only in CMakeLists.txt.  I made a change to the clang path, from 6.0.0 to 7.0.0 because LLVM, with the latest upstream merge, has changed its reference to clang headers to 7.0.0.  I don't have Write authority to this repo.  But could someone who does take a look at resolve this merge conflict?

scchan:
@david-salinas could you take a look at the conflict with `compiler`? thanks

PR #608, opened 2018-01-19
workaround build issue on RHEL

scchan:
This is to workaround the issue that the new std::string implementation not being available on RHEL since their g++ is not using the C++11 ABI

### DIFF
--- a/hc2/headers/types/code_object_bundle.hpp
+++ b/hc2/headers/types/code_object_bundle.hpp
@@ -69,7 +69,7 @@ namespace hc2
                     std::copy_n(it, sizeof(y.cbuf), y.cbuf);
                     it += sizeof(y.cbuf);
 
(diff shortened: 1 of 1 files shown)


AlexVlx:
@whchung @scchan guys, I am still a bit confused about what's wrong with assign, sorry:(. The interface taking a pair of iterators has been there since pre C++11. Separately, not enabling the C++11 ABI, while somewhat orthogonal to this particular item, sounds risky - perhaps we can discuss further? Thanks!

scchan:
@AlexVlx If I understand correctly, you want it changed to `assign` because `triple` is always an empty string that point?

scchan:
discussed @AlexVlx and we think this change is acceptable regardless of the C++11 ABI issue on RHEL

PR #615, opened 2018-01-23
[CMAKE] retrieve the clang version from clang's cmake

scchan:
get clang's version from clang's cmake instead of hardcoding it in hcc's cmake, this avoids breakage due to clang version updates


whchung (approved):
LGTM. Thanks for making this change to get rid of one more hard-coded number.

aaronenyeshi:
I'm running into an issue on Ubuntu using a clean cmake:
```
CMake Error at CMakeLists.txt:333 (get_directory_property):
  get_directory_property DIRECTORY argument provided but requested directory
  not found.  This could be because the directory argument was invalid or, it
  is valid but has not been processed yet.
```

aaronenyeshi:
I have a fix

PR #617, opened 2018-01-25
Fix cmake clang version for clean builds

aaronenyeshi:
Seems that it was checking the binary clang directory instead of the source clang directory. Therefore this fixes cmake error for directory not found.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -330,7 +330,7 @@ if(EXISTS ${PROJECT_SOURCE_DIR}/clang-tools-extra/CMakeLists.txt)
 endif()
 
 add_subdirectory(${CLANG_SRC_DIR})
-get_directory_property(CLANG_VERSION DIRECTORY clang DEFINITION CLANG_VERSION)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -330,7 +330,7 @@ if(EXISTS ${PROJECT_SOURCE_DIR}/clang-tools-extra/CMakeLists.txt)
 endif()
 
 add_subdirectory(${CLANG_SRC_DIR})
-get_directory_property(CLANG_VERSION DIRECTORY clang DEFINITION CLANG_VERSION)
+get_directory_property(CLANG_VERSION DIRECTORY ${CLANG_SRC_DIR} DEFINITION CLANG_VERSION)
```
scchan, line 333: `${CLANG_SRC_DIR}` is what I put initially and it doesn't work on my machine.  
Using just `clang` works for me and on our build machines for clean builds so I'm unable to reproduce the failure observe.

aaronenyeshi:
You're correct, this is wrong

CMakeLists.txt
aaronenyeshi: Hmm I'm not sure why, but a clean build doesn't work for me. So I manually set it to 7.0.0 for now and that works.

PR #622, opened 2018-01-29
don't always assign the same host agent to the GPUs

scchan:


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -3383,16 +3368,13 @@ class HSAContext final : public KalmarContext
...
-        // Iterate over agents to find out the first cpu device as host
-        status = hsa_iterate_agents(&HSAContext::find_host, &host);
+        std::vector<hsa_agent_t>* agent_vectors[] = {&agents, &cpu_agents};
+        status = hsa_iterate_agents(&HSAContext::find_agents, &agent_vectors);
         STATUS_CHECK(status, __LINE__);
 
         for (int i = 0; i < agents.size(); ++i) {
             hsa_agent_t agent = agents[i];
-            auto Dev = new HSADevice(agent, host, i);
+            auto Dev = new HSADevice(agent, cpu_agents[i%cpu_agents.size()], i);
```
whchung: @scchan this doesn't seem to be correct, is it possible to understand the affinity of a GPU agent to a CPU?

whchung (changes requested):

lib/hsa/mcwamp_hsa.cpp
scchan: I went through the HSA headers and it doesn't seem to provide any affinity info?
There's a node # but on my system the CPUs and GPU are all on different nodes

scchan:
@whchung I made the modification to choose a host based on min NUMA distance, could you please review this again?

whchung:
@scchan has the asymmetric NUMA mapping matrix on your test system been resolved?

scchan:
@whchung not yet, trying to update to a newer version of the stack and see if that helps.  I also tested on a different machine with symmetric NUMA distances

whchung:
@scchan the codes generally LGTM. I'm wondering is there a way to create an effective test.

whchung (approved):

whchung:
@scchan i think this is good to be merged in?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -3395,23 +3380,67 @@ class HSAContext final : public KalmarContext
...
+            // pick a CPU agent with the smallest NUMA distance
+            hsa_agent_t cpu_agent = {0};
+            uint32_t cpu_numa_distance = UINT32_MAX;
+            for (int j = 0; j < cpu_agents.size(); ++j) {
+
+              hsa_amd_memory_pool_link_info_t link_info = {0};
+              status = hsa_amd_agent_memory_pool_get_info(gpu_agent, 
+                                                          cpu_mem_pools[j],
+                                                          HSA_AMD_AGENT_MEMORY_POOL_INFO_LINK_INFO,
+                                                          &link_info);
```
scchan, line 3436: We need to figure out how many hops before querying for the link info.  This code here makes an assumption that there's only one hop

scchan:
closing this PR to fix the bug that it's missing the query for the number of hops before querying for the link info

PR #623, opened 2018-01-31
[RUNTIME] fix potential undefined behavior in mutex locking / unlocking

whchung:
In the function changed in this commit, originally an RAII lock_guard is
present for a mutex. And in the later part of the function the unlock/lock
member functions of the mutex are directly invoked. Such codes is not ideal.

It was also proven certain user codes would run into soft hang situation.

In this commit we remove the lock_guard and always use lock/unlock member
function calls. The user codes affected above have been proven to run.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -2267,64 +2267,74 @@ public:
     class UnpinnedCopyEngine      *copy_engine[2]; // one for each direction.
     UnpinnedCopyEngine::CopyMode  copy_mode;
 
-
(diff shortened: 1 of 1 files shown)


whchung:
@scchan After @sunway513 verified user applications work I'm submitting this PR. I created this commit directly on 1.7.1 release branch therefore I'm submitting the PR here. Please kindly make sure it goes into release process.

And please help cherry pick it back to `clang_tot_uprgade`.

whchung:
after discussion with @scchan , close this one and file a new one #624 directly against `clang_tot_upgrade`.

PR #624, opened 2018-01-31
[RUNTIME] fix potential undefined behavior in mutex locking / unlocking

whchung:
In the function changed in this commit, originally an RAII lock_guard is
present for a mutex. And in the later part of the function the unlock/lock
member functions of the mutex are directly invoked. Such codes is not ideal.

It was also proven certain user codes would run into soft hang situation.

In this commit we remove the lock_guard and always use lock/unlock member
function calls. The user codes affected above have been proven to run.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -2288,6 +2288,8 @@ public:
                 if (rq->_hccQueue == nullptr) {
                     DBOUT(DB_QUEUE, "Found unused rocrQueue=" << rq << " for thief=" << thief << ".  hwQueue=" << rq->_hwQueue << "\n")
                     foundRQ = rq;
+                    // update the queue pointers to indicate the theft
+                    foundRQ->assignHccQueue(thief);
                     break;
                 }
             }
@@ -2295,8 +2297,6 @@ public:
             this->rocrQueuesMutex.unlock();
 
             if (foundRQ != nullptr) {
-                // update the queue pointers to indicate the theft
-                foundRQ->assignHccQueue(thief);
                 break; // while !foundRQ
             }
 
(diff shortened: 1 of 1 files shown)


sunway513:
LGTM

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2260,64 +2260,74 @@ class HSADevice final : public KalmarDevice
...
 
-                    // update the queue pointers to indicate the theft:
-                    foundRQ->assignHccQueue(thief);
+            this->rocrQueuesMutex.unlock();
 
-                    break; // while !foundVictim
-                };
+            if (foundRQ != nullptr) {
+                // update the queue pointers to indicate the theft
+                foundRQ->assignHccQueue(thief);
```
scchan: should you do this right after this (https://github.com/RadeonOpenCompute/hcc/pull/624/files#diff-410c18478a22065083790061957b4770R2290) while you still have the lock; otherwise it looks like there will be a race condition

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2260,64 +2260,74 @@ class HSADevice final : public KalmarDevice
...
+                        break; // for
+                    }
+                }
             }
 
+            this->rocrQueuesMutex.unlock();
+
+            if (foundRQ != nullptr) {
+                // update the queue pointers to indicate the theft:
+                foundRQ->assignHccQueue(thief);
```
scchan: similar issue with the previous assignHccQueue

scchan (changes requested):

lib/hsa/mcwamp_hsa.cpp
whchung: thx. I'll change it right away.

whchung:
@scchan updated PR.

@sunway513 please run the test.

scchan (approved):
LGTM

PR #626, opened 2018-02-01
add an env var to change the default GPU

scchan:


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -3390,14 +3393,20 @@ class HSAContext final : public KalmarContext
         status = hsa_iterate_agents(&HSAContext::find_host, &host);
         STATUS_CHECK(status, __LINE__);
 
+        int first_gpu_index = Devices.size();
```
whchung: @scchan it'd be better to leave comments on the contents of what's inside `Devices`. in the current implementation they only contains GPU agents but I assume you want CPU agents inside in the subsequent PRs

whchung (changes requested):

scchan:
@whchung added comments to clarify the logic

whchung (approved):
LGTM

PR #636, opened 2018-02-05
Auto merge 201802051132 est

rocm-hcc:
Auto-upstream merge


scchan (changes requested):
Many files have their permission modified, are they needed?

PR #643, opened 2018-02-09
fix several race conditions in unpinned copy engine

scchan:
- some places are missing a wait on the dependent signal before the copy
- some copy loops would wait on the dependency when copying the 1st data chunk but subsequent chunks will not wait, which could result in data race


whchung (commented):
@scchan if i understand correct, `waitFor` would now never be changed. perhaps it'd be a good idea to change the signatures of relevant functions so they become `const hsa_signal_t *`?

scchan:
@whchung updated the copy functions' signature as you suggested

lib/hsa/unpinned_copy_engine.cpp
```diff
@@ -434,7 +431,7 @@ void UnpinnedCopyEngine::CopyDeviceToHostStaging(void* dst, const void* src, siz
 
                 dstp1 += theseBytes;
             }
-		}
+		    }
```
whchung: @scchan fix indentation?

whchung (commented):
with minor indentation fix i'm good with the PR. still need @bensander 's thumb up i think

scchan:
@whchung fixed indentation issue

whchung (approved):

whchung:
@bensander ping

whchung:
@scchan this is a heads up that we might need to cherry-pick commits in this PR to `roc-1.7.x`. I'll let you know.

whchung:
@scchan based on test results thus far this PR isn't needed in `roc-1.7.x`

PR #645, opened 2018-02-20
re-enable waitcnt insertion

scchan:


whchung:
@scchan I think this PR is good to be merged now

whchung (approved):

PR #648, opened 2018-02-22

rocm-hcc:
This is an upstream merge from commits:

compiler: edef10b8827c730595c95a666f287d1661e97e36
clang: 571f0763bbad15ba703be04be64814f71a8994da


rocm-hcc:
The Build check will fail the GL unit tests; as expected.  The build check failed because originally this PR picked up latest rocdl.  I pushed a reference to an older rocdl, which resolves the build error.

PR #658, opened 2018-03-15

rocm-hcc:


scchan:
it looks like there's a new regression, do we know anything about this? 
HCC :: Unit/Codegen/indirect-func-arg.cpp

scchan (changes requested):
new regression to be investigated

david-salinas:
investigating regression ....

PR #675, opened 2018-04-02
PR for auto_merge_201804020723_EST

rocm-hcc:


scchan (changes requested):
These two tests started failing with `PR for auto_merge_201803300723_EST` , could you please take a look?

    HCC :: Unit/Codegen/opt_level0.cpp
    HCC :: Unit/Codegen/opt_level1.cpp

PR #682, opened 2018-04-06
Fix printf seg fault when using extra args

aaronenyeshi:
This fix will fix a segmentation fault which occurs when we give printf more arguments than specifiers available. For example, using std::printf("test %hd\n", (int16_t) -1, tidx.global[0]); inside kernel.

### DIFF
--- a/include/hc_printf.hpp
+++ b/include/hc_printf.hpp
@@ -317,6 +317,7 @@ static inline void processPrintfPackets(PrintfPacket* packets, const unsigned in
       if (!std::regex_search(formatString, specifierMatches, specifierPattern)) {
         // More printf argument than format specifier??
         // Just skip to the next printf request
+        i+=(numPrintfArgs - j);
         break;
       }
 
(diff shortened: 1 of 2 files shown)


include/hc_printf.hpp
```diff
@@ -317,6 +317,7 @@ static inline void processPrintfPackets(PrintfPacket* packets, const unsigned in
       if (!std::regex_search(formatString, specifierMatches, specifierPattern)) {
         // More printf argument than format specifier??
         // Just skip to the next printf request
+        i++;
```
scchan: you are only skipping one unused arg, it should really be

`i+=(numPrintfArgs - j);`

scchan (changes requested):

scchan:
also could you add a new test?

aaronenyeshi:
I've amended the previous commit and added a new test.

scchan (approved):
LGTM

PR #687, opened 2018-04-09
[CMake][Driver][SDK][Runtime] remove libunwind dependency

whchung:
It has been verified certain applications with complex shared library loading
sequence would have exception handlers misplaced due to libunwind. In this
commit remove libunwind dependency from HCC in different aspect:

- CMake & HCC export CMake scripts
- hcc-config
- hc_rt_debug.h
- Dockerfile

In future commits it can be considered to make this an optional feature which
is enabled at CMake config time.


whchung:
@scchan as `Clang` submodule has also been changed at ( https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/131 ), it is required to have another submodule configuration change after this PR is merged.

Also please kindly help cherry pick this commit into `roc-1.7.x` branch so it goes into the next ROCm 1.7.x release.

include/hc_rt_debug.h
```diff
@@ -78,55 +76,7 @@ extern thread_local ShortTid hcc_tlsShortTid;
...
-      std::snprintf(loc, buffer_size, "0x%016lx:\t%s + 0x%lx\n", pc, print_func_name, offp);
-      bt += std::string(loc);
-
-      if (demangled)
-        free(demangled);
-    }
-    return bt;
-  }
-
   static void print_backtrace() {
```
scchan, line 79: we should just remove this function?

scchan (changes requested):
The clang PR has already been merged, could you update the link to clang with this PR?

https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/131

include/hc_rt_debug.h
whchung: @scchan there are other callsites for the function so I leave it here as an empty function for now. Later on we may want to reintroduce the logic and make it an optional feature.

whchung: @scchan also I added `clang` submodule into this PR

scchan (approved):
LGTM

PR #691, opened 2018-04-12
add additional checks in the hcc linker script

scchan:
When linking together the GPU LLVM IR modules, the linker script calls llvm-link to link all the objects but then pipe the output to opt.  As a consequence, errors from llvm-link (often fatal) won't get caught by the linker script and therefore, the compilation of the program would continue til the end.  This confuses users as hcc would produce an executable that appears correct but without any GPU code.

Additional checks were added to catch other misc errors.


whchung (approved):
LGTM. Thanks @scchan for making the script more robust!

PR #695, opened 2018-04-19
Fix unsupported gfx target by returning an empty isa

aaronenyeshi:
This will fix hcsparse and other apps which attempts to get isa for every target in the bundler. We should just return empty isa for unsupported targets.

### DIFF
--- a/hc2/headers/types/code_object_bundle.hpp
+++ b/hc2/headers/types/code_object_bundle.hpp
@@ -147,6 +147,7 @@ namespace hc2
     std::string isa_name(std::string triple)
     {
         static constexpr const char offload_prefix[]{"hcc-"};
+        const char* unsupported_gfx[] = {"801", "902"};
 
         triple = transmogrify_triple(triple);
         if (triple.empty()) return {};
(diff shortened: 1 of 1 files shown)


scchan:
For the APU, I think we should add support for the feature string (e.g. -xnack) in the triple processing.  
For other gfx arch not support by hcc, the proper way to handle that would be in the FE to check for targets explicitly support by hcc and issue an error for non-supported arch

hc2/headers/types/code_object_bundle.hpp
```diff
@@ -147,13 +147,22 @@ namespace hc2
...
         static constexpr const char offload_prefix[]{"hcc-"};
+        const char* unsupported_gfx[] = {"801", "902"};
 
         triple = transmogrify_triple(triple);
         if (triple.empty()) return {};
 
         triple.erase(0, sizeof(offload_prefix) - 1);
 
         static hsa_isa_t tmp{};
+
```
t-tye, line 158: This does not seem the right fix. Instead it seems the bug is in triple_to_hsa_isa() which is performing:

'''
throwing_hsa_result_check(
  hsa_isa_from_name(isa.c_str(), &r), __FILE__, __func__, __LINE__);
,,,

This function should not be throwing any exception. If given a string that is not a valid/supported-by-the-runtime it should return the NULL hsa_isa_t. The caller will then skip that target in its search for a suitable target to load.

I believe that was what the code originally reviewed did. If checked `` hsa_isa_from_name()`` resutlt was HSA_STATUS_SUCCESS.

aaronenyeshi: Thanks Tony, please review: https://github.com/RadeonOpenCompute/hcc/pull/698

PR #698, opened 2018-04-23
Fix throwing exception for RT old target queries

aaronenyeshi:
For backwards compatibility across different ROCm runtimes, we want to allow queries of older gfx targets instead of throwing an exception. It may be better to return a NULL isa_t. This is related to SWDEV-149306

### DIFF
--- a/hc2/headers/types/code_object_bundle.hpp
+++ b/hc2/headers/types/code_object_bundle.hpp
@@ -175,9 +175,9 @@ namespace hc2
     {
         const auto isa{isa_name(std::move(triple))};
 
-        if (isa.empty()) return hsa_isa_t({});
-
         hsa_isa_t r{};
+        if (isa.empty()) return r;
+
         if(HSA_STATUS_SUCCESS != hsa_isa_from_name(isa.c_str(), &r)) {
             r.handle = 0;
         }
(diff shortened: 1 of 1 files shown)


aaronenyeshi:
also should not compile for older gfx targets in HIP: https://github.com/ROCm-Developer-Tools/HIP/pull/414

aaronenyeshi:
HCC change is here: https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/134

aaronenyeshi:
@t-tye: Please review this patch, thanks!

hc2/headers/types/code_object_bundle.hpp
```diff
@@ -178,8 +178,9 @@ namespace hc2
         if (isa.empty()) return hsa_isa_t({});
 
         hsa_isa_t r{};
-        throwing_hsa_result_check(
-            hsa_isa_from_name(isa.c_str(), &r), __FILE__, __func__, __LINE__);
+        if(HSA_STATUS_SUCCESS != hsa_isa_from_name(isa.c_str(), &r)) {
+            r.handle = 0;
```
t-tye: To be consistent with the line above should this be:

`return hsa_isa_t({})`

t-tye (approved):
LGTM (with minor observation)

PR #702, opened 2018-04-28
PR for auto_merge_201804280723_EST

rocm-hcc:


scchan:
This doesn't seem right...  the merge is date is 28 April but clang's commit is 23 April which is 5 days behind.

PR #703, opened 2018-04-30
[Unpinned Copy Engine]:when initializing/resetting a signal use release ordering

scchan:
When initializing/resetting a signal use release ordering to ensure the store is properly ordered and visible to the GPU before it tries to set to a different value


whchung (approved):
LGTM

PR #709, opened 2018-05-03
Unpinned copy engine: wait for the dependent signal before entering the critical section

scchan:
This change is make all copy routines in unpinned copy engine to wait on the dependent HSA signal before acquiring the lock and entering into the critical section.  This will prevent potential deadlocks where a thread acquires the lock and then wait on a HSA signal from another thread trying to acquire the same lock to perform a copy.


whchung (approved):
@scchan Thanks for making this *IMPORTANT* fix. LGTM

PR #717, opened 2018-05-04
Removing more deprecated functions

aaronenyeshi:
Removing hsa_isa_get_info, hsa_executable_get_symbol and hsa_executable_create. 

Although replacement for hsa_executable_get_symbol is hsa_executable_get_symbol_by_linker_name, it is not yet implemented, so we will use less deprecated hsa_executable_get_symbol_by_name for now.


whchung (approved):
LGTM

PR #719, opened 2018-05-07
Remove grid launch mode

scchan:
HIP has stopped relying on the hc_grid_launch mode for a while so we'd want to start cleaning up hcc and remove hc_grid_launch in the compiler.  The first step would be to remove all grid launch related unit tests and to remove the invocation of the wrapper gen pass from the compiler driver script.  This change has been tested with HIP's unit tests and all of them passed.


whchung (approved):
@scchan Thanks for this PR. Time to put GridLaunch to rest. LGTM

PR #720, opened 2018-05-07
Purge grid launch - don't invoke DirectFuncCall pass

scchan:
remove the invocation and dependency of DirectFuncCall pass


whchung (approved):
LGTM

PR #722, opened 2018-05-09
Remove depr hsa_code_object_t - find ISA from ELF

aaronenyeshi:
Required to use new hsa_code_object_reader_t for previous usage of hsa_code_object_t. Remove deprecated functions hsa_code_object_deserialize and hsa_executable_load_code_object. Also isCompatible ISA should find the target from ELF header to make decision. Remove code_object from ISA verification.

### DIFF
--- a/hc2/external/elfio/elfio_amdgpu.hpp
+++ b/hc2/external/elfio/elfio_amdgpu.hpp
@@ -10,6 +10,8 @@
 
 namespace hc {
 
+// AMDGPU e_flags mirroring from llvm/include/llvm/BinaryFormat/ELF.h
+
 // AMDGPU specific e_flags.
 enum : unsigned {
   // Processor selection mask for EF_AMDGPU_MACH_* values.
--- a/hc2/external/elfio/elfio_amdgpu.hpp
+++ b/hc2/external/elfio/elfio_amdgpu.hpp
@@ -1,0 +1,89 @@
+//===----------------------------------------------------------------------===//
+//
+// This file is distributed under the University of Illinois Open Source
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef ELFIO_AMDGPU_HPP
+#define ELFIO_AMDGPU_HPP
+
+namespace hc {
+
+// AMDGPU specific e_flags.
+enum : unsigned {
+  // Processor selection mask for EF_AMDGPU_MACH_* values.
+  EF_AMDGPU_MACH = 0x0ff,
+
+  // Not specified processor.
+  EF_AMDGPU_MACH_NONE = 0x000,
+
+  // R600-based processors.
+
+  // Radeon HD 2000/3000 Series (R600).
+  EF_AMDGPU_MACH_R600_R600 = 0x001,
+  EF_AMDGPU_MACH_R600_R630 = 0x002,
+  EF_AMDGPU_MACH_R600_RS880 = 0x003,
+  EF_AMDGPU_MACH_R600_RV670 = 0x004,
+  // Radeon HD 4000 Series (R700).
+  EF_AMDGPU_MACH_R600_RV710 = 0x005,
+  EF_AMDGPU_MACH_R600_RV730 = 0x006,
+  EF_AMDGPU_MACH_R600_RV770 = 0x007,
+  // Radeon HD 5000 Series (Evergreen).
+  EF_AMDGPU_MACH_R600_CEDAR = 0x008,
+  EF_AMDGPU_MACH_R600_CYPRESS = 0x009,
+  EF_AMDGPU_MACH_R600_JUNIPER = 0x00a,
+  EF_AMDGPU_MACH_R600_REDWOOD = 0x00b,
+  EF_AMDGPU_MACH_R600_SUMO = 0x00c,
+  // Radeon HD 6000 Series (Northern Islands).
+  EF_AMDGPU_MACH_R600_BARTS = 0x00d,
+  EF_AMDGPU_MACH_R600_CAICOS = 0x00e,
+  EF_AMDGPU_MACH_R600_CAYMAN = 0x00f,
+  EF_AMDGPU_MACH_R600_TURKS = 0x010,
+
+  // Reserved for R600-based processors.
+  EF_AMDGPU_MACH_R600_RESERVED_FIRST = 0x011,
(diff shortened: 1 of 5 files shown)


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2675,24 +2698,32 @@ class HSADevice final : public KalmarDevice
...
+        reader.load(kern_stream);
+
+        // Get ISA from ELF header
+        std::string triple = "amdgcn-amd-amdhsa--gfx";
+        unsigned MACH = reader.get_flags() & 0x000000ff;
+
+        switch(MACH) {
+            case 0x023 : triple.append("701"); break;
+            case 0x02a : triple.append("803"); break;
+            case 0x02c : triple.append("900"); break;
```
whchung: @aaronenyeshi Can we avoid hard-coding ISA version here? We need to reduce future efforts required to support another new gfxXYZ.

whchung (changes requested):

lib/hsa/mcwamp_hsa.cpp
kzhuravl: @aaronenyeshi , those numbers are defined in here: https://github.com/RadeonOpenCompute/llvm/blob/0cd1006f4841272c27c95618acca4f0221b588fb/include/llvm/BinaryFormat/ELF.h#L691

aaronenyeshi: I can't seem to include llvm/include/llvm/BinaryFormat/ELF.h without it conflicting (compile error) with /usr/include/elf.h.

aaronenyeshi: @whchung I know there are many places we have target number such as lib/clamp-device, lit tests, Cmakes, clang driver. Maybe I should expose supported gfxXYZ in clang, and use the structure from there?

whchung: @aaronenyeshi / @kzhuravl compared with including LLVM headers I think it'd be better to do this within clang. Not sure how easy would it be though.

kzhuravl: Which things are you using from /usr/include/elf.h?

kzhuravl: Can you switch from /usr/include/elf.h to llvm's ELF.h?

aaronenyeshi: @AlexVlx Do you know if we can easily make the switch from /usr/include/elf.h to llvm's ELF.h?

scchan: it looks like .note section in the elf object contains the gfx string, can we get it from there?

kzhuravl: I highly do not recommend that. .note entry, which contains the gfx string, is going away in ROCM2.0.

scchan: @aaronenyeshi where is /usr/include/elf.h needed?

aaronenyeshi: @scchan I think in lib/hsa/mcwamp_hsa.cpp, there are many instances of `using namespace ELFIO;` and we use elf.h inside hc2/external/elfio/elfio_header.hpp

kzhuravl: `+` @t-tye 

Maybe a long-term goal, but: should not we switch from using elfio to llvm infrastructure?

Not sure what to do in short-term... Should you redefine flags?

aaronenyeshi: We've decided to mirror the llvm's ELF.h instead of using internal llvm header. It'll be easy to maintain instead of following LLVM's moving header.

scchan: @aaronenyeshi  I don't see where we include elf.h in elfio_header.hpp?

@kzhuravl it doesn't seem that llvm's elf provides an API for parsing the elf object or iterating through the symbols/sections, which elfio provides

aaronenyeshi:
@kzhuravl How is this change to mirror llvm's elf header inside of our ELFIO. There shouldn't be frequent updates to AMDGPU e_flags I believe.

lib/hsa/mcwamp_hsa.cpp
kzhuravl: @scchan , I do not think I agree, take a look at:
https://github.com/llvm-mirror/llvm/blob/master/include/llvm/Object/ELF.h#L52
https://github.com/llvm-mirror/llvm/blob/master/include/llvm/Object/ELFObjectFile.h#L201

kzhuravl:
@aaronenyeshi  They should not be frequently updated. Those LGTM, but @scchan or @whchung  should approve changes.

aaronenyeshi:
@whchung Do you think an unordered_map could work? Maybe I'll associate in this way: 
umap[EF_AMDGPU_MACH_AMDGCN_GFX900] = "gfx900" ? I will create a new commit for you to take a look here

AlexVlx:
@aaronenyeshi you don't need a map, the enum is descriptive and you have the switch already. If something changes WRT the map from enum to string you still need to update something, so you're not saving much by using a map here.

AlexVlx (approved):
LGTM.

aaronenyeshi:
@whchung @scchan what do you think about this implementation?

hc2/external/elfio/elfio_amdgpu.hpp
```diff
@@ -0,0 +1,89 @@
...
+// License. See LICENSE.TXT for details.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef ELFIO_AMDGPU_HPP
+#define ELFIO_AMDGPU_HPP
+
+namespace hc {
+
+// AMDGPU specific e_flags.
```
scchan: could you add a comment to document where these hex values are defined/coming from?

scchan (changes requested):

hc2/external/elfio/elfio_amdgpu.hpp
aaronenyeshi: Good point, just added it.

scchan (approved):

whchung (approved):
The PR looks good to me.

For future reference we probably need a guide to enumerate places within `hcc` (frontend , backend , runtime , device libs) in case a new GCN ISA target is to be added.

PR #724, opened 2018-05-10
Fix deadlock that may occur when disposing an HSAQueue

scchan:
This is a classic example where a thread acquires the mutex A then B.
Then another thread acquires the mutex B then A leading to a deadlock.
In HSAQueue::dispose(), it first acquires the mutex to the HSAQueue
itself, then procede to removeRocrQueue to acquire rocrQueuesMutex;
however in createOrstealRocrQueue(), those mutexes are acquires in
opposite order.  For reference the deadlock was discovered when
running the HipMultiThreadDevice-pyramid test.

Change-Id: Ie6c411cb7ee6c8726ec67cd866d17c18136c011c

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -2346,32 +2346,37 @@ public:
         }
     };
 
+
+private:
+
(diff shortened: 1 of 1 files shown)


whchung (approved):
LGTM. @scchan thanks for this crucial fix!

PR #734, opened 2018-05-17

aaronenyeshi:
This should include fixes for rocRAND build and u/sdot4/8 instructions.


scchan (approved):
verified locally

PR #737, opened 2018-05-22
getSignal() should tell the command launching paths to wait the queue for signals when hsa_signal_create failed

qianfengz:
The modification is to file hcc/lib/hsa/mcwamp_hsa.cpp.  The thought here is that,  in function getSignal(), when hsa_signal_create() returned with failure(caused by unavailable signal events from the kernel driver), getSignal() should return an "NULL" signal handle as indication of failing allocation, rather than call abort() directly.  And with all command launching paths where getSignal() are called, a checking should be done so that when "NULL" signal handle is returned , "hsaQueue()->wait()" is called to flush the queue

[hcc_mcwamp_hsa_cpp.diff.txt](https://github.com/RadeonOpenCompute/hcc/files/2025640/hcc_mcwamp_hsa_cpp.diff.txt)

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -3405,7 +3405,8 @@ public:
         DBOUT(DB_SIG,  " pre-allocate " << HCC_SIGNAL_POOL_SIZE << " signals\n");
         for (int i = 0; i < HCC_SIGNAL_POOL_SIZE; ++i) {
           hsa_signal_t signal;
-          status = hsa_signal_create(1, 0, NULL, &signal);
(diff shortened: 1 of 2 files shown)


scchan:
@fronteer We don't accept a patch submitted as an attachment.  Please put your change in your private branch and then create a pull request.  Also, we've had offline discussion that forcing a wait on the queue when signals have been exhausted is not the proper way to handle this situation.  I agree that we could improve this by not calling the abort() function in the hcc runtime.  I'd prefer that we propagate this exceptional situation back to the user application (e.g. perhaps throwing an exception) and let the application to decide how to handle it.

PR #740, opened 2018-05-24
disable lit pipefail by default

david-salinas:
LLVM upstream has a change that enables pipefail by default, in utils/lit/lit/TestRunner.py.  But pulling this upstream change into the HCC branch causes many unit tests to fail. This is because these HCC unit tests rely on pipefail being disabled.  So we need to ensure pipefail is disabled for HCC unit tests, before we merge the upstream change - which HCC needs to pass all LLVM lit tests.


scchan (approved):
LGTM

PR #748, opened 2018-06-05
adding HCC activity records callback

eshcherb:


whchung:
@eshcherb this PR shares the same issue with the one for HIP where the default branch of HCC, `clang_tot_upgrade`, shall be the target, instead of `roc-1.8.x`.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -5595,3 +5717,17 @@ std::ostream& operator<<(std::ostream& os, const HSAOp & op)
...
 
 //   - remove hsaqueeu
+
+////////////////////////////////////////////////////////////////////////////////////////
+// Profiling primitives
+records_dictionary_t* ActivityRecords::_records = NULL;
+activity_callback_t ActivityCallbacks::_fun = NULL;
+void* ActivityCallbacks::_arg = NULL;
+extern "C" __attribute__((visibility("default")))
+size_t HSAOp_get_record_size() { return sizeof(hcc_record_t); }
```
whchung, line 5727: it's not a good idea exposing interface in an implementation. I guess the idea for these functions are for upper-layer components to use. In such case a separate header file shall be defined.

whchung (changes requested):

whchung:
@eshcherb Could you explain the dependency of this PR with https://github.com/ROCmSoftwarePlatform/roctracer . It's not clear to me why there is no need to explicit include headers of `roctracer` in this PR.

lib/hsa/mcwamp_hsa.cpp
eshcherb: Yes, I agree, we can talk how to name it and where to put the header.

eshcherb: The changes were made against roc-1.8.x branch
To check them against 'master' I will reaply them th enext week I will be back from vacatino

eshcherb:
I think it is better to define activity records types in ROC-tracer and HIP/HCC will include external headers with the records definitions and to check runtime that the correct tracer version is used .

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -687,10 +690,60 @@ struct HSAOpCoord
...
+
+typedef uint64_t record_id_t;
+typedef std::map<uint32_t, record_id_t> records_dictionary_t;
+typedef void (*activity_callback_t)(uint32_t op_id, void* record, void* arg);
+
+class ActivityRecords {
+    public:
+    static void setRecord(const record_id_t& record_id) {
+        const uint32_t thread_id = syscall(__NR_gettid);
+        if (_records == NULL) _records = new records_dictionary_t;
```
scchan, line 721: Not sure why _records needs to be dynamically allocated?
This has a potential race condition since multiple threads may try to construct it at the same time.
I don't see where _records gets destructed?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -687,10 +690,60 @@ struct HSAOpCoord
...
+    OP_COPY_ID = 2,
+    OP_BARRIER_ID = 3,
+};
+
+// HCC activity record type
+struct hcc_record_t {
+  uint32_t domain;                                                    // activity domain id
+  uint32_t op_id;                                                     // operation id, dispatch/copy/barrier
+  uint32_t command_id;                                                // command kind
+  uint64_t correlation_id;                                            // activity correlation ID
```
scchan, line 705: Could you provide documentation on how this is being used?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -687,10 +690,60 @@ struct HSAOpCoord
...
+  uint64_t begin_ns;                                                  // host begin timestamp, nano-seconds
+  uint64_t end_ns;                                                    // host end timestamp, nano-seconds
+  int device_id;
+  uint64_t stream_id;
+  size_t bytes;
```
scchan, line 710: how `bytes` is being used?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -687,10 +690,60 @@ struct HSAOpCoord
     uint64_t    _queueId;
 };
 
+enum OpId {
+    OP_SYNC_ID = 0,
```
scchan, line 694: what the usage of OP_SYNC_ID?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -687,10 +690,60 @@ struct HSAOpCoord
...
+// HCC activity record type
+struct hcc_record_t {
+  uint32_t domain;                                                    // activity domain id
+  uint32_t op_id;                                                     // operation id, dispatch/copy/barrier
+  uint32_t command_id;                                                // command kind
+  uint64_t correlation_id;                                            // activity correlation ID
+  uint64_t begin_ns;                                                  // host begin timestamp, nano-seconds
+  uint64_t end_ns;                                                    // host end timestamp, nano-seconds
+  int device_id;
+  uint64_t stream_id;
```
scchan, line 709: From the usage, I think this should be called queue_id rather than stream_id

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -5595,3 +5717,17 @@ std::ostream& operator<<(std::ostream& os, const HSAOp & op)
...
+// Profiling primitives
+records_dictionary_t* ActivityRecords::_records = NULL;
+activity_callback_t ActivityCallbacks::_fun = NULL;
+void* ActivityCallbacks::_arg = NULL;
+extern "C" __attribute__((visibility("default")))
+size_t HSAOp_get_record_size() { return sizeof(hcc_record_t); }
+extern "C" __attribute__((visibility("default")))
+void HSAOp_set_activity_record(const record_id_t& record_id) { ActivityRecords::setRecord(record_id); }
+extern "C" __attribute__((visibility("default")))
+void HSAOp_set_activity_callback(activity_callback_t callback, void* arg) { ActivityCallbacks::set(callback, arg); }
```
scchan, line 5731: please provide description on how this infrastructure is being used.  It's not clear to me who and how the callback is setup

scchan (changes requested):
Could you provide documentation on where/how this infrastructure is being used?

The 1.8.x branch is a release branch so in general, it will only accept critical bug fixes.  Since this is a new feature, I won't be accepting this PR into 1.8.x unless there is a good justification (please send me and @whchung a PM to discuss).

PR #752, opened 2018-06-08
add git reference in GitSCM command for the submodules too

david-salinas:
We need the submodules class in the GitSCM command to also use the git reference.  The HCC Build/Test jobs are still timing out.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -54,7 +54,7 @@ node( 'hcctest' )
       $class: 'GitSCM',
       branches: scm.branches,
       doGenerateSubmoduleConfigurations: scm.doGenerateSubmoduleConfigurations,
-      extensions: scm.extensions + [[$class: 'CloneOption', reference: '/var/gitcache/hcc.git'], [$class: 'CleanCheckout'], [$class: 'SubmoduleOption', disableSubmodules: false, parentCredentials: false, recursiveSubmodules: true, reference: '', timeout: 60, trackingSubmodules: false]],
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -54,7 +54,7 @@ node( 'hcctest' )
       $class: 'GitSCM',
       branches: scm.branches,
       doGenerateSubmoduleConfigurations: scm.doGenerateSubmoduleConfigurations,
-      extensions: scm.extensions + [[$class: 'CloneOption', reference: '/var/gitcache/hcc.git'], [$class: 'CleanCheckout'], [$class: 'SubmoduleOption', disableSubmodules: false, parentCredentials: false, recursiveSubmodules: true, reference: '', timeout: 60, trackingSubmodules: false]],
+      extensions: scm.extensions + [[$class: 'CloneOption', reference: '/var/gitcache/hcc.git'], [$class: 'CleanCheckout'], [$class: 'SubmoduleOption', disableSubmodules: false, parentCredentials: false, recursiveSubmodules: true, reference: '/var/gitcache/hcc.git', timeout: 60, trackingSubmodules: false]],
```
scchan: Can we put the path into a variable?

scchan (changes requested):

Jenkinsfile
david-salinas: Sure, let me do that now ...

scchan (approved):
LGTM

PR #759, opened 2018-06-18
fix get_group_segment_sizes test 

david-salinas:
Now that we have enabled promote-ponter-kernargs-to-global pass, we can just add volatile to the group declaration.


AlexVlx (approved):
LGTM, thanks @david-salinas.

scchan (approved):
LGTM

PR #761, opened 2018-06-19
move LLD to amd-common branch

david-salinas:
I've also updated the manifest file in Gerrit.  Here is a link to the code review:

http://git.amd.com:8080/#/c/156597/

We will also need to update the upstream merge tool, to ignore (leave alone) the LLD submodule.


scchan (changes requested):
What is the purpose of the .project file?
Also I think the .gitmodules needs to be updated for the lld component to point to the amd-common branch

PR #762, opened 2018-06-20
move LLD to amd-common branch

david-salinas:
This PR moves HCC's LLD submodule to the amd-common branch in https://github.com/RadeonOpenCompute/lld .

I successfully built HCC and ran all HCC unit tests


scchan (approved):
LGTM

PR #764, opened 2018-06-21
Unpin host memory via host pointer instead of GPU pointer

whchung:
This commit fixes to logic in HCC runtime to be compatible with the semantic
asked by ROCR runtime API: hsa_amd_memory_unlock.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -441,19 +441,21 @@ namespace
 
             if (globals.find(x) != globals.cend()) return;
 
-            void* p = nullptr;
+            void* host_ptr =
+                reinterpret_cast<void*>(it1->second.first);
+            void* agent_ptr = nullptr;
             hsa_amd_memory_lock(
-                reinterpret_cast<void*>(it1->second.first),
+                host_ptr,
                 it1->second.second,
                 // Awful cast because ROCr interface is misspecified.
                 const_cast<hsa_agent_t*>(all_agents().data()),
                 all_agents().size(),
-                &p);
+                &agent_ptr);
 
             hsa_executable_agent_global_variable_define(
-                executable, agent, x.c_str(), p);
+                executable, agent, x.c_str(), agent_ptr);
 
-            globals.emplace(x, RAII_global{p, hsa_amd_memory_unlock});
+            globals.emplace(x, RAII_global{host_ptr, hsa_amd_memory_unlock});
         }
     }
 


scchan (approved):
LGTM

fxkamd:
Unpinning with the GPU pointer should work. There was a bug in the Thunk that is being fixed right now.

Using the CPU pointer works OK as long as you never pin the same address multiple times with different sizes. Otherwise you'll get different GPU pointers for the different mappings, and the only way to unpin the correct one is to use the GPU pointer.

AlexVlx:
@fxkamd OK, I'm confused now. The HSA extension itself is specified in terms of the host accessible address, see https://github.com/RadeonOpenCompute/ROCR-Runtime/blob/bfc4b9e98cb5e48d9f96287371d518cb444968ba/src/inc/hsa_ext_amd.h#L1205. Are we going to change that? Furthermore, do we not pin at page granularity, which basically entails that you'd only get different GPU pointers when crossing pages? @hthangirala, can you comment here?

P.S.: strictly in terms of taste and ease of understanding, unpinning the GPU pointer seems significantly more natural, and lends itself to easier wrapping in e.g. unique_ptr, so it would be nice if that were guaranteed to work, however the current specification does not reflect that intent, Thunk bug notwithstanding.

fxkamd:
Pinning and unpinning of host memory supports unaligned pointers and sizes. They don't need to be page granularity.

We had to make some subtle changes to the semantics of the APIs to support OpenCL use cases allowing the same pointer to be pinned multiple times with proper reference counting etc. In some cases the same pointer pinned multiple times can return different GPU pointers. Therefore the CPU pointer cannot be used reliably as a unique identifier for unpinning. This may not have been reflected properly in the spec.

If you allow pinning the same pointer multiple times, potentially with different sizes, I recommend you use the GPU pointer for unpinning.

AlexVlx:
@fxkamd, I wasn't thinking about the user's perspective here, but rather the actual side effect of pinning. Are we doing something beyond what the kernel does here because mlock works at page granularity last unless memory betrays me; this is honest curiosity as from a runtime implementer's perspective these are relevant nuances.

fxkamd:
Right, pinning and mapping memory for GPU access works at page granularity. But the Thunk will do the necessary rounding and return a GPU pointer with the same offset from the start of the page. So these APIs can be used with arbitrary pointers.

AlexVlx:
Thank you for the clarification, much appreciated.

scchan:
@AlexVlx this needs your review and approval, could you take a look?

AlexVlx:
@scchan @whchung My understanding is that this is getting fixed in Thunk (per @fxkamd) so that the current code would be correct (and potentially more flexible). I've not yet reasoned fully whether or not one can construct a scenario where a global is aliased by another global of a different size (seems doable if unlikely). I'm not opposed to this change per se, but we probably need to keep the above on the radar, and possibly revert in a version or two.

AlexVlx (approved):
LGTM modulo the comment I provided elsewhere.

PR #765, opened 2018-06-21
A heuristic to flush only a fraction of the queue when we are out of space for a new kernel

sryap:
When the capacity of the asyncOps vector reaches the
MAX_INFLIGHT_COMMANDS_PER_QUEUE threshold, flush only the first
fraction of the vector to overlap the cleanup process (operator
nullification and GC in removeAsyncOp()) with the ongoing GPU
computation.  The fraction size is tunable via the environment
variable HCC_QUEUE_FLUSHING_RATIO.  The default fraction size is half
of the queue.

Signed-off-by: Sarunya Pumma <sarunya.pumma@amd.com>

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -138,6 +138,10 @@ int HCC_PROFILE_VERBOSE=0x1F;
 
 char * HCC_PROFILE_FILE=nullptr;
 
+int HCC_QUEUE_FLUSHING_RATIO=50;
+
+double QUEUE_FLUSHING_FRAC;
+
 // Profiler:
 // Use str::stream so output is atomic wrt other threads:
 #define LOG_PROFILE(op, start, end, type, tag, msg) \
@@ -1405,7 +1409,6 @@ public:
 
         youngestCommandKind = op->getCommandKind();
 
-        drainingQueue_ = false;
 
         if (DBFLAG(DB_QUEUE)) {
             printAsyncOps(std::cerr);
@@ -1562,8 +1565,13 @@ public:
 
 
         bool foundFirstValidOp = false;
+        int oldAysncOpsSize = asyncOps.size();
+        int lastWaitOp = oldAysncOpsSize - 1;
+        if (drainingQueue_) {
+            lastWaitOp = (oldAysncOpsSize * QUEUE_FLUSHING_FRAC) - 1;
+        }
 
-        for (int i = asyncOps.size()-1; i >= 0;  i--) {
+        for (int i = lastWaitOp; i >= 0;  i--) {
             if (asyncOps[i] != nullptr) {
                 auto asyncOp = asyncOps[i];
                 if (!foundFirstValidOp) {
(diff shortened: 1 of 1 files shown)


whchung:
Empirical benchmark results shows ~6% wall clock time improvement for resnet50 model on TensorFlow 1.3 on 1GPU. ~12% improvement on 4GPUs.

whchung:
@sarunyap will need your help to provide more background information wrt to this PR, and how it could be improved further.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1556,8 +1555,13 @@ class HSAQueue final : public KalmarQueue
 
 
         bool foundFirstValidOp = false;
+        int oldAysncOpsSize = asyncOps.size();
+        int numWaitOps = oldAysncOpsSize - 1;
+        if (drainingQueue_) {
+            numWaitOps = oldAysncOpsSize / 2;
```
whchung: Please notice this is just a heuristic and might have to be tuned per application. So maybe it's better to expose it as an environment variable.

sryap: @whchung I have added the environment variable, namely HCC_QUEUE_FLUSHING_RATIO, to tune the percentage of the queue to be flushed when the queue is full.  The percentage has to be greater than zero.  Any invalid value will be set to the default value.  The default HCC_QUEUE_FLUSHING_RATIO is 50, which is 50 percent of the queue.

scchan (changes requested):
Wouldn't it be simpler to adjust the MAX_INFLIGHT_COMMANDS_PER_QUEUE limit rather than introducing new logic for this?

whchung:
@sarunyap can chime in to share more details but to my understanding enlarging max infight commands in the queue may actually cause more siginificant perf loss because draining the queue now takes longer time and GPU would be idling. Perhaps we need empirical data from @sarunyap

scchan:
I mean you could lower the limit so that the drainqueue flag gets set sooner and there will be less ops to drain?

whchung:
@scchan thanks! That’s indeed a valid argument. I think we do need experiment data then @sarunyap

sryap:
@scchan @whchung Thank you for reviewing my changes.  I think lowering the limit of MAX_INFLIGHT_COMMANDS_PER_QUEUE will not solve the problem because HCC will still wait for all operators in the queue to complete and the cleanup (removeAsyncOp()) of the last operator in the queue will be done when GPU is sitting idle.  This will only make the cleanup faster, but the cleanup time will still show up as a gap in the total time.

What my changes do is to hide the cleanup process of the last operator that we wait behind the ongoing GPU computation.

bensander:
Right we want to avoid draining the queue to zero so we can reclaim some of the resources but also keep the gpu doing useful work

scchan (approved):

scchan:
@whchung any concern or any additional discussion needed for this to be merged?

whchung:
@scchan sorry for belated reply. no concern here.

PR #771, opened 2018-06-26
fix missing system scope release for some situations

scchan:
- after a device to device memory copy, signal that the next sync needs a system release
- before a copy, even when there's no dependent async op, if the queue indicates that the next sync needs a system release, enqueue a marker with system scope release
- add extra debug message to getter/setter for the nextSyncNeedsSysRelease flag

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -5247,6 +5247,10 @@ hsa_status_t HSACopy::hcc_memory_async_copy(Kalmar::hcCommandKind copyKind, cons
     // HSA memory copy requires a system-scope acquire before the next kernel command - set flag here so we remember:
     hsaQueue()->setNextKernelNeedsSysAcquire(true);
 
+    // If it's a device to device, the next system sync will need a release
+    if (copyKind ==  Kalmar::hcMemcpyDeviceToDevice)
+      hsaQueue()->setNextSyncNeedsSysRelease(true);
+
     return status;
 }
 
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -1313,8 +1313,14 @@ public:
     bool nextKernelNeedsSysAcquire() const { return _nextKernelNeedsSysAcquire; };
     void setNextKernelNeedsSysAcquire(bool r) { _nextKernelNeedsSysAcquire = r; };
 
-    bool nextSyncNeedsSysRelease() const { return _nextSyncNeedsSysRelease; };
-    void setNextSyncNeedsSysRelease(bool r) { _nextSyncNeedsSysRelease = r; };
+    bool nextSyncNeedsSysRelease() const { 
+      DBOUT( DB_CMD2, "  HSAQueue::nextSyncNeedsSysRelease(): " <<  _nextSyncNeedsSysRelease << "\n");
+      return _nextSyncNeedsSysRelease; 
+    };
+    void setNextSyncNeedsSysRelease(bool r) {
+      DBOUT( DB_CMD2, "  HSAQueue::setNextSyncNeedsSysRelease(" <<  r << ")\n");
+      _nextSyncNeedsSysRelease = r; 
+    };
 
     uint64_t getSeqNum() const { return queueSeqNum; };
 
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -5298,7 +5298,7 @@ HSACopy::enqueueAsyncCopyCommand(const Kalmar::HSADevice *copyDevice, const hc::
 
 
         int depSignalCnt = 0;
-        hsa_signal_t depSignal;
+        hsa_signal_t depSignal = { .handle = 0x0 };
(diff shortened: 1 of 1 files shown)


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -5241,6 +5247,10 @@ hsa_status_t HSACopy::hcc_memory_async_copy(Kalmar::hcCommandKind copyKind, cons
     // HSA memory copy requires a system-scope acquire before the next kernel command - set flag here so we remember:
     hsaQueue()->setNextKernelNeedsSysAcquire(true);
 
+    // If it's a device to device, the next system sync will need a release
+    if (copyKind ==  Kalmar::hcMemcpyDeviceToDevice)
+      hsaQueue()->setNextSyncNeedsSysRelease(true);
```
whchung, line 5252: @scchan if i understand correctly, this corresponds to the first item in the description:

```
after a device to device memory copy, signal that the next sync needs a system release
```

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -5288,29 +5298,32 @@ HSACopy::enqueueAsyncCopyCommand(const Kalmar::HSADevice *copyDevice, const hc::
...
-        // We need to ensure the copy waits for preceding commands the HCC queue to complete, if those commands exist.
-        // The copy has to be set so that it depends on the completion_signal of the youngest command in the queue.
+        depAsyncOp = std::static_pointer_cast<HSAOp> (hsaQueue()->detectStreamDeps(this->getCommandKind(), this));
         if (depAsyncOp) {
             depSignal = * (static_cast <hsa_signal_t*> (depAsyncOp->getNativeHandle()));
+        }
 
+        // We need to ensure the copy waits for preceding commands the HCC queue to complete, if those commands exist.
+        // The copy has to be set so that it depends on the completion_signal of the youngest command in the queue.
+        if (depAsyncOp || fenceScope != hc::no_scope) {
```
whchung, line 5317: I believe this corresponds to the 2nd item in the description:

```
before a copy, even when there's no dependent async op, if the queue indicates that the next sync needs a system release, enqueue a marker with system scope release
```

May I understand the rationale behind this statement? It's less obvious than the 1st item where D2D copies always require a system scope release.

lib/hsa/mcwamp_hsa.cpp
scchan: yes, that's correct

whchung:
@jeffdaily Could you help cherry-pick this PR and see if it help alleviates the predicament in https://github.com/ROCmSoftwarePlatform/tensorflow-upstream/issues/27 ? From the look of it they should be different issues but I'd like to give it a shot.

lib/hsa/mcwamp_hsa.cpp
scchan: If you refer the code before my patch, it would only go into the logic to create a new marker ONLY if there exists a depAsyncOp (!= nullptr).
The new logic says that even if there's no depAsyncOn, enqueues a marker with system scope release anyway if the queue's nextSyncNeedsSysRelease flag is true.  This is the case that this change is trying to address.

whchung: thanks. it makes sense now.

I'm trying to imagine some offending scenarios based on this new logic. earlier version of this comment was incorrect and i'm wondering could you share some known failing cases?

scchan: For D2D, the scenario exposing the problem looks like:

kernel writes to buffer_A (nextSyncNeedsSysRelease set to true) --> copy buffer_A to buffer_B on the same device (nextSyncNeedsSysRelease incorrectly cleared) --> copy buffer_B to host (incorrect since no system release)

In other case, the scenario looks like:

kernel writes to buffer_A (nextSyncNeedsSysRelease set to true) --> user created event/marker clears the async ops tables --> copy buffer_A to host (no async op, the copy logic didn't insert a marker with system release)

whchung: @scchan I'm wondering would the aggregate effect of the PR causing we _always_ use system scope barriers, thus effectively turning off `HCC_OPT_FLUSH`?

scchan: no, I don't think so.  They don't interfere with HCC_OPT_FLUSH.
For the D2D case, it's a bug that hcc runtime is not tracking the state of nextSyncNeedsSysRelease correctly after a device memory to device memory copy.  
For the case where the async op table is empty but the queue says that the nextSyncNeedsSysRelease, the guard for creating a new sync is incorrect.

bensander:
Thx sc.  I agree with analysis and the fixes. Lgtm

jeffdaily:
@whchung  I tried this PR but it didn't fix https://github.com/ROCmSoftwarePlatform/tensorflow-upstream/issues/27.

scchan:
@jeffdaily this PR fixes memory synchronization problems, it doesn't address memory fault issues

lib/hsa/mcwamp_hsa.cpp
whchung: thanks for the clarification.

whchung (approved):
LGTM

PR #774, opened 2018-06-27
set failing tests as expected fails

david-salinas:
Adding XFAIL for the following tests, so they are counted as expected fails.

  HCC :: Unit/AMDGPU/register-control.cpp
    HCC :: Unit/AmpShortVectors/amp_short_vectors_2files.cpp
    HCC :: Unit/AmpShortVectors/amp_short_vectors_2files_1.cpp
    HCC :: Unit/Codegen/deser_def_body_compound_support_inheritclass.cpp


scchan:
I merged your PR for the register control attribute fix in clang, could you update this PR to remove register-control.cpp test?

scchan (changes requested):
exclude register-control.cpp since the fix has been submitted to clang

david-salinas:
Ok, I've removed the register-control test from the known_fails, and remove XFAIL from the test source.

scchan (approved):

PR #781, opened 2018-07-03
Fixed aligned_alloc ptr tracker range issue

gargrahul:
Fixed aligned_alloc ptr tracker range issue. This was creating issues while freeing device memories allocated after aligned memory allocation.

### DIFF
--- a/lib/hsa/hc_am.cpp
+++ b/lib/hsa/hc_am.cpp
@@ -279,7 +279,7 @@ auto_voidp am_aligned_alloc(std::size_t sizeBytes, hc::accelerator &acc, unsigne
                             ptr = NULL;
                         } else {
                             hc::AmPointerInfo ampi(ptr/*hostPointer*/, ptr /*devicePointer*/, unalignedPtr, sizeBytes, acc, false/*isDevice*/, true /*isAMManaged*/);
-                            g_amPointerTracker.insert(ptr,ampi);
+                            g_amPointerTracker.insert(unalignedPtr,ampi);
 
                             // Host memory is always mapped to all possible peers:
                             auto accs = hc::accelerator::get_all();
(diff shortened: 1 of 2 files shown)


scchan (changes requested):
Could you add a unit test for the scenario that you are trying to fix?

gargrahul:
@scchan Is there any hcc test for testing am_aligned_alloc already? I don't have HCC test though I have a HIP test for this.

scchan:
no, if you do a grep for am_alloc in tests/Unit, there are many tests using am_alloc, can you modify one of them for am_aligned_alloc?

scchan (changes requested):
Is there another way to test this other than printing out the tracker information?  Also it looks like it always passes which is problematic for a unit test.

gargrahul:
I took an existing unit test for modification. The existing unit test also passes always.

https://github.com/RadeonOpenCompute/hcc/blob/2e7519287cc90b02fd4049eeae2d50fba014ac83/tests/Unit/AM/am_memtracker.cpp

Moreover this test should always pass since all allocated memories should get freed always.

If you run this test without my fix, you will see one tracker entry despite 2 allocations., That was the problem.

scchan:
Without your fix, if someone query for the aligned pointer with `am_memtracker_getinfo`, it would return `AM_ERROR_MISC` since the aligned pointer isn't being tracked?

gargrahul:
Without my fix, after aligned memory allocation if someone allocated another device memory, due to overlapping single allocation was listed in tracker info:-
          base-           end: #SeqNum        HostPtr         DevPtr    SizeBytes   SizeMB Dev?  Reg? AppId AppFlags Peers
   0x902010000-   0x90205ffff: #     1              0    0x902010000       327680     0.31 DEV  ALLOC     0        0 0x1e2c5f0

With fix, there are two allocations shown as below:-
base-           end: #SeqNum        HostPtr         DevPtr    SizeBytes   SizeMB Dev?  Reg? AppId AppFlags Peers
   0x901a16000-   0x901a65fff: #     1              0    0x901a20000       327680     0.31 DEV  ALLOC     0        0 0x146a6f0
   0x901a66000-   0x901a85fff: #     2              0    0x901a66000       131072     0.12 DEV  ALLOC     0        0 0x146a6f0

scchan (approved):
LGTM, thanks

PR #783, opened 2018-07-03
fix mGPU race condition in unpinned copy engine

jeffdaily:
This PR is a potential fix for https://github.com/ROCmSoftwarePlatform/tensorflow-upstream/issues/27.  Each GPU has its own instance of `copyDevice`, and each `copyDevice` has its own copy engine instance.  Each copy engine simultaneously attempts to lock the same host memory in H2D copies since there is no mutex across engines.  This PR creates such a class-scoped mutex and set for managing the locked pointers.

### DIFF
--- a/lib/hsa/unpinned_copy_engine.cpp
+++ b/lib/hsa/unpinned_copy_engine.cpp
@@ -27,8 +27,6 @@ THE SOFTWARE.
 #include "unpinned_copy_engine.h"
 #include "hc_rt_debug.h"
 
-static std::mutex _copyLockFileScope;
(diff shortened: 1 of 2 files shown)


whchung:
@jeffdaily Could you publish a docker container so HCC maintainers could validate with TensorFlow should the PR gets improved?

scchan:
rather than fixing the pin-in-place algorithm, we decide to make unpinned copy engine to always use the staging buffer algorithm: https://github.com/RadeonOpenCompute/hcc/pull/791

PR #791, opened 2018-07-10
Unpinned copy engine: use staging buffer by default

scchan:
The pin-in-place algorithm has race a condition where multiple
threads would try to pin/unpin the same buffer concurrently.
There is no easy way for the hcc runtime to track the pointers
to protect against that and experiments show no performance
benefit with the pin-in-place algorithm.

This patch changes the unpinned copy engine to use the staging
algorithm by default, which is a lot safer.  The staging buffer
size has been adjusted to 4 MB, which seems to be the sweetspot
based on test results.  An environment variable has been added
for adjusting the staging buffer size.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -93,7 +93,7 @@ int HCC_PRINT_ENV=0;
 
 int HCC_SIGNAL_POOL_SIZE=512;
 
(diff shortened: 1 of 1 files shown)


scchan:
this addresses https://github.com/ROCmSoftwarePlatform/tensorflow-upstream/issues/27

whchung (approved):
LGTM

PR #793, opened 2018-07-11
initial protyping for std/hc vector

PhilipDeegan:
works with [this example](https://gist.github.com/Dekken/92cfcc313e539e1446f669f5c3cd5324)

my text editor is set up to remove trailing white spaces so sorry about that

I don't specifically like the way this is done as _data/_data_host require a default constructor which may not be the best idea

### DIFF
--- a/include/hc.hpp
+++ b/include/hc.hpp
@@ -194,7 +194,7 @@ public:
      * the parent accelerator.
      */
     // FIXME: dummy implementation now
-    bool get_is_debug() const { return 0; } 
+    bool get_is_debug() const { return 0; }
 
     /**
      * Performs a blocking wait for completion of all commands submitted to the
@@ -205,8 +205,8 @@ public:
      *                     hcWaitModeActive would be used to reduce latency with
      *                     the expense of using one CPU core for active waiting.
      */
-    void wait(hcWaitMode waitMode = hcWaitModeBlocked) { 
-      pQueue->wait(waitMode); 
+    void wait(hcWaitMode waitMode = hcWaitModeBlocked) {
+      pQueue->wait(waitMode);
       Kalmar::getContext()->flushPrintfBuffer();
     }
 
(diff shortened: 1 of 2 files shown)


AlexVlx:
Hello @Dekken, thank you for the PR. I do have a few concerns - to begin with:
1) what is this trying to solve?
2) choosing the name `vector` is extremely unfortunate, as it'll be extremely clashy (we already suffer this when it comes to `array`) and inculcate the expectation that this IS A `std::vector` which it is not (I've noted the inheritance from `std::vector` which is a faux-pas, as the standard library containers are not intended to be inherited from - observe the absence of a virtual destructor, for example).

PhilipDeegan:
1. vectors are useful.
2. the name is arbitrary, but  I see your point regarding the destructor

PhilipDeegan:
Compile time array sizes are all well and good but not always feasible to know in advance.
Using the "new" operator manually goes against the notions of RAII and thus C style arrays are not always the best choice 

Hence a "vector-like" type would resolve both of these

Renaming the type is not an issue IMO - and as you suggest not inheriting from a standard container is not wise - reject to rework

AlexVlx:
I'm still a bit confused as to what gap you observe - `array_view` interops with `std::` containers, `array` is dynamically sized at runtime (that's another terrible name choice though, if only we could travel through time and make a different one). Where do you need to use `operator new` / why can you not use `array_view` on top of a `std::vector` - perhaps answering these can drive further investigation?

PhilipDeegan:
Yes I see how you can use a std::vector with array_view now, however this is *somewhat* restrictive I think as it requires the vector to be completed populated in order to access the ".size()" to then create the array_view - unless I'm mistaken

I'm just curious as to why you would create a wrapper class if you don't need it

PhilipDeegan:
Updated to use composition and changed the name to "hector" @ https://github.com/0xz/hcc/commit/6e4ae265dc91efca46e07de0ca4e57c7e6fae2af

PR #799, opened 2018-07-12
Modify any, all and ballot to use ockl functions and llvm instrinsics

aaronenyeshi:
Not sure if this is needed if HIP will be getting their own functions for __any, __all, and __ballot. But right now that hasn't merged so they are still using hcc's.

### DIFF
--- a/include/hc.hpp
+++ b/include/hc.hpp
@@ -2778,8 +2778,9 @@ extern "C" inline unsigned int __activelanecount_u32_b1(unsigned int input) __HC
  * wavefront and return non-zero if and only if predicate evaluates to non-zero
  * for any of them.
  */
+extern "C" __attribute__((const)) bool __ockl_wfany_i32(int) __HC__;
 extern "C" inline int __any(int predicate) __HC__ {
-    return __popcount_u32_b64(__activelanemask_v4_b64_b1(predicate));
+    return __ockl_wfany_i32(predicate);
 }
 
 /**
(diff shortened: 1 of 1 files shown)


include/hc.hpp
```diff
@@ -2778,17 +2778,19 @@ extern "C" inline unsigned int __activelanecount_u32_b1(unsigned int input) __HC
  * wavefront and return non-zero if and only if predicate evaluates to non-zero
  * for any of them.
  */
+extern "C" __attribute__((const)) bool __ockl_wfany_i32(int) __HC__;
```
b-sumner: Drop the const attribute.

include/hc.hpp
```diff
@@ -2778,17 +2778,19 @@ extern "C" inline unsigned int __activelanecount_u32_b1(unsigned int input) __HC
...
-    return __popcount_u32_b64(__activelanemask_v4_b64_b1(predicate));
+    return __ockl_wfany_i32(predicate);
 }
 
 /**
  * Evaluate predicate for all active work-items in the
  * wavefront and return non-zero if and only if predicate evaluates to non-zero
  * for all of them.
  */
+extern "C" __attribute__((const)) bool __ockl_wfall_i32(int) __HC__;
```
b-sumner: Drop the const attribute

b-sumner (dismissed):

include/hc.hpp
aaronenyeshi: I am wondering why we don't want to match attribute from ockl? https://github.com/RadeonOpenCompute/ROCm-Device-Libs/blob/e1c3a3efbf884a300df4a56aeb8d433a954a14ba/ockl/inc/ockl.h#L209

b-sumner: I'm dropping the attribute from OCKL.  It isn;t correct for these uses of the function.

scchan (approved):

PR #805, opened 2018-07-20
Revert "After a device-to-device copy, set the nextSyncNeedsSysReleas…

scchan:
…e flag to true"

This setting of the need sys release flag in the original patch was incorrect and unnecessary, in fact, it causes a regression in one of the HIP stream copy tests.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -5254,10 +5254,6 @@ hsa_status_t HSACopy::hcc_memory_async_copy(Kalmar::hcCommandKind copyKind, cons
     // HSA memory copy requires a system-scope acquire before the next kernel command - set flag here so we remember:
     hsaQueue()->setNextKernelNeedsSysAcquire(true);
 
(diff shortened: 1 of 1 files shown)


whchung:
@scchan could you elaborate why setting system scope release fence after D2D copy is incorrect? I thought the original design from @bensander was to ensure L2$ on the receiving end gets flushed?

scchan:
This is a change I put in recently as part of this PR: https://github.com/RadeonOpenCompute/hcc/commit/81d8cf3ed8e61b96c512756cab440275c2bf4366#diff-410c18478a22065083790061957b4770R5251

The case you mentioned should not be affect by this change since the receiving side needs a system scope acquire before using the buffer.

Initially I thought that the sys release is required before the d2d copy on the sending gpu, but the nextSyncNeedsSysRelease flag only controls synchronization after the copy.  However, I haven't looked into this deep enough to understand why the extra sys release would cause the HIP stream test to fail

whchung (approved):
Thanks for the clarification.  LGTM

PR #806, opened 2018-07-21
Fix detect stream deps

scchan:
When checking whether there are any pending async ops, check if the async ops vector is empty() but also check whether the last op has a valid pointer because sometimes HSAQueue::removeAsyncOps would just reset the shared_ptr<HSAOp> in the async op vector rather than removing the shared_ptr from the vector.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -1433,7 +1433,7 @@ public:
 
         assert (newCommandKind != hcCommandInvalid);
 
-        if (!asyncOps.empty()) {
+        if (!asyncOps.empty() && asyncOps.back().get()!=nullptr) {
             assert (youngestCommandKind != hcCommandInvalid);
 
             // Ensure we have not already added the op we are checking into asyncOps,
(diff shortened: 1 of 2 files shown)


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1406,9 +1406,8 @@ class HSAQueue final : public KalmarQueue
             wait();
         }
         op->asyncOpsIndex(asyncOps.size());
-        asyncOps.push_back(op);
```
whchung, line 1409: @scchan why is necessary to move this statement after `op->getCommandKind()`?

whchung (commented):
@scchan other than the question regarding to the location of  `asyncOps.push_back(op)` the reset of the PR looks good to me. Surprised that we didn't catch this before.

lib/hsa/mcwamp_hsa.cpp
scchan: I changed `push_back` by moving `op` into the vector instead of making a new copy.  I swapped the order so that `op->getCommandKind()` is not referencing an object in an invalid state.

whchung (approved):
Thanks for the clarification. LGTM

PR #819, opened 2018-07-31
Added support for pinned 2d SDMA copy

gargrahul:
Added support for 2d SDMA copy for pinned memory. This requires rocr which supports hsa_amd_memory_async_copy_rect.


scchan:
@gargrahul  please update this PR based on ROCm 1.9.2's implementation.  Thanks

PR #824, opened 2018-08-02
Update some debug messages

scchan:
- changed debug messages guarded by KALMAR_DEBUG to be controlled by debug flags in hc_rt_debug.h
- remove KALMAR_DEBUG
- some clean up in hc_rt_debug.h

### DIFF
--- a/include/hc_rt_debug.h
+++ b/include/hc_rt_debug.h
@@ -43,25 +43,20 @@ static std::vector<std::string> g_DbStr = {"api", "cmd", "wait", "aql", "queue",
 // DBOUT(" Something happened" << myId() << " i= " << i << "\n");
 #define COMPILE_HCC_DB 1
 
-#define DBFLAG(db_flag) (HCC_DB & (1<<(db_flag)))
+#define DBFLAG(db_flag) (COMPILE_HCC_DB && (HCC_DB & (1<<(db_flag))))
 
 #define DBSTREAM std::cerr
 
 // Use str::stream so output is atomic wrt other threads:
 #define DBOUT(db_flag, msg) \
-if (COMPILE_HCC_DB && (HCC_DB & (1<<(db_flag)))) { \
+if (DBFLAG(db_flag)) { \
     std::stringstream sstream;\
     sstream << "   hcc-" << g_DbStr[db_flag] << " tid:" << hcc_tlsShortTid._shortTid << " " << msg ; \
     DBSTREAM << sstream.str();\
 };
 
 // Like DBOUT, but add newline:
-#define DBOUTL(db_flag, msg) \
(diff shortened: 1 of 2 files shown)


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1749,9 +1735,8 @@ class HSAQueue final : public KalmarQueue
         // do read
         if (dst != device) {
             if (!getDev()->is_unified()) {
-#if KALMAR_DEBUG
-                std::cerr << "read(" << device << "," << dst << "," << count << "," << offset << "): use HSA memory copy\n";
-#endif
+                DBOUT(DB_COPY, "read(" << device << "," << dst << "," << count << "," << offset
```
aaronenyeshi, line 1738: Should we use DBOUTL instead of adding the \n to end of DBOUT for these outputs?

include/hc_rt_debug.h
```diff
@@ -43,25 +43,21 @@ static std::vector<std::string> g_DbStr = {"api", "cmd", "wait", "aql", "queue",
...
+#define DBFLAG(db_flag) (COMPILE_HCC_DB && (HCC_DB & (1<<(db_flag))))
 
-#define DBSTREAM std::cerr
+#define DBSTREAM  std::cerr
+#define DBWSTREAM std::wcerr
 
 // Use str::stream so output is atomic wrt other threads:
 #define DBOUT(db_flag, msg) \
-if (COMPILE_HCC_DB && (HCC_DB & (1<<(db_flag)))) { \
+if (DBFLAG(db_flag)) { \
```
aaronenyeshi, line 53: Would it be useful to add DBWFLAG which can replace the debug outputs that look like:
```
        if (DBFLAG(DB_INIT)) {
          DBWSTREAM << L"Path: " << path << L"\n";
          DBWSTREAM << L"Description: " << description << L"\n";
        }
```
with `DBWOUTL(DB_INIT,  L"Path: " << path << L);` and `DBWOUTL(DB_INIT,  L"Description: " << description << L);`

aaronenyeshi (commented):
Other than my comments, looks good to me

lib/hsa/mcwamp_hsa.cpp
scchan: I won't change existing debug messages.  In fact I wouldn't consider DBOUTL a necessity in the first place

include/hc_rt_debug.h
scchan: Going through the runtime code, I could only find once instance where the extra macro would help so I would not add DBWOUT at this point.

aaronenyeshi (approved):
LGTM

PR #827, opened 2018-08-03
Remove trampolines and all of the associated creaky machinery (and many associated cleanups)

AlexVlx:
This is a catch-all PR which includes many bits alongside the trampoline purging. Things that can be separated shall be separated into compact PRs that focus on a singular aspect (e.g. updating ELFIO or switching over the atomics), but this is presented like so to create a perception of the scope of the work. At the moment it is incomplete, since it's missing `array` and `array_view`, and contains various WiP stubs - I considered it useful to start reviewing early. It goes in tandem with https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/148.


AlexVlx:
@sameerds FYI.

PR #834, opened 2018-08-10
Add custom timers to profile HCC runtime

sryap:
A prototype for printing HCC runtime profiles.  Changes are as
follows:
1) Add get_host_timestamp() to get Unix timestamp in microseconds.
2) Add print_hcc_runtime_profile() to print the HCC runtime profiles
   in a specific format.
3) Print host time and GPU time for host time and GPU time conversion
   in HSAContext's constructor
4) Print profile of HSAQueue::wait()
5) Add HCC_PROFILE_VERBOSE masks to print profiles

Note that this PR is per @whchung request.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -45,6 +45,7 @@
 #include "hc_printf.hpp"
 
 #include <time.h>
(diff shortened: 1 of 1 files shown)


scchan:
hcc rt now relies on roctracer for gathering profiling traces

PR #838, opened 2018-08-14
hc::array operator[] should check for accessibility 

scchan:
- access to elements in hc::array on the host should check for accessibility
- KalmarDevice's access_type should be initialized based on HSA device profile


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2348,6 +2348,22 @@ class HSADevice final : public KalmarDevice
...
+        access_type at;
+        status = hsa_agent_get_info(agent, HSA_AGENT_INFO_PROFILE, &agentProfile);
+        STATUS_CHECK(status, __LINE__);
+        if (agentProfile == HSA_PROFILE_BASE) {
+            profile = hcAgentProfileBase;
+            at = access_type_none;
+        } else if (agentProfile == HSA_PROFILE_FULL) {
+            profile = hcAgentProfileFull;
+            at = access_type_read_write;
+        }
```
aaronenyeshi, line 2393: do you need an else statement for `profile = hcAgentProfileNone;` ?

aaronenyeshi (commented):
LGTM except my one comment

david-salinas (approved):
LGTM.

PR #840, opened 2018-08-16
Add support for all ROCm supported targets. Fix triple to match AMDGPU.

AlexVlx:
This adds back all targets supported by ROCm as per https://llvm.org/docs/AMDGPUUsage.html#processors. It also normalises the triple to match that specified by the same source. Goes hand in hand with https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/149.


aaronenyeshi:
I think you will have to add the new targets here as well: https://github.com/RadeonOpenCompute/hcc/blob/clang_tot_upgrade/lib/hsa/mcwamp_hsa.cpp#L2672

The MACH codes exist here: https://github.com/RadeonOpenCompute/hcc/blob/clang_tot_upgrade/hc2/external/elfio/elfio_amdgpu.hpp#L66

aaronenyeshi (approved):
LGTM

PR #841, opened 2018-08-20
Add missing bitselect, bitinsert and bitextract funcs

aaronenyeshi:


scchan (changes requested):
Could you add a unit test for these?  Thanks.

include/hc.hpp
```diff
@@ -2407,10 +2407,12 @@ extern "C" inline unsigned int __popcount_u32_b64(unsigned long long int input)
  * Please refer to <a href="http://www.hsafoundation.com/html/Content/PRM/Topics/05_Arithmetic/bit_string.htm">HSA PRM 5.7</a> for more detailed specification of these functions.
  */
 extern "C" inline unsigned int __bitextract_u32(unsigned int src0, unsigned int src1, unsigned int src2) __HC__ {
-  return (src0 << (32 - src1 - src2)) >> (32 - src2);
+  return (src2&31) == 0 ? 0 : (src0 << (32 - (src1&31) - (src2&31))) >> (32 - (src2&31));
```
scchan: can we add meaningful variables for src1&31 and src2&31 so that the algorithm is more understandable?  Same for 64-bit bitextract, bitinsert

scchan (changes requested):

include/hc.hpp
aaronenyeshi: Thanks, just added it, also now the test for bitinsert_u64 works

scchan (approved):

PR #846, opened 2018-08-26
activity API

eshcherb:


scchan (changes requested):
As discussed in several offline conversations, I have several concerns regarding
   - extensive state (queue ID, device ID) tracking in the runtime with TLS variables, the poor visibility from runtime in terms of how these states were set and used and the manpulation of these states through extern functions by unknown parties.
   - timestamp collection mechanism
   - usage of record IDs
   - instrumentation for memory copies
   - the OP IDs

More details have been provided in the internal issue tracking system

eshcherb:
closing this since #937 is a rebase of this PR

PR #851, opened 2018-08-30
[Config] move compiler/llvm reference to amd-common

david-salinas:
HCC customizations were merged into amd-common branch.  Updating the reference for HCC build


scchan (changes requested):
reason for changing the cycle and cycle2 tests?

david-salinas:
sorry, I accidentally included these changed test in the commit ... removing the changes from this commit.  I also forgot to update the .gitmodules file.

PR #852, opened 2018-08-31
Use llvm-objdump to generate the ISA dump

scchan:
- When generating an ISA dump, use llvm-objdump to disassemble
  the binary instead of inoking the backend to recompile the IR
  to ASM text
- Update extractkernel to match the llvm-objdump switches used
  for the ISA dump


b-sumner:
llvm-objdump doesn't produce a result that will assemble to the same thing as assembling the back-end produced assembly.  This may not be desirable for those using the dumps.

scchan:
superseded by other PRs that updated clamp-device and extractkernel

PR #853, opened 2018-08-31
Combine the opt invocations in clamp-device and clamp-link

scchan:
For historical reason, we had to invoke opt load a customized pass
to fix up the address spaces for global pointers before invoking
opt again for general optimizations on the "fixed" IR.  This
is no longer needed and therefore there's no reason why there
should be 2 separate opt invocations


aaronenyeshi (approved):
LGTM

PR #865, opened 2018-09-12
replace __hc_ math functions with __ocml_ functions

david-salinas:
This resolves the issue with the function name changes in the device libs for native functions.  But it also replaces the __hc_ function names in kalmar_math.h with the __ocml_ function names.

### DIFF
--- a/include/kalmar_math.h
+++ b/include/kalmar_math.h
@@ -10,286 +10,288 @@
 #include <cmath>
 #include <stdexcept>
 
-extern "C" __fp16 __hc_acos_half(__fp16 x) restrict(amp);
-extern "C" float __hc_acos(float x) restrict(amp);
-extern "C" double __hc_acos_double(double x) restrict(amp);
 
-extern "C" __fp16 __hc_acosh_half(__fp16 x) restrict(amp);
-extern "C" float __hc_acosh(float x) restrict(amp);
-extern "C" double __hc_acosh_double(double x) restrict(amp);
+extern "C" __fp16 __ocml_acos_half(__fp16 x) [[hc]];
+extern "C" float __ocml_acos_f32(float x) [[hc]];
+extern "C" double __ocml_acos_f64(double x) [[hc]];
 
-extern "C" __fp16 __hc_asin_half(__fp16 x) restrict(amp);
-extern "C" float __hc_asin(float x) restrict(amp);
-extern "C" double __hc_asin_double(double x) restrict(amp);
+extern "C" __fp16 __ocml_acosh_half(__fp16 x) [[hc]];
+extern "C" float __ocml_acosh_f32(float x) [[hc]];
+extern "C" double __ocml_acosh_f64(double x) [[hc]];
 
-extern "C" __fp16 __hc_asinh_half(__fp16 x) restrict(amp);
-extern "C" float __hc_asinh(float x) restrict(amp);
-extern "C" double __hc_asinh_double(double x) restrict(amp);
+extern "C" __fp16 __ocml_asin_half(__fp16 x) [[hc]];
+extern "C" float __ocml_asin_f32(float x) [[hc]];
+extern "C" double __ocml_asin_f64(double x) [[hc]];
 
-extern "C" __fp16 __hc_atan_half(__fp16 x) restrict(amp);
-extern "C" float __hc_atan(float x) restrict(amp);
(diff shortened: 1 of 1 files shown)


include/kalmar_math.h
```diff
@@ -10,286 +10,288 @@
...
-
-extern "C" __fp16 __hc_log_half(__fp16 x) restrict(amp);
-extern "C" float __hc_log(float x) restrict(amp);
-extern "C" float __hc_log_native(float x) restrict(amp);
-extern "C" double __hc_log_double(double x) restrict(amp);
 
-extern "C" __fp16 __hc_log10_half(__fp16 x) restrict(amp);
-extern "C" float __hc_log10(float x) restrict(amp);
-extern "C" double __hc_log10_double(double x) restrict(amp);
+extern "C" __fp16 __ocml_acos_half(__fp16 x) [[hc]];
```
b-sumner, line 14: The correct OCML suffix is _f16 not _half.  Also should hcc be using __fp16 directly here?  Isn't there a better typedef like "half"?

AlexVlx: Actually this should use _Float16, which is the Clang kosher 16-bit float type. `typedef` half is awful though because there are a great many uses of that word, that need not have anything to do with a type (e.g. a function that halves). It's unlikely the C++ Standard will take `half` as the type name precisely for that reason (it's prime real-estate). Boost tries to emulate `<cstdint>` for FP here: https://www.boost.org/doc/libs/1_65_1/boost/math/cstdfloat/cstdfloat_types.hpp. But _Float16 is nice and explicative and also avoids the trap of looking too cute.

include/kalmar_math.h
```diff
@@ -1538,91 +1540,91 @@ namespace Kalmar
...
         HCC_MATH_LIB_FN
-        float sqrtf(float x) { return __hc_sqrt(x); }
+        float sqrtf(float x) { return __ocml_sqrt_f32(x); }
 
         HCC_MATH_LIB_FN
-        __fp16 sqrt(__fp16 x) { return __hc_sqrt_half(x); }
+        __fp16 sqrt(__fp16 x) { return __ocml_sqrt_half(x); }
 
         HCC_MATH_LIB_FN
         float sqrt(float x) { return precise_math::sqrtf(x); }
```
scchan, line 1565: change to call __ocml_sqrt_f32 instead

include/kalmar_math.h
```diff
@@ -1538,91 +1540,91 @@ namespace Kalmar
...
         HCC_MATH_LIB_FN
-        float tgammaf(float x) { return __hc_tgamma(x); }
+        float tgammaf(float x) { return __ocml_tgamma_f32(x); }
 
         HCC_MATH_LIB_FN
-        __fp16 tgamma(__fp16 x) { return __hc_tgamma_half(x); }
+        __fp16 tgamma(__fp16 x) { return __ocml_tgamma_half(x); }
 
         HCC_MATH_LIB_FN
         float tgamma(float x) { return precise_math::tgammaf(x); }
```
scchan, line 1577: replace this with the __ocml_tgamma_f32
In fact, please replace all the ones with the precise_math namespace with an ocml equivalent

include/kalmar_math.h
scchan: `make test` didn't catch the problem so looks like there's a gap in the unit tests.  Could you add unit tests for these?

scchan (changes requested):

include/kalmar_math.h
david-salinas: OpenCL has the "half" type:  https://www.khronos.org/registry/OpenCL/sdk/1.0/docs/man/xhtml/scalarDataTypes.html

AlexVlx: That is of no consequence here (what OpenCL decided to do). OpenCL doesn't have as rich an ecosystem, doesn't support overloading etc. Let us not copy their less inspired ideas. It's also debatable if having seas of cute type aliases actually helps with anything.

aaronenyeshi (changes requested):
I agree with other reviews. To summarize requests:
1) Please use the *_f16 suffix to replace the *_half suffix in ocml math functions. 
2) Instead of type __fp16, please use _Float16. 
3) Some unit tests needed for these _Float16 math functions.
4) Change precise_math namespace functions to use ocml equivalent.
For 1,2 and 4, just need find and replace keywords. For 3, add to existing math tests the half versions.

PR #876, opened 2018-09-20
Remove AMP headers. Remove restrict use. Refactor tests to just use HC.

AlexVlx:
This is a rather mechanical first step, which removes the headers and the usage of C++AMP specific constructs whilst shifting everything to HC. This does outline a few things we've not implemented for HC such as tiled_extent::truncate, for example. These will be addressed with HC Next.


tests/Unit/Atomic/atomic_xor_global.cpp
```diff
@@ -1,9 +1,9 @@
 // RUN: %cxxamp %s -o %t.out && %t.out
```
aaronenyeshi, line 1: do we need to clean up the RUN line's cxxamp or leave it as it is?

tests/Unit/Design/pass_by_ref.cpp
```diff
@@ -11,47 +11,47 @@
...
   }
  private:
-  Concurrency::array_view<int> a_, b_;
-  Concurrency::array_view<int>& c_;
+  hc::array_view<int> a_, b_;
+  hc::array_view<int>& c_;
 };
-void bar(void) restrict(amp,cpu) {
+void bar(void) [[cpu, hc]] {
   int* foo = reinterpret_cast<int*>(&myVecAdd::__cxxamp_trampoline);
```
aaronenyeshi, line 27: what is cxxamp_trampoline?

aaronenyeshi (approved):
LGTM - Thanks Alex. I haven't read everything in detail, this patch is very long. Should we remove: https://github.com/RadeonOpenCompute/hcc/wiki#c-amp

tests/Unit/Design/pass_by_ref.cpp
AlexVlx: It's a compiler generated entry point for the kernel (long-ish story). It's going out as well as part of the AMP removal / HC Next refactoring, but that's part of another PR.

tests/Unit/Atomic/atomic_xor_global.cpp
AlexVlx: Eventually yes, we'll remove that one.

tests/Unit/Design/pass_by_ref.cpp
aaronenyeshi: Okay sounds good to me

scchan (approved):
LGTM

AlexVlx:
@scchan @aaronenyeshi laddies, can we progress this one / merge it? It would very much ease with the staging of the following steps, and make sure that we don't run afoul of the upcoming cutoff. Thank you.

scchan:
@AlexVlx   there are still 3 unexpected failures, could you address them please?

********************
Failing Tests (3):
    HCC :: Unit/Copy/copy.cpp
    HCC :: Unit/Design/transpose.cpp
    HCC :: Unit/Indexing/tile_index.cpp

AlexVlx:
No they are expected, I think I already explained this over Slack? Could you please review that discussion?

PR #878, opened 2018-09-21
remove the -force-import-weak switch when invoking thinlto

scchan:


aaronenyeshi:
If you remove this option, will the cross-module import still import all weak functions if a stronger function doesn't exist?

scchan:
yes, you are right that weak functions won't get imported.  However, the code to bypass the weak symbol check was not merged into the amd-common branch so this switch is currently invalid.  I don't think we could upstream this patch either.
I'm thinking the strategy to deal with this is to do a noopt link and then run the amdgpu internalize symbol pass to resolve the weak symbols first before running thinlto.

aaronenyeshi (approved):
I'm not sure if this will effect anyone's work flow, but we need it since its not properly enabled

PR #886, opened 2018-10-03
Implement queue priority for accelerator::create_view()

mangupta:
queue_priority in accelerator_view is mapped to hsa_amd_queue_priority_t. Requires ROCm 1.9 or higher.

TODO:
* In this patch, I enforce the same limit for the max hsa queues we can allocate for in each priority. Do we want to have finer control over it i.e. different limits for each priority?
* The same rocrQueuesMutex is used irrespective of the queue priority. Do we want to use a different mutex for each priority? We will need to modify HSAQueue::dispose in that case.
* In this patch, the decision to hard or soft remove a rocrQueue is based on comparing number of hcc queues with the sum of the rocr queues across all priorities. Do we want to track hcc queues per priority as well? Alternatively we could modify queue stealing to steal an unused rocrQueue from another priority pool and change the priority as required.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -2255,9 +2255,9 @@ public:
 
         this->rocrQueuesMutex.lock();
 
-        // TODO: Do we want the max queues for each priority to be different?
         // Allocate a new queue when we are below the HCC_MAX_QUEUES limit
-        if (rocrQueues[priority].size() < HCC_MAX_QUEUES) {
+        auto rqSize = rocrQueues[0].size()+rocrQueues[1].size()+rocrQueues[2].size();
+        if (rqSize < HCC_MAX_QUEUES) {
             foundRQ = new RocrQueue(agent, this->queue_size, thief, priority);
             rocrQueues[priority].push_back(foundRQ);
             DBOUT(DB_QUEUE, "Create new rocrQueue=" << foundRQ << " for thief=" << thief << "\n")
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -1157,18 +1157,20 @@ struct RocrQueue {
     }
 
     RocrQueue(hsa_agent_t agent, size_t queue_size, HSAQueue *hccQueue, queue_priority priority)
+        : _priority(priority)
     {
         // Map queue_priority to hsa_amd_queue_priority_t
+        hsa_amd_queue_priority_t queue_priority;
         switch (priority) {
             case priority_low:
-                _priority = HSA_AMD_QUEUE_PRIORITY_LOW;
+                queue_priority = HSA_AMD_QUEUE_PRIORITY_LOW;
                 break;
             case priority_high:
-                _priority = HSA_AMD_QUEUE_PRIORITY_HIGH;
+                queue_priority = HSA_AMD_QUEUE_PRIORITY_HIGH;
                 break;
             case priority_normal:
             default:
-                _priority = HSA_AMD_QUEUE_PRIORITY_NORMAL;
+                queue_priority = HSA_AMD_QUEUE_PRIORITY_NORMAL;
                 break;
         }
 
@@ -1181,8 +1183,8 @@ struct RocrQueue {
         STATUS_CHECK(status, __LINE__);
 
         // Set queue priority
-        status = hsa_amd_queue_set_priority(_hwQueue, _priority);
-        DBOUT(DB_QUEUE, "  " <<  __func__ << ": set priority for HSA command queue: " << _hwQueue << " to " << _priority << "\n");
+        status = hsa_amd_queue_set_priority(_hwQueue, queue_priority);
+        DBOUT(DB_QUEUE, "  " <<  __func__ << ": set priority for HSA command queue: " << _hwQueue << " to " << queue_priority << "\n");
         STATUS_CHECK(status, __LINE__);
 
         // TODO - should we provide a mechanism to conditionally enable profiling as a performance optimization?
@@ -1215,7 +1217,7 @@ struct RocrQueue {
     // Track profiling enabled state here. - no need now since all hw queues have profiling enabled.
 
     // Priority could be tracked here:
-    hsa_amd_queue_priority_t _priority;
+    queue_priority _priority;
 };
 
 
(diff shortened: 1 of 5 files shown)


mangupta:
I have also created a HC unit test case that tests create_view with different priorities. I am yet to push the commit for the same since I had changed the order of the create_view arguments. The change impacts the execute_order test case. Not sure if it impacts any existing pure-HC application.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2228,15 +2248,31 @@ class HSADevice final : public KalmarDevice
     UnpinnedCopyEngine::CopyMode  copy_mode;
 
     // Creates or steals a rocrQueue and returns it in theif->rocrQueue
-    void createOrstealRocrQueue(Kalmar::HSAQueue *thief) {
+    void createOrstealRocrQueue(Kalmar::HSAQueue *thief, queue_priority priority = priority_normal) {
         RocrQueue *foundRQ = nullptr;
 
+        // Map Kalmar::enums:queue_priority to rqIndex which is used for indexing into rocrQueues
+        int rqIndex = 1;
+        switch (priority) {
```
scchan: I think it would be more clear to use `priority` to index into rocrQueues rather than creating another index `rqIndex`

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1156,16 +1156,33 @@ struct RocrQueue {
         STATUS_CHECK(status, __LINE__);
     }
 
-    RocrQueue(hsa_agent_t agent, size_t queue_size, HSAQueue *hccQueue)
+    RocrQueue(hsa_agent_t agent, size_t queue_size, HSAQueue *hccQueue, queue_priority priority)
     {
+        // Map queue_priority to hsa_amd_queue_priority_t
```
scchan, line 1162: It seems to me that it would be better that `_priority` has the type `queue_priority` rather than `hsa_amd_queue_priority_t` since the hsa priority is only needed in this ctor to set the priority by calling the hsa API.  This would save you from doing another reserve mapping when removing the queue from the pool

tests/Unit/HC/queue_priority.cpp
```diff
@@ -0,0 +1,34 @@
...
+  accelerator_view av_high_priority = acc.create_view(execute_in_order, queuing_mode_automatic, priority_high);
+
+  // test dispatch a kernel to av
+  parallel_for_each(av, extent<1>(1), []() [[hc]] {});
+
+  // test dispatch a kernel to av_normal_priority
+  parallel_for_each(av_normal_priority, extent<1>(1), []() [[hc]] {});
+
+  // test dispatch a kernel to av_high_priority
+  parallel_for_each(av_high_priority, extent<1>(1), []() [[hc]] {});
```
scchan: Could you add a unit test that actually produces some results and that could be verified?

scchan (dismissed):

lib/hsa/mcwamp_hsa.cpp
mangupta: ```priority``` is an enum. Using an enum for index arrays is usually a bad practice. Especially if we add additional enum values later or change the associated values later. Hence the mapping. However, if you prefer using the enum for indexing, I can clean this up.

lib/hsa/mcwamp_hsa.cpp
mangupta: The reverse mapping when removing the queue would only go away if we don't mind using enums values for array indexing. Otherwise there would still be a mapping. This can be removed in case we go with enum based array indexing.

tests/Unit/HC/queue_priority.cpp
mangupta: Do you mean that the test should prove that higher priority queues indeed have a higher priority? For that we would need to simultaneously launch two long running kernels - one on normal priority and other on higher priority queue. And then measure the kernel execution time to determine that higher priority queues indeed work. I am not sure how to go about doing that in HC. I have one in HIP that is functional (see https://github.com/ROCm-Developer-Tools/HIP/pull/703/files#diff-9a4ff6e79f138c757779b6102d9e2699).

lib/hsa/mcwamp_hsa.cpp
scchan: I don't think creating another set of ids that needs to be maintained manually would be a better solution to the issue you described.  The enumerators have defined values so just need documentation that we use them to get to the right queue pool.  You could use a map if the value of enumerators are non-contiguous but that seems overkill

lib/hsa/mcwamp_hsa.cpp
scchan: yes, switching to `queue_priority` for indexing is a precondition

tests/Unit/HC/queue_priority.cpp
scchan: I'm not sure if there's a way to write a robust unit test to test priority by relying on timestamps...
For this, I'm fine with having a test that shows the array_views created with different priority flags are functional

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2228,15 +2250,16 @@ class HSADevice final : public KalmarDevice
...
         RocrQueue *foundRQ = nullptr;
 
         this->rocrQueuesMutex.lock();
 
+        // TODO: Do we want the max queues for each priority to be different?
         // Allocate a new queue when we are below the HCC_MAX_QUEUES limit
-        if (rocrQueues.size() < HCC_MAX_QUEUES) {
-            foundRQ = new RocrQueue(agent, this->queue_size, thief);
-            rocrQueues.push_back(foundRQ);
+        if (rocrQueues[priority].size() < HCC_MAX_QUEUES) {
```
scchan: `HCC_MAX_QUEUES` should be the overall limit.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2315,12 +2338,16 @@ class HSADevice final : public KalmarDevice
         // This defers expensive queue deallocation if an hccQueue that holds an hwQueue is destroyed -
         // keep the hwqueue around until the number of hccQueues drops below the number of hwQueues
         // we have already allocated.
-        auto rqSize = rocrQueues.size();
+        // TODO: Do we want to track HCC queues independently for each priority?
+        // Or maybe modify queue stealing to steal unused HSA queue from other priority pools &
+        // change priority as required?
+        auto rqSize = rocrQueues[0].size()+rocrQueues[1].size()+rocrQueues[2].size();
```
scchan, line 2344: I think this heuristic needs to be refined since it doesn't seem right that we only compare the global sizes and then decide on whether to remove a queue for a particular priority

scchan (dismissed):

scchan:
@AlexVlx ping... could you review this PR?  thanks

lib/hsa/mcwamp_hsa.cpp
mangupta: By overall limit I guess you mean the total number of rocrQueues across all priorities should not exceed ```HCC_MAX_QUEUES```.

lib/hsa/mcwamp_hsa.cpp
mangupta: How do you want this to be handled? As mentioned in the TODO, ```hccSize``` currently tracks number of av instances across all priorities. Do we want to track it independently for each priority?
Is it fine to address this in a follow up patch?

AlexVlx (approved):
As far as I am concerned this is OK, it doesn't seem any worse than the stuff that is already there. I shall not delve into the overall usefulness of this stuff as it's a different kettle of fish. What I will point out is that the test is pretty useless, as it doesn't do anything specific to at least validate that priority is present - setting queue priority may be a NOP or completely unimplemented, and that would still pass. At a minimum, it should at least try to use the spammy accessor to validate that the priority it tried to set is present. It's up to @scchan if that is blocking or not.

lib/hsa/mcwamp_hsa.cpp
scchan: ok, I'm fine with a follow up PR

lib/hsa/mcwamp_hsa.cpp
scchan: yes, `HCC_MAX_QUEUES` should be the total number of queues

scchan (approved):
LGTM

PR #891, opened 2018-10-06
-Werror ignoring return value of function

PhilipDeegan:
hcc/include/kalmar_aligned_alloc.h:30:5: error: ignoring return value of function declared with 'warn_unused_result' attribute [-Werror,-Wunused-result]
    posix_memalign(&memptr, alignment, size);

### DIFF
--- a/include/kalmar_aligned_alloc.h
+++ b/include/kalmar_aligned_alloc.h
@@ -27,7 +27,8 @@ inline void* kalmar_aligned_alloc(std::size_t alignment, std::size_t size) noexc
     }
     void* memptr = NULL;
     // posix_memalign shall return 0 upon successfully allocate aligned memory
-    posix_memalign(&memptr, alignment, size);
+    auto ignore = posix_memalign(&memptr, alignment, size);
+    (void) ignore;
     assert(memptr);
 
     return memptr;


PhilipDeegan:
oh weird!

include/kalmar_aligned_alloc.h
```diff
@@ -27,7 +27,8 @@ inline void* kalmar_aligned_alloc(std::size_t alignment, std::size_t size) noexc
     }
     void* memptr = NULL;
     // posix_memalign shall return 0 upon successfully allocate aligned memory
-    posix_memalign(&memptr, alignment, size);
+    auto ignore = posix_memalign(&memptr, alignment, size);
+    (void) ignore;
```
scchan, line 31: I think we could put an assert to make sure the return value is zero and that should fix the problem

scchan (changes requested):

include/kalmar_aligned_alloc.h
PhilipDeegan: does that work for -DNDEBUG ?

scchan: you are right, you'd still need `(void) ignore;` to make `-DNDEBUG` happy

scchan (approved):

PR #893, opened 2018-10-09
Remove the IRIF library from HCC

aaronenyeshi:
Since ROCm-Device-Libraries has removed the irif from being installed, it needs to be removed from HCC. IRIF has already been dropped in rocdl.


aaronenyeshi:
irif.amdgcn.bc file was removed since this commit: https://github.com/RadeonOpenCompute/ROCm-Device-Libs/commit/14e6b7b9c1d7c676ae1ab33387b1ebb331807468

scchan (approved):

PR #896, opened 2018-10-10
Remove deprecated hsa isa compatible call

aaronenyeshi:
Required to use the hsa_agent_iterate_isas for previous deprecated hsa_isa_compatible. Since an agent will allow multiple ISAs from now on, we need to check if code object is compatible with any of them. We leverage passing a struct into the callback function for checking isa handles.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -2695,8 +2695,10 @@ public:
                 agent,
                 [](hsa_isa_t agent_isa, void* data) {
                     isa_comp_data* co_data = static_cast<isa_comp_data*>(data);
-                    if (agent_isa == co_data->isa_type)
+                    if (agent_isa == co_data->isa_type) {
                         co_data->is_compatible = true;
+                        return HSA_STATUS_INFO_BREAK;
+                    }
                     return HSA_STATUS_SUCCESS;
                 },
                 &co_data);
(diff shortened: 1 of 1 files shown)


aaronenyeshi:
This commit passes build and all tests on my machine. However I haven't tried any agents with multiple ISAs. I think that requires APUs which support multiple ISA handles.

david-salinas (approved):
LGTM

aaronenyeshi:
ping reviewers @scchan @kzhuravl @AlexVlx , is this ready for merge?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2682,19 +2680,32 @@ class HSADevice final : public KalmarDevice
...
-        status = hsa_isa_compatible(co_isa, agentISA, &isCompatible);
+        co_data.is_compatible = false;
+        status =
+            hsa_agent_iterate_isas(
+                agent,
+                [](hsa_isa_t agent_isa, void* data) {
+                    isa_comp_data* co_data = static_cast<isa_comp_data*>(data);
+                    if (agent_isa == co_data->isa_type)
+                        co_data->is_compatible = true;
+                    return HSA_STATUS_SUCCESS;
```
scchan, line 2702: I think when currrent isa is incompatible then the callback should return `HSA_STATUS_INFO_BREAK` so it goes to the next isa.  See other hsa related callbacks like `get_memory_pools` for example in hcc RT

scchan (changes requested):

lib/hsa/mcwamp_hsa.cpp
aaronenyeshi: I've added that status_t as well, thanks!

scchan (approved):
LGTM

PR #897, opened 2018-10-11
[CMAKE] add dependencies to the cmake-test

scchan:
make sure that cmake-test only starts after the compiler and runtime has been built

### DIFF
--- a/cmake-tests/CMakeLists.txt
+++ b/cmake-tests/CMakeLists.txt
@@ -4,6 +4,7 @@ if(NOT CMAKE_CXX_COMPILER MATCHES ".*hcc")
 endif()
 
 add_executable(cmake-test cmake-test.cpp)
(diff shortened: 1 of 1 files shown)


cmake-tests/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@ if(NOT CMAKE_CXX_COMPILER MATCHES ".*hcc")
 endif()
 
 add_executable(cmake-test cmake-test.cpp)
+add_dependencies(cmake-test hc_am mcwamp_hsa mcwamp)
```
pfultz2: This should happen under `if(TARGET hccrt)` below, because this is ran both in build and standalone(in order to test `find_package(hcc)`).

Also `hc_am` is already added as a dependency through `target_link_libraries`.

PR #910, opened 2018-10-23
new tests for async streams

ssahasra:
This is a review-only request for now. This change should *not* be merged until after the OCKL implementation has landed in the device library.

### DIFF
--- a/tests/Unit/OCKL/AsyncStreams/drop-packets.cpp
+++ b/tests/Unit/OCKL/AsyncStreams/drop-packets.cpp
@@ -1,0 +1,101 @@
+// RUN: %hc %s -I%hsa_header_path -L%hsa_library_path -lhsa-runtime64 -o %t.out && %t.out -s
+
+#include "as_consumer.h"
+
+class test_handler_t : public ockl_as_consumer_t {
+  unsigned int *status;
+
+  bool check_packet(uint8_t service, uint64_t connection_id, uint8_t *payload,
+                    uint8_t bytes, uint8_t flags) {
+    if (service != 42)
+      return false;
+    if (bytes != sizeof(unsigned int))
+      return false;
+    if (flags != (__OCKL_AS_CONNECTION_BEGIN | __OCKL_AS_CONNECTION_END))
+      return false;
+    unsigned int pp = read_uint(payload);
+    if (pp >= NUM_THREADS)
+      return false;
+    return true;
+  }
+
+  bool check_final() {
+    if (stream->write_index != NUM_PACKETS_INSUFFICIENT)
+      return false;
+
+    unsigned int errorsExpected = NUM_THREADS - NUM_PACKETS_INSUFFICIENT;
+    for (int ii = 0; ii != NUM_THREADS; ++ii) {
+      switch (status[ii]) {
+      case __OCKL_AS_STATUS_OUT_OF_RESOURCES:
+        if (errorsExpected == 0)
+          return false;
+        --errorsExpected;
+        break;
+      case __OCKL_AS_STATUS_SUCCESS:
+        break;
+      default:
+        return false;
+      }
+    }
+    if (errorsExpected != 0)
+      return false;
+
+    return true;
+  }
+
+public:
+  test_handler_t(__ockl_as_stream_t *stream, unsigned int *s)
+      : ockl_as_consumer_t(stream), status(s){};
+};
+
+void dropPackets(__ockl_as_stream_t *stream, unsigned int *status,
+                 unsigned int tid) [[hc]] {
+  uint64_t connection_id;
+
+  status[tid] =
+      hc::as_write_block(stream, 42, &connection_id,
+                         reinterpret_cast<const unsigned char *>(&tid),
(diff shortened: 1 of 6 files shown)


ssahasra:
All the tests run in asynchronous as well as synchronous mode, except drop-packets.cpp, which is meant only for synchronous mode by definition. All tests pass with the latest version of the device library implementation.

tests/Unit/OCKL/AsyncStreams/mixed-producers.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+        if (flags & __OCKL_AS_CONNECTION_END) {
+            strRecd.insert(buf);
+            buffers.erase(connection_id);
+        }
+
+        return true;
+    }
+
+    bool check_final() {
```
b-sumner: This checks that each string was found, but doesn't check that the number of strings of each type that the kernel successfully sent is correct.  I think we should make sure that everything the kernel successfully sent is actually found and consumed.

tests/Unit/OCKL/AsyncStreams/multiple-services.cpp
```diff
@@ -0,0 +1,72 @@
...
+    for (unsigned char ii = 0; ii != 4; ++ii) {
+        __ockl_as_write_block(stream, ii, &connection_id, &ii, 1,
+                              __OCKL_AS_CONNECTION_BEGIN | __OCKL_AS_CONNECTION_END);
+    }
+}
+
+bool run_test()
+{
+    unsigned int num_packets = (synchronous_mode ? 4 : 1);
+    unsigned int num_threads = 1;
```
b-sumner: This test is really weak.  Would you make it stronger?

tests/Unit/OCKL/AsyncStreams/drop-packets.cpp
```diff
@@ -0,0 +1,89 @@
...
+    if (hsa_memory_allocate(*region, NUM_THREADS * sizeof(unsigned int), &status) != HSA_STATUS_SUCCESS)
+        return false;
+
+    test_handler_t test_handler(stream, (unsigned int *)status);
+
+    hc::extent<1> ex(NUM_THREADS);
+    hc::tiled_extent<1> et(ex.tile(THREADS_PER_BLOCK));
+
+    hc::completion_future cf = parallel_for_each(et, [=](hc::tiled_index<1> & idx) [[hc]] {
+                                                         dropPackets(stream, (unsigned int*)status, idx.global[0]);
```
AlexVlx: Things LGTM in general, but this indentation is the stuff of nightmares, please do not use it, just indent normally. As a general observation, since it is not uncommon for tests to be taken for examples of how to use a language / API, I think I would prefer if we used C++ named casts instead of the cute and all too dangerous C ones. Whilst it is not of consequence for your tests, it may be of consequence for somebody else who will get the wrong impression that C-casts are cool (they are not).

tests/Unit/OCKL/AsyncStreams/mixed-producers.cpp
ssahasra: Fixed. check_final() now checks that the number of strings received of each type is correct.

tests/Unit/OCKL/AsyncStreams/multiple-services.cpp
ssahasra: Removed this file entirely. All the other files now check for the service ID too.

ssahasra:
Waiting for comments from @b-sumner on the recently uploaded changes to the tests. Once these tests are accepted, the plan is to update the equivalent HIP tests too, and then move the review forward on the device library implementation under ockl.

AlexVlx (changes requested):

b-sumner:
I'm not really understanding doing hcc before hip but these tests are acceptable to me at this point.

ssahasra:
> 
> I'm not really understanding doing hcc before hip but these tests are acceptable to me at this point.

There is no functional difference between the hcc tests for synchronous mode and the hip tests (which are only for synchronous mode). My earlier comment was meant to clarify that I am holding any further changes to the hip tests until the hcc tests are fully reviewed and accepted, at which point, I will simply port them over to hip.

@b-sumner if the hcc tests are acceptable, can we move ahead with the gerrit review for the device library?

tests/Unit/OCKL/AsyncStreams/drop-packets.cpp
ssahasra: Good catch about the C-style casts. Replaced all of them with reintepret_cast. Also reindented all the files with default settings in clang-format. I assume that the default style is LLVM, which should be right for HCC in the absence of any other coding style. But clang-format does get confused about lambda's so some of it had to be fixed by hand.

ssahasra: clang-format messed up the RUN lines in a way that allowed the tests to compile successfully but not run the resulting binary. Latest change restores the RUN lines. The tests pass with the upstream version of device-libs.

AlexVlx (approved):

Issue #917, opened 2018-11-04
Compilation is failing when using KMTHINLTO=1 together with KMDUMPISA=1


misos1:
Seems it is trying to write dump hsaco file to root. When I enable this permission then it is not failing but this is not good.

``` bash
$ KMTHINLTO=1 KMDUMPISA=1 hcc -hc main.cpp
```
``` bash
objdump: /usr/lib/x86_64-linux-gnu/libm.a: File format not recognized
cp: cannot create regular file '/dump-gfx900.hsaco': Permission denied
clang-7: error: linker command failed with exit code 1 (use -v to see invocation)
```

aaronenyeshi:
Thank you for finding this bug @misos1 . I've added the fix to HCC PR https://github.com/RadeonOpenCompute/hcc/pull/918

misos1:
Nice to know that I can change dumping dir using KMDUMPDIR. At least for files dump-gfx900.hsaco  and dump-gfx900.isa. Seems dump-gfx900.isabin is ignoring this variable and is dumping always to current dir.

Also is ThinLTO Phase 2 still under development? I can see that linking is few times faster when I have multiple files and kernels and using thinlto and more processor cores are used instead of only one without thinlto. There is stated under "ThinLTO Phase 2 - Under development" that "...which will perform in-program opt and codegen in parallel." so I am guessing phase 1 was not processing in parallel?

aaronenyeshi:
Thanks for another good catch, let me fix those ones. 

Also about ThinLTO Phase 2, I've had to halt on that development since early this year, due to other projects, and issues. It was ready for use, but wasn't upstreamable material. (nor production ready). I believe we will work on it again soon. (I also want it released, as it has significantly faster link time improvements)

aaronenyeshi:
Added the fix for .isabin files to the same PR.

aaronenyeshi:
Please confirm that it is fixed in clang_tot_upgrade. Re-open if still persists. thanks!

misos1:
Seems ok.

PR #918, opened 2018-11-05
Fix Compilation for KMDUMPISA dump dir

aaronenyeshi:
HCC is dumping the hsaco object to root directory, but should be dumping
to the default dump directory. Fix this path for KMTHINLTO=1 and
KMDUMPISA=1. Fixes compilation error reported in #917


david-salinas (approved):
LGTM

aaronenyeshi:
Thanks for verifying @david-salinas

PR #929, opened 2018-11-14
Remove unused targets from tests lit.cfg

aaronenyeshi:
These are older targets that we don't need to compile with out default tests in HCC.


scchan (approved):
LGTM

PR #930, opened 2018-11-15
Fix tracking of host memory pointer locking/unlocking

gargrahul:
Requires for fixing SWDEV-167207


gargrahul:
@scchan I see Unit/HC/wg_register_limit1.cpp is failing though I don't think changes are related to this test. Could you please check?

scchan (approved):
LGTM

PR #936, opened 2018-11-22
Fix HCC Assembler Scripts

aaronenyeshi:
There are multiple issues with hc-host-assemble, hc-kernel-assemble, and clamp-assemble scripts. Firstly, we use exit code 1 for errors, which is incorrect as clang driver does not halt to exit code 1 (only for assembler tool, not linker). Secondly, we are not checking the exit code values of the tools they call upon. Thirdly, exit code checking is not very informative.

So I've created these changes:
- Add new error-handling.in script for present and future error handling functions
- Changed exit codes to -1 for failures (since clang driver requires -1 for failure and -2 for crash)
- Added check_exit_status function for exit code checking, and it will print file and line number along with the failed status number, or kill signal used to kill a process.
- Added above checking to assembler scripts


AlexVlx (approved):

scchan (approved):
LGTM

david-salinas (approved):
LGTM

PR #937, opened 2018-11-23
Roctracer hip frontend 181113

eshcherb:
rebased #846


include/hc_hsa_op_id.h
```diff
@@ -0,0 +1,36 @@
...
+
+namespace hc {
+
+// HCC Op IDs enumeration
+enum HSAOpId {
+    HSA_OP_ID_DISPATCH = 0,
+    HSA_OP_ID_COPY = 1,
+    HSA_OP_ID_BARRIER = 2,
+    HSA_OP_ID_NUM = 3,
+    HSA_OP_ID_ANY = 4
```
scchan: Could you please provide the meaning and usage of HSA_OP_ID_ANY in a comment?  
Also, I think conceptually NUM should include ANY since it's counting how many ID's has been defined.  This makes the code quite confusing

include/hc_hsa_op_id.h
```diff
@@ -0,0 +1,36 @@
...
+};
```
scchan: Do we need a separate header just for this enum?  Can this go into activity_prof.h?

lib/hsa/activity_prof.h
```diff
@@ -0,0 +1,175 @@
...
+    typedef std::recursive_mutex mutex_t;
+
+    static void set_id_callback(const id_callback_fun_t& fun) { _table.id_callback = fun; }
+    static id_callback_fun_t get_id_callback() { return _table.id_callback; }
+    
+    static bool set_async_callback(const op_id_t& op_id, const callback_fun_t& fun, const callback_arg_t& arg) {
+        std::lock_guard<mutex_t> lck(_mutex);
+        if (op_id == hc::HSA_OP_ID_ANY) {
+            for (op_id_t i = 0; i < hc::HSA_OP_ID_NUM; ++i) {
+                set_async_callback(i, fun, arg);
```
scchan: Recursive call here is inefficient and unnecessary since fun, arg could be assigned to table

lib/hsa/activity_prof.h
```diff
@@ -0,0 +1,175 @@
...
+        callback_fun_t fun[hc::HSA_OP_ID_NUM];
+        callback_arg_t arg[hc::HSA_OP_ID_NUM];
+    };
+    typedef std::recursive_mutex mutex_t;
+
+    static void set_id_callback(const id_callback_fun_t& fun) { _table.id_callback = fun; }
+    static id_callback_fun_t get_id_callback() { return _table.id_callback; }
+    
+    static bool set_async_callback(const op_id_t& op_id, const callback_fun_t& fun, const callback_arg_t& arg) {
+        std::lock_guard<mutex_t> lck(_mutex);
```
scchan: Other than the recursive call below, it seems that `SetActivityCallback` is the only path that could reach to this function?  Is that the function where roctracer calls to setup the callbacks?
Why is a mutex needed here?  Shouldn't this be done only once for the entire application?

lib/hsa/activity_prof.h
```diff
@@ -0,0 +1,175 @@
...
+        if (op_id == hc::HSA_OP_ID_ANY) {
+            for (op_id_t i = 0; i < hc::HSA_OP_ID_NUM; ++i) {
```
scchan: why ID_ANY doesn't need a callback?

lib/hsa/activity_prof.h
```diff
@@ -0,0 +1,175 @@
...
+            _table.fun[op_id] = fun;
+            _table.arg[op_id] = arg;
+        } else {
+          return false;
+        }
+        return true;
+    }
+    
+    static void get_async_callback(const op_id_t& op_id, callback_fun_t* fun, callback_arg_t* arg) {
+        std::lock_guard<mutex_t> lck(_mutex);
```
scchan: Should avoid having a mutex here to retrieve the callback function and its arg.  I think what we need is a flag to indicate whether the roctracer is active and its infrastructure has been initialized

lib/hsa/activity_prof.h
```diff
@@ -0,0 +1,175 @@
...
+    // Domain ID
+    static const int ACTIVITY_DOMAIN_ID = ACTIVITY_DOMAIN_HCC_OPS;
+    // Timeer type
+    typedef hsa_rt_utils::Timer timer_t;
+
+    ActivityProf(const op_id_t& op_id, const OpCoord& op_coord) :
+        _op_id(op_id),
+        _op_coord(op_coord),
+        _callback_fun(NULL),
+        _callback_arg(NULL),
```
scchan: replace `NULL` with `nullptr`

lib/hsa/activity_prof.h
```diff
@@ -0,0 +1,175 @@
...
+        _record_id(0)
+    {}
+
+    // Initialization
+    void initialize() {
+        CallbacksTable::get_async_callback(_op_id, &_callback_fun, &_callback_arg);
+        if (_callback_fun != NULL) {
+            if (_timer == NULL) _timer = new timer_t;
```
scchan: There is a race condition here since multiple threads may try to create the same timer.  I think there could be a memory leak because it's not obvious to me where this object gets deleted.  Changing the timer into a statically allocated object would solve these 2 issues.

lib/hsa/activity_prof.h
```diff
@@ -0,0 +1,175 @@
...
+        _op_coord(op_coord),
+        _callback_fun(NULL),
+        _callback_arg(NULL),
+        _record_id(0)
+    {}
+
+    // Initialization
+    void initialize() {
+        CallbacksTable::get_async_callback(_op_id, &_callback_fun, &_callback_arg);
+        if (_callback_fun != NULL) {
```
scchan: In what circumstances`_callback_fun` would get a `nullptr`?  Does it only occur when roctracer is not in use?  If that's the case, having a flag to indicate whether roctracer is active would be useful (see previous comments) rather than checking if the callback function ptr is null

lib/hsa/activity_prof.h
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    private:
+    static table_t _table;
+    static mutex_t _mutex;
+};
+
+// Activity profile class
+template <typename OpCoord>
+class ActivityProf {
```
scchan, line 91: There's only one kind of ActivityProf, does it need to be a template?

lib/hsa/activity_prof.h
```diff
@@ -0,0 +1,175 @@
...
+        CallbacksTable::table_t CallbacksTable::_table{};                            \
+        CallbacksTable::mutex_t CallbacksTable::_mutex;                              \
+        template<> HSAOp::ActivityProf::timer_t* HSAOp::ActivityProf::_timer = NULL; \
+        template<> std::atomic<record_id_t> HSAOp::ActivityProf::_glob_record_id(0); \
+    } // activity_prof
+
+namespace activity_prof {
+typedef uint64_t record_id_t;
+typedef uint32_t op_id_t;
+typedef uint32_t activity_kind_t;
```
scchan: Are these `typedef`s common across all language runtimes?  If that's the case, should that go into `prof_protocol.h`?

lib/hsa/activity_prof.h
```diff
@@ -0,0 +1,175 @@
...
+        CallbacksTable::get_async_callback(_op_id, &_callback_fun, &_callback_arg);
+        if (_callback_fun != NULL) {
+            if (_timer == NULL) _timer = new timer_t;
+            _record_id = _glob_record_id.fetch_add(1, std::memory_order_relaxed);
+            (CallbacksTable::get_id_callback())(_record_id);
+        }
+    }
+
+    // Activity callback routine
+    inline void callback(const command_id_t& command_id, const uint64_t& begin_ts, const uint64_t& end_ts,
```
scchan: remove `inline` since it's already implicit

lib/hsa/activity_prof.h
```diff
@@ -0,0 +1,175 @@
...
+#include "prof_protocol.h"
+#include "hsa_rt_utils.hpp"
+
+#define ACTIVITY_PROF_INSTANCES()                                                    \
+    namespace activity_prof {                                                        \
+        CallbacksTable::table_t CallbacksTable::_table{};                            \
+        CallbacksTable::mutex_t CallbacksTable::_mutex;                              \
+        template<> HSAOp::ActivityProf::timer_t* HSAOp::ActivityProf::_timer = NULL; \
+        template<> std::atomic<record_id_t> HSAOp::ActivityProf::_glob_record_id(0); \
+    } // activity_prof
```
scchan, line 39: Could you change this to avoid using a macro?  It makes the code that uses this macro (`mcwamp_hsa.cpp`) looks very confusing (a function call at the file scope) and it really hurts readability

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -708,7 +713,9 @@ struct HSAOpCoord
 // Base class for the other HSA ops:
 class HSAOp : public Kalmar::KalmarAsyncOp {
 public:
-    HSAOp(Kalmar::KalmarQueue *queue, hc::hcCommandKind commandKind) ;
+    typedef activity_prof::ActivityProf<HSAOpCoord> ActivityProf;
```
scchan: prefer not creating a `typedef` for `ActivityProf`

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -5007,6 +5018,9 @@ HSABarrier::getEndTimestamp() override {
 }
 
 
+// Activity profiling instantiation
+ACTIVITY_PROF_INSTANCES();
```
scchan: Should replace usage of this macro, see previous comment on this

scchan (changes requested):

include/hc_hsa_op_id.h
eshcherb: HSA_OP_ID_ANY is used when you want to initialize all callbacks with same function:
        if (op_id == hc::HSA_OP_ID_ANY) {
            for (op_id_t i = 0; i < hc::HSA_OP_ID_NUM; ++i) {
                set_async_callback(i, fun, arg);

How you would like to change it? - I can use just HSA_OP_ID_NUM for this purpose.

include/hc_hsa_op_id.h
eshcherb: This is a public API which used to set callbacks for a specific OP and 'activity_prof.h' is internal implementation header.

lib/hsa/activity_prof.h
eshcherb: Actually it is not performance critical function, I will change it.

lib/hsa/activity_prof.h
eshcherb: I saw in TensorFlow that the tracing was turning only for some periods, steps.
The API is CUPTI like and CUPTI let you arbitrary turn tracing on/off.

eshcherb: But set is not performance affecting functions so the mutex here is fine.

eshcherb: And I can use lock free barrier mechanism for get()

lib/hsa/activity_prof.h
eshcherb: HSA_OP_ID_ANY means enabling tracing of all OPs

lib/hsa/activity_prof.h
eshcherb: And I probably should have them in prof_protocol.h already, I will fix it

lib/hsa/activity_prof.h
eshcherb: sure

lib/hsa/activity_prof.h
eshcherb: The macro is used to instantiate static fields.
I tried to incapsulate all implementation in the header.
I see two solutions:
a) Actually the header is included only in one file mcwamp_hsa.cpp so I can just put the instantiation directly in the header.

b) Also I can do the instantiation explicitly in  mcwamp_hsa.cpp under USE_PROF_API macro.

Which solution do you prefer or maybe you have another proposal?

lib/hsa/mcwamp_hsa.cpp
eshcherb: I see two solutions I wrote in the previous comment on this.

lib/hsa/activity_prof.h
eshcherb: Yes, I agree, I will add a flag and a barrier for lock-free access mechanism.

lib/hsa/activity_prof.h
eshcherb: Timer cannot be statically allocated:
 - It has non-trivial constructor and the moment when he constructor will be called is undefined
 - More over the constructor queries HSA system frequency and cannot be called before HSA runtime is initialize with hsa_init()

lib/hsa/activity_prof.h
eshcherb: sure

lib/hsa/activity_prof.h
eshcherb: The problem that you can enable tracing of specific OPs

lib/hsa/activity_prof.h
eshcherb: The OpCoord struct is defined in mcwamp_hsa.cpp

lib/hsa/mcwamp_hsa.cpp
eshcherb: Because ActivityProf is template I'm accessing it from HSAOp class for static fields instantiating like the following:
template<> std::atomic<record_id_t> HSAOp::ActivityProf::_glob_record_id(0);

I can not pass op-coord and pass device_id and queue_id explicitly so I will be able to remove the template parameter and it will resolve the concerns about the template.

eshcherb: I mean instead of passing op_coord reference I can pass explicitly device_id and queue_id

lib/hsa/activity_prof.h
eshcherb: I think the best solution to make Timer singleton class

include/hc_hsa_op_id.h
scchan: I'd rather not having an ID for this special case.  Could roctracer just makes multiple calls set_async_callback to do that?  If this is important to do that in a single function, I'd suggest having another function to do that rather than inventing an ID for that purpose.

include/hc_hsa_op_id.h
scchan: How about `hc_prof_runtime.h`, is that part of the public API?

lib/hsa/activity_prof.h
```diff
@@ -0,0 +1,179 @@
...
+typedef void* callback_arg_t;
+
+class CallbacksTable {
+    public:
+    struct table_t {
+        id_callback_fun_t id_callback;
+        callback_fun_t fun[hc::HSA_OP_ID_NUM];
+        callback_arg_t arg[hc::HSA_OP_ID_NUM];
+    };
+    typedef std::recursive_mutex mutex_t;
```
scchan: Don't think the recursive_mutex is needed anymore after the recursive call in set_async_callback has been removed, could you change this into a std::mutex?

include/hc_hsa_op_id.h
eshcherb: Yes, hc_prof_runtime.h is public API but as I understand it is generic and the HSA_OP is specific to HSA implementation.
I can put the enum to hc_prof_runtime.h

lib/hsa/activity_prof.h
eshcherb: yes, will fix it, thanks

lib/hsa/activity_prof.h
scchan: > I saw in TensorFlow that the tracing was turning only for some periods, steps.
> The API is CUPTI like and CUPTI let you arbitrary turn tracing on/off.

Is that on a per thread basis or per process?

lib/hsa/activity_prof.h
scchan: b) seems fine

lib/hsa/activity_prof.h
scchan: Essentially `fun` and `arg` are the same for any kind of `op_id_t`?

lib/hsa/activity_prof.h
scchan: I still prefer having an `enabled` flag.  It could be one per OP.  It makes the code more explicit about what it's doing.
The callback function is always the same for a given during the entire lifetime of the application?

lib/hsa/activity_prof.h
eshcherb: You can set a specific callback for a specific OP

lib/hsa/activity_prof.h
eshcherb: I like (a) more because everything in one header but can change for (b) if you want me

lib/hsa/activity_prof.h
eshcherb: It's for runtime library instance as CUPTI record data has thread id and process id

eshcherb: Activity tracing can be enabled and disabled in any moment and it is enabled for runtime library instance as CUPTI record data has thread id and process id.

eshcherb: Will add the flag.

lib/hsa/activity_prof.h
scchan: a) is fine

lib/hsa/activity_prof.h
```diff
@@ -0,0 +1,180 @@
...
+
+#include "kalmar_runtime.h"
+
+#if USE_PROF_API
+#include "prof_protocol.h"
+#include "hsa_rt_utils.hpp"
+
+#define ACTIVITY_PROF_INSTANCES()                                                    \
+    hsa_rt_utils::Timer* hsa_rt_utils::TimerFactory::instance_ = nullptr;            \
+    hsa_rt_utils::TimerFactory::mutex_t hsa_rt_utils::TimerFactory::mutex_;          \
```
scchan, line 35: do we still need `instance_` and `mutex_` here?  I don't see any reference to them?

eshcherb: It is TimerFactory timer instance and creation mutex.

eshcherb: It is used inside TimerFactory implementation to create a single timer instance.

scchan (approved):
LGTM

eshcherb:
Thanks,


Evgeny

________________________________

scchan:
@eshcherb thanks for working on profiler integration

PR #939, opened 2018-11-23
link to irif.amdgcn.bc to fix linking error related to __llvm_fence_*

scchan:


aaronenyeshi:
I think @b-sumner wanted us to remove irif from being used, but I'm not sure if that's still the case

scchan:
@aaronenyeshi That doesn't address the issue that HIP is currently broken without linking to irif since those functions are being used in HIP __atomic_work_item_fence implementation.   Not sure what changed causing this regression but if someone wants to deprecate irif then he/she should also fix all the components that currently depend on it.

aaronenyeshi:
HIP's __atomic_work_item_fence implementation was moved inside of ROCDL. and that calls the irif library internally.

I'm not sure why @mangupta moved the __atomic_work_item_fence back into HIP. https://github.com/ROCm-Developer-Tools/HIP/commit/0703a2d0f0781885d38af97c97880b494f2c872d I removed it completely from HIP before when I deprecated.

scchan:
ok, I see what the problem is now.  I'm trying to build the 1.9.2 HIP with the latest HCC

mangupta:
@aaronenyeshi If you observe, when I added back the ```__atomic_work_item_fence```, I guarded it against hcc 1.3 or higher. This is because we want HIP to be build-able with hcc from ROCm 1.9.x.

So the following combinations should work correctly:
* hip from 1.9.x branch + hcc from 1.9.x branch (hip will use its ```__atomic_work_item_fence``` implementation)
* hip from master branch + hcc clang_tot_upgrade branch (hip will use rocdl for the functionality) 
* hip from master branch + hcc from 1.9.x branch (hip will use its ```__atomic_work_item_fence``` implementation)
* hip from master branch + hcc from 2.0.x branch (hip will use rocdl for the functionality)

Any other combination will probably not work.

aaronenyeshi (changes requested):
Should we remove this PR? And make a new PR to remove 901 there, and refactoring those symlinks

PR #946, opened 2018-11-26
Improve the exit code output to show file and line number

aaronenyeshi:
Add a new function to handle the exit codes, and give more information as to exit location.


scchan (commented):
improving clamp-assemble seems moot since it won't get used after https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/165
are there other scripts that could be improved by exit_with_code()?

aaronenyeshi:
Oh right, I had made the changes locally to all the assemble scripts, but some reason only added one with `git add`. Let me force update this commit

aaronenyeshi:
Please check the updated commit, which includes the hc-kernel-assemble and hc-host-assemble

scchan (approved):
LGTM

PR #949, opened 2018-11-28
[DO NOT MERGE] Just for CI

AlexVlx:


scchan:
I don't think we need this anymore

PR #952, opened 2018-11-29
replace __hc math functions with __ocml functions

david-salinas:
remove the forward declarations of the __hc_ math functions and call the ocml functions (in rocdl) directly.

### DIFF
--- a/include/kalmar_math.h
+++ b/include/kalmar_math.h
@@ -10,286 +10,285 @@
 #include <cmath>
 #include <stdexcept>
 
-extern "C" __fp16 __hc_acos_half(__fp16 x) restrict(amp);
-extern "C" float __hc_acos(float x) restrict(amp);
-extern "C" double __hc_acos_double(double x) restrict(amp);
+extern "C" _Float16 __ocml_acos_f16(_Float16 x) [[hc]];
+extern "C" float __ocml_acos_f32(float x) [[hc]];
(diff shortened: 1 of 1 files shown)


aaronenyeshi:
Thanks David, should we address these comments from https://github.com/RadeonOpenCompute/hcc/pull/865 in another PR? 
```
Instead of type __fp16, please use _Float16.
Some unit tests needed for these _Float16 math functions.
```

david-salinas:
> Thanks David, should we address these comments from #865 in another PR?
> 
> ```
> Instead of type __fp16, please use _Float16.
> Some unit tests needed for these _Float16 math functions.
> ```

I'll make the change to _Float16, that does seem to be what clang recommends.  And adding the type to the tests, will come in an update.  I'll see if I can add it to this PR, otherwise I'll create a separate PR.

aaronenyeshi (changes requested):
Thanks @david-salinas , I'll re-review after you add replace __fp16 with _Float16.

david-salinas:
So, I've changed __fp16 to _Float16.  But I haven't added changes to existing AmpMath tests to test for _Float16 types.  Because we need to add a stdlib that supports the half float type, and right now the stdc++ we use does not.  I'm opening a ticket for myself (SWDEV-173564) to add half float testing to our HCC unit tests.  In the meantime we might as well let these cosmetic changes go in.

aaronenyeshi (approved):
LGTM

PR #957, opened 2018-12-05
add centOS docker builds for hcc

david-salinas:
With growing interest in CentOS, adding CentOS builds to regular HCC build/tests.  This change adds a build parameter to the HCC build jenkins file.  For now the default is ubuntu 16.04.  But we may add randomization to this parameter in the near future.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,7 +9,8 @@ properties([buildDiscarder(logRotator(
     numToKeepStr: '10')),
     disableConcurrentBuilds(),
     parameters([booleanParam( name: 'run_hip_integration_testing', defaultValue: false, description: 'Build hip with this compiler and run hip unit tests' ),
-                string( name: 'hip_integration_branch', defaultValue: 'ROCm-Developer-Tools/HIP/master', description: 'Path to hip branch to build & test' )]),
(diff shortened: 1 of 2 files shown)


aaronenyeshi (approved):
Thanks David, LGTM. Also would it be better to build and test both environments on each jenkins job?

david-salinas:
sigh ... re the check failure above ... there is a bug in Jenkins that doesn't let us use the proper syntax for the "choices" parameter type:  https://issues.jenkins-ci.org/browse/JENKINS-38995  I need to update the Jenkinsfile.

david-salinas:
@aaronenyeshi I've thought about doing a double build too.  I dunno, with our current resources, I think it may slow things down too much.  Though, there is a way to do Jenkins jobs in parallel - the openCL PSDB job runs the OpenCL conformance tests on separate GPUs, in parallel.

pfultz2:
I think it would be better to run both. Also, we use `parallel` directives on our builds to run them both across several machines at the same time. Also, do you guys no longer use the rocmtest nodes?

david-salinas:
> I think it would be better to run both. Also, we use `parallel` directives on our builds to run them both across several machines at the same time. Also, do you guys no longer use the rocmtest nodes?

I agree, we should do both ubuntu and centos.  I'm going to add parallel steps, but it will likely be in another PR.  And, we're not currently using rocmtest nodes, for hcc build/test.  If we can use them, that would help alleviate our resource issues.

PR #962, opened 2018-12-10
add attr address_space(5) aka private to ocml math function fwd decls…

david-salinas:
… and calls

Missed these in the last update.  The OCML math function pointer arguments need to be in the private address space.  So we need the forward declarations to match what is actually in the ROCr Device Libs OCML library.  As well, for our wrapped functions (kalmar::fast_math:: and kalmar::precise::) we need to explicit address space cast arguments when calling the ocml functions.

### DIFF
--- a/include/kalmar_math.h
+++ b/include/kalmar_math.h
@@ -486,10 +486,18 @@ namespace Kalmar
         float fmod(float x, float y) { return fast_math::fmodf(x, y); }
 
         HCC_MATH_LIB_FN
-        float frexpf(float x, int *exp) { return __ocml_frexp_f32(x, (__attribute__((address_space(5))) int*)exp); }
+        float frexpf(float x, int *exp) {
+        	int e;
+        	float ret = __ocml_frexp_f32(x, (__attribute__((address_space(5))) int*) &e);
+        	*exp = e; return ret;
+        }
 
         HCC_MATH_LIB_FN
-        _Float16 frexp(_Float16 x, int *exp) { return __ocml_frexp_f16(x, (__attribute__((address_space(5))) int*) exp); }
+        _Float16 frexp(_Float16 x, int *exp) {
+        	int e;
+        	_Float16 ret = __ocml_frexp_f16(x, (__attribute__((address_space(5))) int*) &e);
+            *exp = e; return ret;
+        }
 
         HCC_MATH_LIB_FN
         float frexp(float x, int *exp) { return fast_math::frexpf(x, exp); }
@@ -568,11 +576,16 @@ namespace Kalmar
         float log2(float x) { return fast_math::log2f(x); }
 
         HCC_MATH_LIB_FN
-        float modff(float x, float *iptr) { return __ocml_modf_f32(x, (__attribute__((address_space(5))) float*)iptr); }
+        float modff(float x, float *iptr) {
+        	float i;  float ret = __ocml_modf_f32(x, (__attribute__((address_space(5))) float*)&i);
+        	*iptr = i; return ret;
+        }
 
         HCC_MATH_LIB_FN
-        _Float16 modf(_Float16 x, _Float16 *iptr) { return __ocml_modf_f16(x, (__attribute__((address_space(5))) _Float16*)iptr); }
-
+        _Float16 modf(_Float16 x, _Float16 *iptr) {
+        	_Float16 i; _Float16 ret = __ocml_modf_f16(x, (__attribute__((address_space(5))) _Float16*) &i);
+        	*iptr = i; return ret;
+        }
 
         HCC_MATH_LIB_FN
         float modf(float x, float *iptr) { return fast_math::modff(x, iptr); }
@@ -623,12 +636,18 @@ namespace Kalmar
         float sin(float x) { return fast_math::sinf(x); }
 
         HCC_MATH_LIB_FN
-        void sincosf(float x, float *s, float *c) { *s = __ocml_sincos_f32(x, (__attribute__((address_space(5))) float*)c); }
+        void sincosf(float x, float *s, float *c) {
+        	float lc;
+        	*s = __ocml_sincos_f32(x, (__attribute__((address_space(5))) float*)&lc);
+        	*c=lc;
+        }
 
         HCC_MATH_LIB_FN
         void sincos(_Float16 x, _Float16 *s, _Float16 *c)
         {
-            *s = __ocml_sincos_f16(x, (__attribute__((address_space(5))) _Float16*) c);
+        	_Float16 lc;
+            *s = __ocml_sincos_f16(x, (__attribute__((address_space(5))) _Float16*) &lc);
+            *c = lc;
         }
 
         HCC_MATH_LIB_FN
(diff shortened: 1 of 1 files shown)


include/kalmar_math.h
```diff
@@ -568,10 +568,10 @@ namespace Kalmar
         float log2(float x) { return fast_math::log2f(x); }
 
         HCC_MATH_LIB_FN
-        float modff(float x, float *iptr) { return __ocml_modf_f32(x, iptr); }
+        float modff(float x, float *iptr) { return __ocml_modf_f32(x, (__attribute__((address_space(5))) float*)iptr); }
 
         HCC_MATH_LIB_FN
-        _Float16 modf(_Float16 x, _Float16 *iptr) { return __ocml_modf_f16(x, iptr); }
+        _Float16 modf(_Float16 x, _Float16 *iptr) { return __ocml_modf_f16(x, (__attribute__((address_space(5))) _Float16*)iptr); }
```
aaronenyeshi: Just curious, why is there a second signature `_Float16 modf(_Float16 x, _Float16* p)` in line 1316?

include/kalmar_math.h
```diff
@@ -568,10 +568,10 @@ namespace Kalmar
         float log2(float x) { return fast_math::log2f(x); }
 
         HCC_MATH_LIB_FN
-        float modff(float x, float *iptr) { return __ocml_modf_f32(x, iptr); }
+        float modff(float x, float *iptr) { return __ocml_modf_f32(x, (__attribute__((address_space(5))) float*)iptr); }
```
aaronenyeshi: Second definition in line 1313.

include/kalmar_math.h
```diff
@@ -486,10 +486,10 @@ namespace Kalmar
         float fmod(float x, float y) { return fast_math::fmodf(x, y); }
 
         HCC_MATH_LIB_FN
-        float frexpf(float x, int *exp) { return __ocml_frexp_f32(x, exp); }
+        float frexpf(float x, int *exp) { return __ocml_frexp_f32(x, (__attribute__((address_space(5))) int*)exp); }
 
         HCC_MATH_LIB_FN
-        _Float16 frexp(_Float16 x, int *exp) { return __ocml_frexp_f16(x, exp); }
+        _Float16 frexp(_Float16 x, int *exp) { return __ocml_frexp_f16(x, (__attribute__((address_space(5))) int*) exp); }
```
aaronenyeshi: second definition in line 1160.

include/kalmar_math.h
```diff
@@ -486,10 +486,10 @@ namespace Kalmar
         float fmod(float x, float y) { return fast_math::fmodf(x, y); }
 
         HCC_MATH_LIB_FN
-        float frexpf(float x, int *exp) { return __ocml_frexp_f32(x, exp); }
+        float frexpf(float x, int *exp) { return __ocml_frexp_f32(x, (__attribute__((address_space(5))) int*)exp); }
```
aaronenyeshi: second definition in line 1163

aaronenyeshi (changes requested):
Please see my comments. Also, do we not provide ocml functions fract, lgamma_r, and sincospi? I know HIP uses sincospi. + @b-sumner

include/kalmar_math.h
david-salinas: the one at 1163 is in the kalmar::precise_math namespace.  We also have a kalmar::fast_math namespace.  I don't know what the original requirements were for creating these namespaces, but they're there.  Originally, both would just call the __hc math functions.  Now, we're calling the ocml functions directly, but I left the fast_math and precise_math namespaces.  Maybe we don't need them anymore?  But I don't know if there are any dependencies on them in ROCm.

include/kalmar_math.h
david-salinas: ... different namespace.

include/kalmar_math.h
AlexVlx: They are there by design, do not remove. Thanks.

include/kalmar_math.h
david-salinas: .. different namespace.

include/kalmar_math.h
david-salinas: .. different namespace.

include/kalmar_math.h
david-salinas: ok.  Thanks Alex.

aaronenyeshi: Okay lets keep it as it is. Do we need ocml functions fract, lgamma_r, and sincospi?

david-salinas: yeah, it seems we've missed these math functions.  There may be others; I'd have to go through all of the ocml math functions to check.  Unless someone knows of a reason why we excluded these?

aaronenyeshi: I have a feeling those missing math functions were added to ocml after Kalmar worked on these in HCC. Similar to bitselect, bitinsert, and bitextract which I added in Aug.

include/kalmar_math.h
b-sumner: You can't cast a generic pointer to a private pointer and expect it to work unless the data being pointed at was allocated in private memory.  To handle all cases the implementation should look something like:  { int e; _Float16 ret = __ocml_frexp_f16(x, (__attribute__((address_space(5))) int*)&e); *exp = e; return ret; }
The compiler will usually clean up the extra load/store.

david-salinas: Thanks Brian.  I think we'd need to do this for all of the functions with pass-by-reference args like this.

include/kalmar_math.h
david-salinas: Thanks Aaron.  I'll add these.  I need to correct the math functions with pass-by-ref args as well.  See comment above.

david-salinas: ok ... correction to my above comment.  the bit* functions you mentioned above are builtins - which is a bit lower than these math RT functions, which may call the builtins.  As for fract, lgamma_r, and sincospi; we could add these, but there are a bunch of other OCML functions we're also missing.  Does anyone have objections to adding more than just those 3?

david-salinas: for the record, this is the list of OCML math functions not wrapped in this header:
acospi
add
asinpi
atanpi
atan2pi
atanpired
atanred
erfcx
fmuladd
fract
len3
len4
lgamma_r
mad
max
maxmag
min
minmag
mul
ncdf
ncdfinv
pown
powr
rhypot
rint
rootn
sincospi
sincospired
sub
tanpired
tanred
tgamma

aaronenyeshi: Thanks @david-salinas, that's a lot. Please take your time. also I do not disagree with adding these.

david-salinas: @AlexVlx @b-sumner Do either of you know why we seem to avoid double (f64) floats when we wrap the ocml math functions for HC users?  Is f64 support on accelerators frowned upon?  Should I add them?

b-sumner: My answers are: No idea.  No.  Yes.

AlexVlx: No you should not. By design fast_math does not expose doubles, only floats (and FP16). No, you should not add doubles into that namespace. No you should not add all functions in ROCDL, only if you identify some function that is in the {C++ standard library, C++AMP} union but is missing here should we add them. If both of you look, precise_math has lots and lots of doubles. This is intentional and not an accident.

david-salinas: sorry, I was just looking at fast_math, which rightly shouldn't have doubles.  As far as I can tell, most of the functions I listed are in amp_math.h and/or C++ stdlib.  But, I can ... Double ... check.  ;-)

AlexVlx: I think that we should be matching C++11 / C++AMP (without some functions in AMP 1.2, which were... special). It was the case way back when I checked / touched these. Having said that, we may have some extra barnacles due to HIP, which we may be able to clean now that HIP moved on to do the right thing i.e. consume ROCDL directly.

david-salinas: ok, I may not be looking at the official C++AMP amp_math.h, but as far as I can tell, from my list above only, rint and tgamma are in amp_math.h and C++ stdlib, and are missing here.

AlexVlx: `tgamma` not in fast_math. fast_math != precise_math in terms of the functions that are exposed. Do not look at headers that were perhaps hacked for reasons lost to history, look at official specs: http://download.microsoft.com/download/2/2/9/22972859-15C2-4D96-97AE-93344241D56C/CppAMPOpenSpecificationV12.pdf. `rint` is more subtle since it needs to access the current rounding mode to meet the spec: https://en.cppreference.com/w/cpp/numeric/math/rint, which is possibly problematic. It was intentionally omitted from AMP, but we may want to consider adding it in the future.

david-salinas: ok.  So we really don't need to add any then.  Looks like we're covered.

include/kalmar_math.h
david-salinas: I've pushed an updated commit that should resolve this.

include/kalmar_math.h
david-salinas: updated commit pushed

david-salinas:
@b-sumner @AlexVlx @aaronenyeshi   any objections to the latest update to this PR?

b-sumner:
No objection from me.

Indrashish:
Issue confirmed to resolve. Current HCC able to build successfully. Closing the ticket

PR #973, opened 2018-12-20
Fixes for solving some application failure in dependent multiple threads usi…

qianfengz:
The patch has two fixing points to solve issues occurred with multiple threads and multiple streams application. 

    1) The first issue occurs when Thread B calls hipStreamSynchronize() on stream A which just en-queued a Copy command and will en-queue a kernel by Thread A.

    The fix method here is to postpone the calling of "overridAcquireFenceIfNeeded" to a late point

    2) The second issue occurs when Thread B calls hipEventSynchronize() to wait on the completion of a Copy en-queued by Thread A on stream A, and the Event is created by hipEventDefaults, rather than hipEventReleaseToSystem.

    The fix method here is by absorbing the System Scope fencing request in HSABarrier::enqueueAsync().

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -4293,9 +4293,6 @@ HSAQueue::dispatch_hsa_kernel(const hsa_kernel_dispatch_packet_t *aql,
     Kalmar::HSADevice* device = static_cast<Kalmar::HSADevice*>(this->getDev());
 
     std::shared_ptr<HSADispatch> sp_dispatch = std::make_shared<HSADispatch>(device, this/*queue*/, nullptr, aql);
-    if (HCC_OPT_FLUSH) {
-        sp_dispatch->overrideAcquireFenceIfNeeded();
-    }
 
     HSADispatch *dispatch = sp_dispatch.get();
     waitForStreamDeps(dispatch);
@@ -4602,6 +4599,8 @@ HSADispatch::dispatchKernelWaitComplete() {
         // extract hsa_queue_t from HSAQueue
         hsa_queue_t* rocrQueue = hsaQueue()->acquireLockedRocrQueue();
 
+        overrideAcquireFenceIfNeeded();
+
         // dispatch kernel
         status = dispatchKernel(rocrQueue, arg_vec.data(), arg_vec.size(), true);
         STATUS_CHECK(status, __LINE__);
@@ -4643,6 +4642,8 @@ HSADispatch::dispatchKernelAsync(const void *hostKernarg, int hostKernargSize, b
     {
         // extract hsa_queue_t from HSAQueue
         hsa_queue_t* rocrQueue = hsaQueue()->acquireLockedRocrQueue();
+
+        overrideAcquireFenceIfNeeded();
 
         // dispatch kernel
         status = dispatchKernel(rocrQueue, hostKernarg, hostKernargSize, allocSignal);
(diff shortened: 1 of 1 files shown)


aaronenyeshi (approved):
LGTM, thanks!

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4602,6 +4599,8 @@ HSADispatch::dispatchKernelWaitComplete() {
         // extract hsa_queue_t from HSAQueue
         hsa_queue_t* rocrQueue = hsaQueue()->acquireLockedRocrQueue();
 
+        overrideAcquireFenceIfNeeded();
```
scchan, line 4602: should be guarded by `HCC_OPT_FLUSH`

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4644,6 +4643,8 @@ HSADispatch::dispatchKernelAsync(const void *hostKernarg, int hostKernargSize, b
...
```
scchan, line 4646: should be guarded by `HCC_OPT_FLUSH`

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4896,11 +4896,27 @@ HSABarrier::waitComplete() {
...
-        hsaQueue()->setNextSyncNeedsSysRelease(false);
+    hsa_queue_t* rocrQueue = hsaQueue()->acquireLockedRocrQueue();
+
+    if ( hsaQueue()->nextSyncNeedsSysRelease() ) {
+         toDoReleaseFenceScope = hc::system_scope;
+         DBOUTL( DB_CMD2, "absorb System Scope Release Fencing request from the HSAQueue") ;
+    };
+
+    if ( toDoReleaseFenceScope == hc::system_scope )
+         hsaQueue()->setNextSyncNeedsSysRelease(false);
```
scchan, line 4910: Should be moved into the `if` branch above since it only needs to be set to `false` when the original value is `true`

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4896,11 +4896,27 @@ HSABarrier::waitComplete() {
...
+
+    if ( hsaQueue()->nextKernelNeedsSysAcquire() ) {
+         toDoAcquireFenceScope = hc::system_scope;
+         DBOUTL( DB_CMD2, "absorb System Scope Acquire Fencing request from the HSAQueue") ;
     };
 
+    if ( toDoAcquireFenceScope == hc::system_scope )
+         hsaQueue()->setNextKernelNeedsSysAcquire(false);
```
scchan, line 4918: should be moved to inside the `if` branch above

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4995,6 +5011,8 @@ HSABarrier::enqueueAsync(hc::memory_scope fenceScope) {
         hsaQueue()->releaseLockedRocrQueue();
```
scchan, line 5011: Since you lock the rocr queue earlier in this function, could you remove the locking/unlocking of the queue within this scope?

scchan (changes requested):
Could you split this into 2 separate PRs, one for moving the `overrideAcquireFenceIfNeeded` and the other one for optimizing the memory scope?

scchan:
This PR has been superseded by 
https://github.com/RadeonOpenCompute/hcc/pull/991
https://github.com/RadeonOpenCompute/hcc/pull/992

PR #978, opened 2019-01-03
Update README with Printf Usage

aaronenyeshi:
Users had trouble enabling the printf usage with HCC. This small guide helps them use the current printf implementation.


david-salinas (approved):
LGTM

PR #989, opened 2019-01-16
add the roctracer-dev package dependency

scchan:
Adding the roctracer-dev package dependency to the hcc package


eshcherb:
Looks good to me.

PR #994, opened 2019-01-18
PR for auto_merge_201901180723_EST

rocm-hcc:


david-salinas:
The Check fails, because there is now a HC Unit test that passes (wg_register_limit1.cpp).  I've created another PR to make this test as expecting to pass now (PR #995 ).

david-salinas (approved):
even though the check "fails", it is actually because a HC unit test is actually passing now; because of an upstream fix.

PR #998, opened 2019-01-22
Flush a fraction of the queue when we are out of space for a new kernel

jeffdaily:
This PR begins with the original queue flushing code, fixes bugs, and improves its performance.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -1620,6 +1620,8 @@ public:
         for (int i = lastWaitOp-1; i >= 0;  i--) {
             if (asyncOps[i] != nullptr) {
                 auto asyncOp = asyncOps[i];
+                // we only drain starting at first found marker
+                if (drainingQueue_ && asyncOp->getCommandKind() != hcCommandMarker) continue;
 #if !defined(NDEBUG)
                 if (!foundFirstValidOp) {
                     hsa_signal_t sig =  *(static_cast <hsa_signal_t*> (asyncOp->getNativeHandle()));
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -137,6 +137,10 @@ int HCC_PROFILE_VERBOSE=0x1F;
 
 char * HCC_PROFILE_FILE=nullptr;
 
+int HCC_QUEUE_FLUSHING_RATIO=50;
+
+double QUEUE_FLUSHING_FRAC;
+
 // Profiler:
 // Use str::stream so output is atomic wrt other threads:
 #define LOG_PROFILE(op, start, end, type, tag, msg) \
@@ -1451,7 +1455,6 @@ public:
         youngestCommandKind = op->getCommandKind();
         asyncOps.push_back(std::move(op));
 
-        drainingQueue_ = false;
 
         if (DBFLAG(DB_QUEUE)) {
             printAsyncOps(std::cerr);
@@ -1608,8 +1611,13 @@ public:
 
 
         bool foundFirstValidOp = false;
+        int oldAysncOpsSize = asyncOps.size();
+        int lastWaitOp = oldAysncOpsSize - 1;
+        if (drainingQueue_) {
+            lastWaitOp = (oldAysncOpsSize * QUEUE_FLUSHING_FRAC) - 1;
+        }
 
-        for (int i = asyncOps.size()-1; i >= 0;  i--) {
+        for (int i = lastWaitOp; i >= 0;  i--) {
             if (asyncOps[i] != nullptr) {
                 auto asyncOp = asyncOps[i];
                 if (!foundFirstValidOp) {
@@ -1625,7 +1633,16 @@ public:
             }
         }
         // clear async operations table
-        asyncOps.clear();
+        if (drainingQueue_) {
+            if (oldAysncOpsSize == asyncOps.size()) {
+                asyncOps.erase(asyncOps.begin(), asyncOps.begin() + lastWaitOp);
+            }
+        }
+        else {
+            asyncOps.clear();
+        }
+
+        drainingQueue_ = false;
    }
 
     void LaunchKernel(void *ker, size_t nr_dim, size_t *global, size_t *local) override {
@@ -3779,6 +3796,15 @@ void HSAContext::ReadHccEnv()
 
     // Enable printf support
     GET_ENV_INT (HCC_ENABLE_PRINTF, "Enable hc::printf");
+
+    GET_ENV_INT (HCC_QUEUE_FLUSHING_RATIO, "Percentage of HCC's queue to be flushed when the space to dispatch a new kernel is not sufficient.  The percentage has to be greater than zero.  Any invalid value will be set to the default value.  Default=50");
+
+    if (HCC_QUEUE_FLUSHING_RATIO > 0 && HCC_QUEUE_FLUSHING_RATIO <= 100) {
+        QUEUE_FLUSHING_FRAC = HCC_QUEUE_FLUSHING_RATIO / 100.0;
+    }
+    else {
+        QUEUE_FLUSHING_FRAC = 0.5;
+    }
 
     GET_ENV_INT    (HCC_PROFILE,         "Enable HCC kernel and data profiling.  1=summary, 2=trace");
     GET_ENV_INT    (HCC_PROFILE_VERBOSE, "Bitmark to control profile verbosity and format. 0x1=default, 0x2=show begin/end, 0x4=show barrier");
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -1391,6 +1391,7 @@ public:
     //
     void printAsyncOps(std::ostream &s = std::cerr)
     {
+        std::lock_guard<std::recursive_mutex> lg(qmutex);
         hsa_signal_value_t oldv=0;
         s << *this << " : " << asyncOps.size() << " op entries\n";
         for (int i=0; i<asyncOps.size(); i++) {
@@ -1431,6 +1432,8 @@ public:
     // Save the command and type
     // TODO - can convert to reference?
     void pushAsyncOp(std::shared_ptr<HSAOp> op) {
+
+        std::lock_guard<std::recursive_mutex> lg(qmutex);
 
         op->setSeqNumFromQueue();
 
@@ -1480,6 +1483,8 @@ public:
     // Also different modes and optimizations can control when dependencies are added.
     // TODO - return reference if possible to avoid shared ptr overhead.
     std::shared_ptr<KalmarAsyncOp> detectStreamDeps(hcCommandKind newCommandKind, KalmarAsyncOp *kNewOp) {
+
+        std::lock_guard<std::recursive_mutex> lg(qmutex);
 
         const auto newOp = static_cast<const HSAOp*> (kNewOp);
 
@@ -1544,6 +1549,7 @@ public:
 
 
     int getPendingAsyncOps() override {
+        std::lock_guard<std::recursive_mutex> lg(qmutex);
         int count = 0;
         for (int i = 0; i < asyncOps.size(); ++i) {
             auto &asyncOp = asyncOps[i];
@@ -1568,6 +1574,8 @@ public:
         // Have to walk asyncOps since it can contain null pointers (if event is waited on and removed)
         // Also not all commands contain signals.
         
+        std::lock_guard<std::recursive_mutex> lg(qmutex);
+
         bool isEmpty = true;
 
         const auto& oldest = find_if(
@@ -1616,7 +1624,7 @@ public:
             printAsyncOps(std::cerr);
         }
 
-
+        std::lock_guard<std::recursive_mutex> lg(qmutex);
 
 #if !defined(NDEBUG)
         bool foundFirstValidOp = false;
@@ -1627,8 +1635,7 @@ public:
             lastWaitOp = (lastWaitOp * QUEUE_FLUSHING_FRAC);
         }
 
-        int i;
-        for (i = lastWaitOp-1; i >= 0;  i--) {
+        for (int i = lastWaitOp-1; i >= 0;  i--) {
             if (asyncOps[i] != nullptr) {
                 auto asyncOp = asyncOps[i];
                 // we only drain starting at first found marker
@@ -1682,7 +1689,7 @@ public:
             asyncOps.clear();
             asyncOps_offset = 0;
         }
-   }
+    }
 
     void LaunchKernel(void *ker, size_t nr_dim, size_t *global, size_t *local) override {
         LaunchKernelWithDynamicGroupMemory(ker, nr_dim, global, local, 0);
@@ -2221,6 +2228,9 @@ public:
 
     // remove finished async operation from waiting list
     void removeAsyncOp(HSAOp* asyncOp) {
+
+        std::lock_guard<std::recursive_mutex> lg(qmutex);
+
         int targetIndex = asyncOp->asyncOpsIndex() - asyncOps_offset;
 
         // Make sure the opindex is still valid.
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -1646,8 +1646,17 @@ public:
             }
             /* we can only safely clear from the first nullptr and older */
             if (first_nullptr >= 0) {
-                // TODO something faster than std::vector erase?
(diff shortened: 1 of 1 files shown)


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1613,30 +1624,72 @@ class HSAQueue final : public KalmarQueue
...
 
+        std::lock_guard<std::recursive_mutex> lg(qmutex);
+
+#if !defined(NDEBUG)
+        bool foundFirstValidOp = false;
+#endif
+        int first_nullptr = -1;
+        int lastWaitOp = asyncOps.size();
+        if (drainingQueue_) {
+            lastWaitOp = (lastWaitOp * QUEUE_FLUSHING_FRAC);
```
whchung: nit: maybe a static_cast here?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -137,6 +134,10 @@ int HCC_PROFILE_VERBOSE=0x1F;
 
 char * HCC_PROFILE_FILE=nullptr;
 
+int HCC_QUEUE_FLUSHING_RATIO=50;
+
+double QUEUE_FLUSHING_FRAC;
```
whchung: nit: all calculation based on this value seems to be in `float`?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1613,30 +1624,72 @@ class HSAQueue final : public KalmarQueue
...
-                    std::shared_future<void>* future = asyncOp->getFuture();
-                    if (future && future->valid()) {
-                        future->wait();
-                    }
+        for (int i = lastWaitOp-1; i >= 0;  i--) {
+            if (asyncOps[i] != nullptr) {
+                auto asyncOp = asyncOps[i];
+                // we only drain starting at first found marker
+                if (drainingQueue_ && asyncOp->getCommandKind() != hcCommandMarker) continue;
+#if !defined(NDEBUG)
```
whchung: @jeffdaily could you clarify the rationale of this check? it seems `foundFirstValidOp` would NOT be set `true` in debug build?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1292,6 +1294,7 @@ class HSAQueue final : public KalmarQueue
     // will be waited on.
     //
     std::vector< std::shared_ptr<HSAOp> > asyncOps;
+    int asyncOps_offset;
```
whchung, line 1302: comment for this important field is required.

lib/hsa/mcwamp_hsa.cpp
whchung: a deeper look suggest `foundFirstValidOp` is not really used anywhere else... should it be removed?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1613,30 +1624,72 @@ class HSAQueue final : public KalmarQueue
...
+                // because erasing from front of vectors will repeatedly move remaining elements.
+                // We can also do better than the improved
+                // asyncOps.erase(std::remove(asyncOps.begin(), asyncOps.begin()+first_nullptr+1, nullptr), asyncOps.end());
+                // because we know that all elements from 0..first_nullptr are nullptrs,
+                // so we can avoid the redundant checks of each element against nullptr.
+                for (size_t i=first_nullptr+1,new_i=0,end=asyncOps.size(); i<end; ++i) {
+                    asyncOps[new_i++] = std::move(asyncOps[i]);
+                }
+                asyncOps.resize(asyncOps.size() - first_nullptr - 1);
+                asyncOps_offset += first_nullptr + 1;
```
whchung: nit: `(first_nullptr + 1)`

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1613,30 +1624,72 @@ class HSAQueue final : public KalmarQueue
...
+            }
+            /* we can only safely clear from the first nullptr and older */
+            if (first_nullptr >= 0) {
+                // We can do something faster than
+                // asyncOps.erase(asyncOps.begin(), asyncOps.begin() + first_nullptr + 1);
+                // because erasing from front of vectors will repeatedly move remaining elements.
+                // We can also do better than the improved
+                // asyncOps.erase(std::remove(asyncOps.begin(), asyncOps.begin()+first_nullptr+1, nullptr), asyncOps.end());
+                // because we know that all elements from 0..first_nullptr are nullptrs,
+                // so we can avoid the redundant checks of each element against nullptr.
```
whchung, line 1680: the comment seems not finished to me. could you amend the comment why erase/remove idiom not used here?

lib/hsa/mcwamp_hsa.cpp
jeffdaily: `foundFirstValidOp` and its use was present before my changes.  I was trying to preserve the original code, but I noticed that `foundFirstValidOp` was only used to trigger the code path for the `assert(sig.handle != 0)`.  Perhaps it was premature optimization, but the `assert` would be a no-op if `NDEBUG` is defined, so I tried removing the entire use of `foundFirstValidOp` if `NDEBUG` was defined.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1599,9 +1610,9 @@ class HSAQueue final : public KalmarQueue
         //
```
whchung, line 1617: as the algorithm becomes much more complicated with the introduction of `asyncOps_offset` field, I think it's prudent to have a paragraph of comment to explain the algorithm better.

whchung (changes requested):
@jeffdaily overall I feel more comments are needed, plus some minor editorial changes.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -77,10 +77,7 @@
 // If limit is exceeded, HCC will force a queue wait to reclaim
 // resources (signals, kernarg)
 // MUST be a power of 2.
-#define MAX_INFLIGHT_COMMANDS_PER_QUEUE  (2*8192)
-
-// threshold to clean up finished kernel in HSAQueue.asyncOps
-#define ASYNCOPS_VECTOR_GC_SIZE (2*8192)
+int HCC_MAX_INFLIGHT_COMMANDS_PER_QUEUE=(2*8192);
```
whchung, line 80: @jeffdaily I'm not very fond of this less-than-desirable coding style introduced by @bensander ... ALL-CAPITAL global variables... but we'll leave it for now

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1442,19 +1445,27 @@ class HSAQueue final : public KalmarQueue
 
 
 
-        if (!drainingQueue_ && (asyncOps.size() >= MAX_INFLIGHT_COMMANDS_PER_QUEUE-1)) {
+        if (!drainingQueueFailed_ && !drainingQueue_ && (asyncOps.size() >= HCC_MAX_INFLIGHT_COMMANDS_PER_QUEUE-1)) {
```
whchung, line 1453: a second re-read of the logic here makes it hard to distinguish which value is variable, and which is constant.

to my understanding, originally `HCC_MAX_INFLIGHT_COMMANDS_PER_QUEUE` was a constant (macro), but now it's a global variable modifiable by environment variables. it's less than desirable.

whchung: .. but I think we can live with it for now as long as these values are set / accessed in a consistent fashion.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1613,30 +1624,72 @@ class HSAQueue final : public KalmarQueue
...
-                        hsa_signal_t sig =  *(static_cast <hsa_signal_t*> (asyncOp->getNativeHandle()));
-                        assert(sig.handle != 0);
-                        foundFirstValidOp = true;
-                    }
-                    // wait on valid futures only
-                    std::shared_future<void>* future = asyncOp->getFuture();
-                    if (future && future->valid()) {
-                        future->wait();
-                    }
+        for (int i = lastWaitOp-1; i >= 0;  i--) {
```
whchung, line 1647: comment for this loop is necessary I think. the purpose is to find the first op which has become `nullptr`, and trigger the cleaning process in the next block?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1613,30 +1624,72 @@ class HSAQueue final : public KalmarQueue
...
+                asyncOps.resize(asyncOps.size() - first_nullptr - 1);
+                asyncOps_offset += first_nullptr + 1;
+                DBOUTL(DB_RESOURCE, "asyncOps=" << &asyncOps << " * erasing " << first_nullptr+1 << " ops. New size " << asyncOps.size());
+            }
+        }
+        else if (drainingQueue_) {
+            /* we were not able to find a suitable op to wait on */
+            drainingQueueFailed_ = true;
+        }
+        else {
```
whchung, line 1700: there are 3 checks in this block:
```
if (drainingQueue_ && first_nullptr >= 0) {
} else if (drainingQueue_) {
} else {
}
```

would it be better to re-organize it to:

```
if (drainingQueue_) {
  if (first_nullptr >= 0) {
  } else {
  }
} else {
}
```

so perhaps it's better comprehensible?

whchung (changes requested):

lib/hsa/mcwamp_hsa.cpp
whchung: I think the logic is to ensure `asyncOp->getFuture()` won't crash later on. From `git blame` it doesn't really seem to offer any specific help. Maybe we should remove the logic here.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2178,7 +2231,7 @@ class HSAQueue final : public KalmarQueue
 
         std::lock_guard<std::recursive_mutex> lg(qmutex);
 
-        int targetIndex = asyncOp->asyncOpsIndex();
+        int targetIndex = asyncOp->asyncOpsIndex() - asyncOps_offset;
```
whchung, line 2252: any statements affected by `asyncOps_offset` deserves to have comment I think

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2215,13 +2268,9 @@ class HSAQueue final : public KalmarQueue
...
         }
 
-
-        // GC for finished kernels
-        if (asyncOps.size() > ASYNCOPS_VECTOR_GC_SIZE) {
-            DBOUTL(DB_RESOURCE, "asyncOps size=" << asyncOps.size() << " exceeds collection size, compacting");
-            asyncOps.erase(std::remove(asyncOps.begin(), asyncOps.end(), nullptr),
-                         asyncOps.end());
-        }
+        /* We used to garbage collect nullptr entries from asyncOps here.
```
whchung: nit: use `//` in all comments, both single-line and multi-line.

whchung (changes requested):

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1442,19 +1445,27 @@ class HSAQueue final : public KalmarQueue
...
+
+            if (drainingQueueFailed_) {
+                /* draining queue failed, so truly force a wait */
+                DBOUTL(DB_RESOURCE, "draining queue failed, truly forcing sync");
+                wait();
+                drainingQueueFailed_ = false;
+            }
         }
-        op->asyncOpsIndex(asyncOps.size());
+        op->asyncOpsIndex(asyncOps.size()+asyncOps_offset);
```
whchung, line 1472: any statement be associated with `asyncOps_offset` deserves to have comment on the effect brought by this important field.

lib/hsa/mcwamp_hsa.cpp
jeffdaily: static_cast the asyncOps.size() to int?  Or static_cast `lastWaitOp*QUEUE_FLUSHING_FRAC` to int?

lib/hsa/mcwamp_hsa.cpp
jeffdaily: I believe you are suggesting to have only one variable instead of the later extra step of `QUEUE_FLUSHING_FRAC = HCC_QUEUE_FLUSHING_RATIO / 100.0`?  I could see having the user specify a value between 0 and 1, e.g., 0.5, 0.4.

Otherwise I took your comment to say use `float` instead of `double`, although I thought the default without the `f` suffix was to use `double`, so the later calculations that use `QUEUE_FLUSHING_FRAC` would be using `double` anyhow.

lib/hsa/mcwamp_hsa.cpp
whchung: static_cast `lastWaitOp * QUEUE_FLUSHING_FRAC`. I don't really like implicit casting for floating point numbers.

lib/hsa/mcwamp_hsa.cpp
jeffdaily: Comment added in next commit.

lib/hsa/mcwamp_hsa.cpp
whchung: yes, it'd be better if we could remove this global variable, or at least keep it an integer.

jeffdaily: I removed `QUEUE_FLUSHING_FRAC` from my next commit.  `HCC_QUEUE_FLUSHING_RATIO` remains an `int` and is used directly when finding the index to start draining from.

lib/hsa/mcwamp_hsa.cpp
jeffdaily: Done.

lib/hsa/mcwamp_hsa.cpp
jeffdaily: `QUEUE_FLUSHING_FRAC` was removed.  Logic here is updated and uses `int` calculations only.

lib/hsa/mcwamp_hsa.cpp
jeffdaily: Done.

lib/hsa/mcwamp_hsa.cpp
jeffdaily: Logic removed.

lib/hsa/mcwamp_hsa.cpp
jeffdaily: Done.

lib/hsa/mcwamp_hsa.cpp
jeffdaily: Amended comment.

lib/hsa/mcwamp_hsa.cpp
jeffdaily: I agree, it is now more clear when reorganized as suggested.

lib/hsa/mcwamp_hsa.cpp
jeffdaily: Done.

lib/hsa/mcwamp_hsa.cpp
jeffdaily: Done.

jeffdaily (commented):
New commit addressing comments has been pushed.

whchung (approved):
Thanks for the revision! LGTM

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1613,30 +1631,77 @@ class HSAQueue final : public KalmarQueue
...
+        // We may also not find a nullptr, or not find a valid op to wait upon.
+        for (int i = lastWaitOp-1; i >= 0;  i--) {
+            if (asyncOps[i] != nullptr) {
+                auto asyncOp = asyncOps[i];
+                // we only drain starting at first found marker
+                if (drainingQueue_ && asyncOp->getCommandKind() != hcCommandMarker) continue;
+                // wait on valid futures only
+                std::shared_future<void>* future = asyncOp->getFuture();
+                if (future && future->valid()) {
+                    future->wait();
```
scchan, line 1655: We are holding the lock to this asyncOps and wait so this may cause a potential deadlock here, need to take a closer look

scchan (changes requested):

lib/hsa/mcwamp_hsa.cpp
jeffdaily: `future->wait()` happens in the original code for `HSAQueue::wait()`, so this deadlock might not be specific to my PR.   I'm looking at this code closely now, and indeed the thread holding the lock for `asyncOps` will trigger `hsa_signal_wait_scacquire` because of the `future->wait()`.  If queue stealing occurs by another thread while the signal wait is blocking, could this cause the deadlock?

scchan:
superseded by PR1046

PR #1001, opened 2019-01-24
enabling PROF_API find_path

eshcherb:


pfultz2:
Is this a build-only dependency? Or will users need to add the profiling paths when using the hcc runtime?

CMakeLists.txt
```diff
@@ -175,7 +175,13 @@ endif (USE_CODEXL_ACTIVITY_LOGGER EQUAL 1)
 if(USE_PROF_API EQUAL 1)
 add_definitions(-DUSE_PROF_API=1)
 if (NOT DEFINED PROF_API_HEADER_PATH)
-  set(PROF_API_HEADER_PATH ${ROCM_ROOT}/roctracer/include/ext)
+  find_package(PkgConfig)
+  pkg_check_modules(ROCTRACERPROTO QUIET roctracer-proto)
+  if(ROCTRACERPROTO_FOUND EQUAL 1)
+    pkg_get_variable(PROF_API_HEADER_PATH roctracer-proto prefix)
```
pfultz2: Why does it just get the root path? Shouldn't it get all usage requirements?

eshcherb: This is build dependencies.
This path to tracer protocol header.

eshcherb: Which usage requirements do you mean?

pfultz2: > Which usage requirements do you mean?

The usage requirements of roctracer-proto. If roctracer-proto needs other flags in the future then hcc won't need to change its build.

> This is build dependencies.

So users of hcc will never need the protocol header.

pfultz2: > This path to tracer protocol header.

If this is one file, then why not just use `find_path`?

eshcherb: No ,it is internal protocol for tracing data between HCC and tracer

eshcherb: To use pkg-config is a requirement of CI team.
I was quite happy with just CMAKE PATH parameter.

eshcherb: Also as you mention you can provide more parameters with pkg-config.
Don't have currently any but it can be easily extended.

eshcherb: Previously I was working with Siu Chi on the tracer integration.

eshcherb: Just sent you his request for adding pkg-config support.

pfultz2: > I was quite happy with just CMAKE PATH parameter.

And you should be able to continue do so, since cmake sets the pkgconfig paths based on the `CMAKE_PREFIX_PATH`.

> Don't have currently any but it can be easily extended.

Yes, pkgconfig provides more than just include paths, however it isn't being used, so I dont see the point.

If you use pkgconfig, it should be used something like this:

```cpp
pkg_check_modules(ROCTRACERPROTO QUIET roctracer-proto)
add_library(PkgConfig::ROCTRACERPROTO INTERFACE IMPORTED)

set_target_properties(PkgConfig::ROCTRACERPROTO PROPERTIES
  PUBLIC_INCLUDE_DIRECTORIES "${ROCTRACERPROTO_INCLUDE_DIRS}"
  PUBLIC_LINK_LIBRARIES "${ROCTRACERPROTO_LIBRARIES}"
  PUBLIC_COMPILE_OPTIONS "${ROCTRACERPROTO_CFLAGS_OTHER}"
  PUBLIC_SYSTEM_INCLUDE_DIRECTORIES "${ROCTRACERPROTO_INCLUDE_DIRS}"
)
# Link roctracer-proto
target_link_libraries(mcwamp_hsa PRIVATE PkgConfig::ROCTRACERPROTO)
```

eshcherb: roctracer library is not linked to HCC

pfultz2: Then what is the header path used for?

pfultz2: > roctracer library is not linked to HCC

It doesnt "link", in the compiler sense. `PkgConfig::ROCTRACERPROTO` is an interface target. The `target_link_libararies` are used to consume the usage requirements of another target.

It calls `include_directories ( ${PROF_API_HEADER_DIR} )`, instead a target should be used. 

It also looks like the pkgconfig is wrong for roctracer-proto. The prefix should be the install prefix, not the include directory.

Also, is roctracer-proto written in cmake? If so, it should provide a `find_package` instead of using pkgconfig.

eshcherb: Header is used to provide data structures types for callbacks common for different ROCm runtimes.
My understanding that CI team would like to use pkg_config to provide path to actual installation of the header which can be different.

pfultz2: But that can be done with `find_path` as long as you dont add `NO_DEFAULT_PATH` to the command.

eshcherb: Yes, 'find_pah', but my previous comment was for 'find_package'.
Argument against 'find_path' that pkg_config potentially can provide more parameters, not just installation path.

eshcherb: Actually for find_path you still need to provide some root path.

eshcherb: What variable I need to use of include path?
I can rename it to 'includedir', does it make huge difference?

pfultz2: > Argument against 'find_path' that pkg_config potentially can provide more parameters, not just installation path.

But you aren't using those, so there seems to be no need for pkgconfig.

> I can rename it to 'includedir', does it make huge difference?

If you are using pkgconfig you shouldn't be reading the variables directly. You should be getting the compile and link flags, put them in a target and link against it.

Ultimately, I dont understand why roctracer can't support cmake(ie find_package) instead of going through pkgconfig which is more complicated. I think it would be much better to remove `NO_DEFAULT_PATH` from `find_path` and when roctracer has more complicated usage requirements then it can add support for `find_package` just like the all the rest of our libraries and runtimes do. Then hcc can just call `find_package(roctracer)`.

eshcherb: To support find_package I need to provide <package>_FOUND
Is there an example of find_package usage?

eshcherb: Looks like the 'package'_FOUND file eth CMAKE_MODULE_PATH should be set.
How the header path will bee queried?

pfultz2: > To support find_package I need to provide _FOUND

No, cmake will provide that.

> Is there an example of find_package usage?

Yes, hcc provides a [hcc-config.cmake](https://github.com/RadeonOpenCompute/hcc/blob/clang_tot_upgrade/lib/hcc-config.cmake.in) file so users can call `find_package(hcc)`. There is no need for setting a module path.

To support `find_package` in rorctracer, roctracer would install a `roctracer-config.cmake` file in `lib/cmake/roctracer`. Then you can put whatever variables you want in it. Usually libraries export their targets so they can be used directly. You can do it manually like what is done in hcc or you can use [`rocm_export_targets`](https://github.com/RadeonOpenCompute/rocm-cmake#rocm_export_targets) from rocm-cmake modules. That is what is done in [miopen](https://github.com/ROCmSoftwarePlatform/MIOpen/blob/master/src/CMakeLists.txt#L443), [rocblas](https://github.com/ROCmSoftwarePlatform/rocBLAS/blob/develop/library/src/CMakeLists.txt#L197), [rocfft](https://github.com/ROCmSoftwarePlatform/rocFFT/blob/develop/library/src/CMakeLists.txt#L114), [migraphx](https://github.com/ROCmSoftwarePlatform/AMDMIGraphX/blob/develop/src/CMakeLists.txt#L44), and [miopengemm](https://github.com/ROCmSoftwarePlatform/MIOpenGEMM/blob/master/miopengemm/CMakeLists.txt#L39).

pfultz2: Then you would call `find_package(roctracer PATHS /opt/rocm/roctracer)` to get the roctracer target.

eshcherb: I can change pkgconfig code to use includedir and external target properties to link mcwamp_hsa

pfultz2 (approved):

PR #1007, opened 2019-01-28
Revert Combine opt invocations in clamp-device

aaronenyeshi:
This is meant to fix a rocPRIM unit test failure: rocprim.hc.device_scan

### DIFF
--- a/lib/clamp-device.in
+++ b/lib/clamp-device.in
@@ -185,15 +185,24 @@ if [ $KMDUMPLLVM == "1" ]; then
 fi
 
 # Invoke HCC-specific opt passes
-# Optimization notes:
-#  -disable-simplify-libcalls:  prevents transforming loops into library calls such as memset, memcopy on GPU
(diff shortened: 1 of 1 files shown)


aaronenyeshi:
This commit partially reverts commit: https://github.com/RadeonOpenCompute/hcc/commit/05f3bcc682ea1446ae6c50660383d54806ad4149

david-salinas:
Does the change in clamp-link.in also need to be reverted, or just clamp-device.in ?

aaronenyeshi:
Just the clamp-device.in change will fix the failure

aaronenyeshi:
The root cause is that -O3 will change the structure of custom passes. So when we combined the -O3 argument with our two custom passes, the custom passes stop working as intended. Addition pre-pass and post-pass optimizations are added to our custom passes which causes a big discrepancy in LLVM IR

david-salinas (approved):
LGTM

scchan (approved):
LGTM

PR #1009, opened 2019-01-29
createOrstealRocrQueue: fix wrong assumption about pointer to victim queue

scchan:
In createOrstealRocrQueue: fix the wrong assumption in the 2nd pass about the pointer to the victim queue being always non-null, refactor the code


david-salinas (approved):
LGTM

PR #1013, opened 2019-01-29
protect accesses to HSAQueue::asyncops with a mutex

scchan:
Accesses to the async ops vector in HSAQueue need to be protected by a mutex since the queue stealing algorithm may run into a race condition.  When the runtime runs out of queue resources, the queue stealing algorithm will look at queues already assigned to an hc::accelerator_view/hipStream and try to find one that is unused/empty.  The problem is that these hc::accelerator_views/hipStreams could be owned by other threads and may concurrently enqueue new work while being inspected by the queue stealing algorithm run by another thread.  Protect the accesses to the async ops vector with a lock avoids the race condition.

This issue was exposed by caffe2 training with multiple GPUs.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -1388,6 +1388,7 @@ public:
     //
     void printAsyncOps(std::ostream &s = std::cerr)
     {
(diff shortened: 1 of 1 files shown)


david-salinas (approved):
LGTM

Issue #1011, opened 2019-01-29
hcc::__activelaneid_u32() doesn't seem to work in HCC 2.0


dragontamer:
This was originally reported in the ROCm issue tracker. (https://github.com/RadeonOpenCompute/ROCm/issues/688) They asked me to report it here instead.

-------------

Summary of your hardware: Threadripper 1950x + Vega64

PCIe Information: PCIe 3.0.

Here's my test code.

```
#include<iostream>
#include<stdint.h>

#include <hc.hpp>

int main(){
	const int N = (1 << 30) / sizeof(uint32_t); 
	uint32_t* stuff = new uint32_t[N];
	uint32_t* output = new uint32_t[N];

	for(int i=0; i<N; i++){
		stuff[i] = i;
	}

	hc::array_view<uint32_t, 1> av_in(N, stuff);
	hc::array_view<uint32_t, 1> av_out(N, output);

	hc::parallel_for_each(hc::extent<1>(N), [=](hc::index<1> i) [[hc]] {
		av_out[i[0]] = hc::__activelaneid_u32();
	});

	for(int i=0; i< 100; i++){
		std::cout << i << "    " << av_out[i] << "\n"; 
	}
}
```

When I compile...

```
hcc `hcc-config --cxxflags --ldflags` test.cpp -o test
ld.lld: error: relocation R_AMDGPU_REL32_LO cannot be used against symbol __activelaneid_u32; recompile with -fPIC
>>> defined in /tmp/tmp.hO38AglmSw/kernel-gfx900.hsaco.isabin
>>> referenced by /tmp/tmp.hO38AglmSw/kernel-gfx900.hsaco.isabin:(main::$_0::__cxxamp_trampoline(unsigned int*, int, int, int, int, int, int, int))

ld.lld: error: relocation R_AMDGPU_REL32_HI cannot be used against symbol __activelaneid_u32; recompile with -fPIC
>>> defined in /tmp/tmp.hO38AglmSw/kernel-gfx900.hsaco.isabin
>>> referenced by /tmp/tmp.hO38AglmSw/kernel-gfx900.hsaco.isabin:(main::$_0::__cxxamp_trampoline(unsigned int*, int, int, int, int, int, int, int))
Generating AMD GCN kernel failed in ld.lld for target: gfx900
clang-8: error: linker command failed with exit code 1 (use -v to see invocation)
Makefile:2: recipe for target 'test' failed
make: *** [test] Error 1
```

Version information:

```
[...]

PR #1018, opened 2019-02-01
have activelaneid_u32 call ockl_activelane_u32

david-salinas:
for https://github.com/RadeonOpenCompute/hcc/issues/1011

### DIFF
--- a/include/hc.hpp
+++ b/include/hc.hpp
@@ -2817,7 +2817,10 @@ extern "C" inline __attribute((always_inline)) std::uint64_t __cycle_u64() __HC_
  *
  * @return The result will be in the range 0 to WAVESIZE - 1.
  */
-extern "C" unsigned int __activelaneid_u32() __HC__;
+extern "C" unsigned int __ockl_activelane_u32(void);
+extern "C" inline unsigned int __activelaneid_u32() __HC__ {
+	return __ockl_activelane_u32();
+}
 
 /**
  * Return a bit mask shows which active work-items in the
--- a/tests/Unit/AMDGPU/activelaneid.cpp
+++ b/tests/Unit/AMDGPU/activelaneid.cpp
@@ -1,0 +1,107 @@
+
+// RUN: %hc %s -o %t.out && %t.out
+
+#include <hc.hpp>
+
+#include <iostream>
+#include <random>
+#include <vector>
+
+#define WAVEFRONT_SIZE (64) // as of now, all HSA agents have wavefront size of 64
+
+#define GRID_SIZE (WAVEFRONT_SIZE * WAVEFRONT_SIZE)
+
+#define TEST_DEBUG (0)
+
+// A test case to verify builtin function
+// - __activelaneid_u32
+
+// test __activelaneid_u32
+bool test() {
+  using namespace hc;
+  bool ret = true;
+
+  // initialize test data
+  // test is a table of size WAVEFRONT_SIZE * WAVEFRONT_SIZE
+  std::vector<uint32_t> test(GRID_SIZE);
+
+  std::random_device rd;
+  std::uniform_int_distribution<int> int_dist(0, WAVEFRONT_SIZE - 1);
+
+  // for each block of WAVEFRONT_SIZE, we randomly set 1s inside the block
+  // the number of 1s in the block equals to the index of the block
+  // (the 1st block of WAVEFRONT_SIZE has 0 1s, the 2nd block of WAVEFRONT_SIZE has 1 1, and so on)
+  for (int i = 0; i < WAVEFRONT_SIZE; ++i) {
+    for (int j = 0; j < WAVEFRONT_SIZE; ++j) {
+      if (j < i) {
+        test[i * WAVEFRONT_SIZE + j] = 1;
+      } else {
+        test[i * WAVEFRONT_SIZE + j] = 0;
+      }
+    }
+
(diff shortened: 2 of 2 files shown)


b-sumner:
Is there a test that checks that all/most of the __foo functions compile?  If so it needs updates too.

david-salinas:
> Is there a test that checks that all/most of the __foo functions compile? If so it needs updates too.

sorry, I should have already had a test ready.  Let me cook one up, and update this PR.

tests/Unit/AMDGPU/activelaneid.cpp
```diff
@@ -0,0 +1,97 @@
...
+        test[i * WAVEFRONT_SIZE + j] = 0;
+      }
+    }
+
+    // randomly shuffle items in the block
+    for (int j = 0; j < WAVEFRONT_SIZE * 10; ++j) {
+      int k1 = int_dist(rd);
+      int k2 = int_dist(rd); 
+      if (k1 != k2) {
+        test[i * WAVEFRONT_SIZE + k1] ^= test[i * WAVEFRONT_SIZE + k2] ^= test[i * WAVEFRONT_SIZE + k1] ^= test[i * WAVEFRONT_SIZE + k2];
```
b-sumner: Why not just code a swap?

tests/Unit/AMDGPU/activelaneid.cpp
```diff
@@ -0,0 +1,97 @@
...
+      }
+    }
+  }
+
+  for (int i = 0; i < WAVEFRONT_SIZE; ++i) {
+    for (int j = 0; j < WAVEFRONT_SIZE; ++j) {
+#if TEST_DEBUG
```
b-sumner: I would move this outside the entire loop nest

tests/Unit/AMDGPU/activelaneid.cpp
```diff
@@ -0,0 +1,97 @@
...
+#endif
+  }
+
+  array<uint32_t, 1> test_GPU(GRID_SIZE);
+  copy(test.begin(), test_GPU);
+
+  array<uint32_t, 1> output_GPU(GRID_SIZE);
+  extent<1> ex(GRID_SIZE);
+  parallel_for_each(ex, [&](index<1>& idx) [[hc]] {
+    output_GPU(idx) = __activelaneid_u32();
```
b-sumner: Every lane is sure to be active.  Control flow is needed to deactivate some lanes.

tests/Unit/AMDGPU/activelaneid.cpp
```diff
@@ -0,0 +1,97 @@
...
+      std::cout << test[i * WAVEFRONT_SIZE + j] << " ";
+#endif
+    }
+#if TEST_DEBUG
+    std::cout << "\n";
+#endif
+  }
+
+  array<uint32_t, 1> test_GPU(GRID_SIZE);
+  copy(test.begin(), test_GPU);
```
scchan, line 63: `test_GPU` is not being used so the data generation in `test` is redundant so they should be removed

include/hc.hpp
```diff
@@ -2817,7 +2817,11 @@ extern "C" inline __attribute((always_inline)) std::uint64_t __cycle_u64() __HC_
  *
  * @return The result will be in the range 0 to WAVESIZE - 1.
  */
-extern "C" unsigned int __activelaneid_u32() __HC__;
+extern "C" __attribute__((always_inline))
```
scchan: I don't think this needs the always_inline attribute here

include/hc.hpp
```diff
@@ -2817,7 +2817,11 @@ extern "C" inline __attribute((always_inline)) std::uint64_t __cycle_u64() __HC_
...
+unsigned int __ockl_activelane_u32(void);
+extern "C" unsigned int __activelaneid_u32() __HC__ {
```
scchan: this would be marked as `inline` to avoid multiple definitions

scchan (changes requested):

scchan:
@david-salinas  ping...

tests/Unit/AMDGPU/activelaneid.cpp
david-salinas: Looking at this test (I based it on the existing "activelanecount.cpp" test) most of this set up of "test" can be scrapped, I think.  We can just create a PFE, with a suitable extent, and an output grid/vector and call activelaneid to verify that we get correct/valid laneid values.

tests/Unit/AMDGPU/activelaneid.cpp
david-salinas: ok.  I'm thinking I can scrap "test" too.  I based this test on "activelanecount.cpp".  For activelaneid, we really just need a PFE and an output grid, to verify activelaneid is returning the correct lane id.  I think I can reduce this test, by a lot.

b-sumner:
I'm not sure I'm seeing the latest version of the test code.  I agree it can be small, but the kernel should have some control flow so that we can see that activelaneid is reporting the correct ID on lanes that are truly active.

scchan:
I don't see any update to the code either

david-salinas:
Sorry,  I was about to push an update test, when I realized I need to add some control flow to deactivate some lanes.  I should have an update shortly.

scchan (approved):

PR #1022, opened 2019-02-04
Update extractkernel to support multiple embedded device code blobs

scchan:
This adds to the extractkernel tool the support for extracting multiple embedded device code blobs from a host executable.


whchung (approved):
verified with actual applications

Issue #1021, opened 2019-02-03
hc::accelerator_view::get_hsa_queue() returns RocrQueue* instead of hsa_queue_t*


misos1:
RocrQueue is private structure inside mcwamp_hsa.cpp so there is no clean way how to retrieve its property "hsa_queue_t *_hwQueue". Examples like this are not working:

https://gpuopen.com/rocm-with-harmony-combining-opencl-hcc-hsa-in-a-single-program/
https://github.com/ROCm-Developer-Tools/HCC-Example-Application/tree/master/BitonicSort-CL-from-HCC

From above example:

``` c++
    /*
     * Extract the hsaQueue from the HCC acclerator_view:
     */
    hc::accelerator_view av = _acc.get_default_view();
    hsa_queue_t  *hsaQueue = static_cast<hsa_queue_t*> (av.get_hsa_queue());
```

Instead of above one have to do it in this way:

``` c++
    /*
     * Extract the hsaQueue from the HCC acclerator_view:
     */
    hc::accelerator_view av = _acc.get_default_view();
    hsa_queue_t  *hsaQueue = *static_cast<hsa_queue_t**> (av.get_hsa_queue());
```

But this is based on assumption that _hwQueue will be always the first member of RocrQueue.

``` c++
    /**
     * Returns an opaque handle which points to the underlying HSA queue.
     *
     * @return An opaque handle of the underlying HSA queue, if the accelerator
     *         view is based on HSA.  NULL if otherwise.
     */
    void* get_hsa_queue() {
        return pQueue->getHSAQueue();
    }
```

``` c++
    RocrQueue    *rocrQueue;
...
    void* getHSAQueue() override {
        return static_cast<void*>(rocrQueue);
    }
```

Maybe getHSAQueue should look rather like this (not tested):

``` c++
    void* getHSAQueue() override {
        return static_cast<void*>(rocrQueue->_hwQueue);
    }
```

david-salinas:
Agreed.  I'll create a PR to fix this.

david-salinas:
Good News.  Change has been made in PR #1029.

PR #1029, opened 2019-02-07
getHSAQueue returns RocrQueue* instead of has_queue_t*

david-salinas:
https://github.com/RadeonOpenCompute/hcc/issues/1021


scchan (approved):
LGTM

PR #1031, opened 2019-02-08
Enable lazy initialization by default

scchan:
- enable the lazy init flag by default
- teach memcpySymbol to invoke initialization 
- update and refactor the accelerator_get_all_views_mt test

### DIFF
--- a/lib/mcwamp.cpp
+++ b/lib/mcwamp.cpp
@@ -378,17 +378,16 @@ private:
   RuntimeImpl* runtime;
 public:
   KalmarBootstrap() : runtime(nullptr) {
-    bool to_init = true;
-    char* lazyinit_env = getenv("HCC_LAZYINIT");
-    if (lazyinit_env != nullptr) {
-      if (std::string("ON") == lazyinit_env) {
-        to_init = false;
-      } else if (strtol(lazyinit_env, nullptr, 0)) {
-        to_init = false;
+    bool lazy_init = true;
+    const char* lazyinit_env = getenv("HCC_LAZYINIT");
+    if (lazyinit_env != std::nullptr &&
+         (std::string("OFF") == lazyinit_env ||
+            strtol(lazyinit_env, nullptr, 0) == 0)) {
+        lazy_init = false;
       }
     }
 
-    if (to_init) {
+    if (!lazy_init) {
       // initialize runtime
       runtime = CLAMP::GetOrInitRuntime();
 
--- a/lib/mcwamp.cpp
+++ b/lib/mcwamp.cpp
@@ -329,17 +329,23 @@ static void read_code_bundles(std::vector<_code_bundle>& bundles) {
   }
 }
 
-void LoadInMemoryProgram(KalmarQueue* pQueue) {
+
+
+void LoadInMemoryProgram(KalmarDevice* k_device) {
 
   static std::vector<_code_bundle> bundles;
   static std::once_flag f;
   std::call_once(f, [&](){ read_code_bundles(bundles); });
 
   for (auto&& b : bundles) {
-    if (pQueue->getDev()->IsCompatibleKernel((void*) b.size, (void*) b.device_binary)) {
-      pQueue->getDev()->BuildProgram((void*) b.size, (void*) b.device_binary);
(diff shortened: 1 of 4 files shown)


lib/mcwamp.cpp
```diff
@@ -329,19 +329,25 @@ static void read_code_bundles(std::vector<_code_bundle>& bundles) {
   }
 }
 
-void LoadInMemoryProgram(KalmarQueue* pQueue) {
+
+
+void LoadInMemoryProgram(KalmarDevice* k_device) {
 
   static std::vector<_code_bundle> bundles;
   static std::once_flag f;
```
whchung, line 337: `std::once_flag` may be broken in case:

- applications somehow link with HCC runtime statically
- applications somehow loads 2 different versions of HCC runtime

whchung: please refer to examples in SWDEV-178515 for reduced cases to break `std::once_flag` and possible ways to fix it with FIFO.

tests/Unit/HC/accelerator_get_all_views_mt.cpp
```diff
@@ -5,61 +5,41 @@
...
 
-#if !TLS_QUEUE
-  ret &= (v5.size() == (1 + c1 + c2 + c3 + c4));
-#else
-  // now default accelerator view is thread local, so the total number of views
-  // would the the number of threads (4), plus additional views created in each
-  // thread, plus another one inside HCC runtime
-  ret &= (v5.size() == (4 + c1 + c2 + c3 + c4 + 1));
+#if 0
+  std::cout << "number of views: expected=" << expected_unique_views << " actual=" << all_views.size() << std::endl;
```
whchung: nit-picking: this block is probably not needed?

whchung (commented):
Other than potential pitfall of `std::once_flag` which probably deserves another PR to address it, the PR looks largely fine to me. Please help address some minor editorial issues.

lib/mcwamp.cpp
AlexVlx: Somehow loading 2 different versions of the HCC runtime is a bug, and you have more problems than this. We don't provide a static library (anymore, thankfully) so it's unclear exactly how applications "somehow" statically link in a supported way. We'd need to actually provide an archive / static version of the HCC RT, which is not yet / no longer provided. So this should be dealt with there, since it's in and of itself a different kettle of fish. This has nothing to do with `once_flag` and everything to do with how internal linkage works. You could use and `atomic_flag`, or just a static `bool`, and the challenges for those two unsupported scenarios would be exactly the same.

scchan: Fixing potential issues exposed by a static version of HCC RT (we don't have one) is going to be future work unrelated to this PR.

aaronenyeshi (approved):

mangupta:
This PR is causing the following HIP unit tests to fail:

         89 - directed_tests/deviceLib/hipDoublePrecisionMathDevice.tst (OTHER_FAULT)
         91 - directed_tests/deviceLib/hipSinglePrecisionIntrinsics.tst (OTHER_FAULT)
         95 - directed_tests/deviceLib/hipSinglePrecisionMathDevice.tst (OTHER_FAULT)
        103 - directed_tests/deviceLib/hipDoublePrecisionIntrinsics.tst (OTHER_FAULT)
        105 - directed_tests/deviceLib/hipIntegerIntrinsics.tst (OTHER_FAULT)
        115 - directed_tests/kernel/hipEmptyKernel.tst (OTHER_FAULT)

When I run any one of these failed tests directly, I see the following output:
```
$  ./directed_tests/kernel/hipEmptyKernel
terminate called after throwing an instance of 'std::runtime_error'
  what():  No device code available for function: _Z5Emptyi
Aborted (core dumped)
```
If roll back HCC to the commit prior to this PR, all the tests pass. Do we need any changes to be done in the HIP unit tests? Looking at the changes in this PR, I don't think it should have impacted these tests. The one thing I notice is that all the failing tests have kernels which do no real work.

scchan:
@mangupta thanks for the information, I'll revert the lazy init commit for now and investigate

PR #1041, opened 2019-02-15
activity prof optimizing - read callbacks params only if enabled

eshcherb:


lib/hsa/activity_prof.h
```diff
@@ -113,27 +113,45 @@ class ActivityProf {
...
+        if (_enabled == true) {
+            const command_id_t command_id = obj->getCommandKind();
+            uint64_t start = obj->getBeginTimestamp();
+            uint64_t end   = obj->getEndTimestamp();
+            callback(command_id, start, end, bytes);
+        }
+    }
+
+    template <class T>
+    inline void callback_sys(T* obj, const size_t& bytes = 0) {
```
scchan: These two templates should have a more meaningful name and they aren't the actual callback functions.  How about renaming them to `report_gpu_timestamps` and `report_system_ticks` respectively?

scchan (changes requested):
I only have a nitpick on the names of the `callback_*` templates, other LGTM

scchan (approved):
LGTM

PR #1046, opened 2019-02-16
Deferred cleanup of HSAOps.

jeffdaily:
Treat asyncOps vector like a circular buffer.  New ops overwrite oldest ops, implicitly waiting for the old ops to complete and reclaiming resources.

PR #1048, opened 2019-02-20
fix potential deadlock triggered by a cleanup of the asyncops

scchan:
This fixes a deadlock introduced by https://github.com/RadeonOpenCompute/hcc/pull/1013 .  The important thing is avoid locking a queue while waiting on the completion of an event.
https://github.com/RadeonOpenCompute/hcc/pull/1046 will be the next step will providing improvements on how the asyncops vector is being used and managed.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -1714,8 +1714,9 @@ public:
                      });
         }
 
-        waitForStreamDeps(dispatch);
+        std::lock_guard<std::recursive_mutex> lg(qmutex);
 
+        waitForStreamDeps(dispatch);
 
         // create a shared_ptr instance
         std::shared_ptr<KalmarAsyncOp> sp_dispatch(dispatch);
@@ -2070,7 +2071,7 @@ public:
         STATUS_CHECK(status, __LINE__);
 
         // associate the barrier with this queue
-        pushAsyncOp(std::move(barrier));
+        pushAsyncOp(barrier);
 
         return barrier;
     }
@@ -2153,7 +2154,7 @@ public:
             STATUS_CHECK(status, __LINE__);
 
             // associate the barrier with this queue
-            pushAsyncOp(std::move(barrier));
+            pushAsyncOp(barrier);
 
             return barrier;
         } else {
@@ -4226,6 +4227,8 @@ std::shared_ptr<KalmarAsyncOp> HSAQueue::EnqueueAsyncCopyExt(const void* src, vo
     const Kalmar::HSADevice *copyDeviceHsa = static_cast<const Kalmar::HSADevice*> (copyDevice);
     std::shared_ptr<HSACopy> copyCommand = std::make_shared<HSACopy>(this, src, dst, size_bytes);
 
+    std::lock_guard<std::recursive_mutex> lg(qmutex);
+
     // euqueue the async copy command
     status = copyCommand.get()->enqueueAsyncCopyCommand(copyDeviceHsa, srcPtrInfo, dstPtrInfo);
     STATUS_CHECK(status, __LINE__);
@@ -4246,6 +4249,8 @@ std::shared_ptr<KalmarAsyncOp> HSAQueue::EnqueueAsyncCopy2dExt(const void* src, 
     //create shared_ptr instance
     const Kalmar::HSADevice *copy2dDeviceHsa = static_cast<const Kalmar::HSADevice*> (copyDevice);
     std::shared_ptr<HSACopy> copy2dCommand = std::make_shared<HSACopy>(this, src, dst, width*height);
+
+    std::lock_guard<std::recursive_mutex> lg(qmutex);
 
     //euqueue the async copy command
     status = copy2dCommand.get()->enqueueAsyncCopy2dCommand(width, height, srcPitch, dstPitch, copy2dDeviceHsa, srcPtrInfo, dstPtrInfo);
@@ -4297,6 +4302,8 @@ std::shared_ptr<KalmarAsyncOp> HSAQueue::EnqueueAsyncCopy(const void *src, void 
         copyDevice = nullptr; // H2H
     }
 
+    std::lock_guard<std::recursive_mutex> lg(qmutex);
+
     // enqueue the async copy command
     status = copyCommand.get()->enqueueAsyncCopyCommand(copyDevice, srcPtrInfo, dstPtrInfo);
     STATUS_CHECK(status, __LINE__);
@@ -4330,6 +4337,8 @@ HSAQueue::dispatch_hsa_kernel(const hsa_kernel_dispatch_packet_t *aql,
 
 
     Kalmar::HSADevice* device = static_cast<Kalmar::HSADevice*>(this->getDev());
+
+    std::lock_guard<std::recursive_mutex> lg(qmutex);
 
     std::shared_ptr<HSADispatch> sp_dispatch = std::make_shared<HSADispatch>(device, this/*queue*/, nullptr, aql);
     if (HCC_OPT_FLUSH) {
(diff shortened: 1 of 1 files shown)


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1590,6 +1590,33 @@ class HSAQueue final : public KalmarQueue
...
+        std::shared_ptr<HSAOp> r;
+        {
+          bool found = false;
+          std::lock_guard<std::recursive_mutex> lg(qmutex);
+          for (int i = 0; i < asyncOps.size(); i++) {
+            if (asyncOps[i] != nullptr) {
+              // wait on valid futures only
+              std::shared_future<void>* future = asyncOps[i]->getFuture();
+              if (future && future->valid()) {
+                r = std::move(asyncOps[i]);
```
david-salinas, line 1614: Do we need to wait on this op to complete before we remove it?  Do we need a future->wait() here ?

whchung: since @scchan put a `while(1)` loop here I believe his intention is to let those ops complete naturally without explicitly `wait()` on them?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1590,6 +1590,33 @@ class HSAQueue final : public KalmarQueue
     };
 
 
+    void drain() {
+      while (1) {
```
whchung: I believe comments are needed to clarify the semantic of `drain()`, especially there is a `while(1)` loop here.

lib/hsa/mcwamp_hsa.cpp
scchan: the `r.reset()` will destroy the HSAOp and will wait if it is not yet completed.  It occurs outside the lock_guard of the queue to prevent a deadlock.  It keeps going through all the ops until no valid op is found and at that point, it will shrink the asyncOps vector and return.

whchung (commented):
@scchan I think the semantic of `drain()` shall be properly commented.

Also I think the question from @david-salinas shall be clarified.

lib/hsa/mcwamp_hsa.cpp
whchung: Also if it's possible to forcibly abort `drain()` by letting the function check an flag?
I can foresee we may possible run into soft hang here in case the hardware goes wrong.

scchan: who will set the flag and how it will be set?

whchung: how about let the flag be an environment variable? users / sysadmins could set such an environment variable to break this loop and let the application abort, shall the application acts funky.

It could be an overkill though. I am just not used to see a `while(1)` loop in runtime without an explicit way to deterministically escape from the loop.

whchung (approved):

lib/hsa/mcwamp_hsa.cpp
jeffdaily: Is there any chance that another Op would be holding a shared_ptr to this Op such that the reset only decrements the ref count and does not destroy/wait on it?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1589,6 +1589,46 @@ class HSAQueue final : public KalmarQueue
...
+              // wait on valid futures only
+              std::shared_future<void>* future = asyncOps[i]->getFuture();
+              if (future && future->valid()) {
+                // move the op from the vector into a temp
+                // so that we can wait outside lock
+                r = std::move(asyncOps[i]);
+                found = true;
+                break;
+              }
+            }
```
jeffdaily, line 1618: Shouldn't there be a `break` if an Op is a nullptr?

jeffdaily:
I just built this PR and ran with TF CNN benchmark model resnet50 and it's showing a large performance regression on 4xMI25.  Down from ~710 images/sec down to ~640 images/sec.  I ran it again through linux perf tool and it shows `HSAQueue::drain()` is > 10% of execution time.  That's why I added a comment that perhaps we should break out of drain()'s Op search loop if a nullptr is found.  Isn't it still true that once a nullptr is found, the older Ops will also be nullptrs?  Otherwise, the drain loop will iterate through the entire asyncOps vector looking for a valid Op even when it is already drained.

jeffdaily:
Well, I tried adding the `break` as suggested, and I now get a seg fault, so that wasn't the right fix for the perf regression.

scchan:
@jeffdaily I think the problem is the loop is traversing in ascending order.  If I change it to descending order, it may work?

jeffdaily:
I just built your latest commit.  Same perf test as before is now up to ~670 images/sec.

jeffdaily:
@scchan the original code for HSAQueue::wait() would start at the youngest Op, so I think you're right to point out the drain() should also perform a descending search (with a break).

EDIT: Okay, adding a break for nullptr is just causing all kinds of seg faults.  But lets try the descending search.

jeffdaily:
@scchan what if instead of changing the loop to traverse in descending order, instead you don't restart at index 0 each time?  Is there a possibility that the asyncOps vector would change in size due to the wait on a found Op?  Something like:

```diff
diff --git a/lib/hsa/mcwamp_hsa.cpp b/lib/hsa/mcwamp_hsa.cpp
index dfd6097..2606801 100644
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -1598,13 +1598,19 @@ public:
       // keep going through the asyncOps vector until
       // no valid op is found
       bool found = false;
+      int i = 0;
+      size_t last_size = 0;
       do {
         std::shared_ptr<HSAOp> r;
         found = false;
         {
           // lock the queue and find the first valid op
           std::lock_guard<std::recursive_mutex> lg(qmutex);
-          for (int i = 0; i < asyncOps.size(); i++) {
+          if (last_size != asyncOps.size()) {
+              last_size = asyncOps.size();
+              i = 0;
+          }
+          for ( ; i < asyncOps.size(); i++) {
             if (asyncOps[i] != nullptr) {
               // wait on valid futures only
               std::shared_future<void>* future = asyncOps[i]->getFuture();
```

lib/hsa/mcwamp_hsa.cpp
scchan: updated, the exit condition is not controlled by whether a valid is found

scchan: updated, the exit condition is not controlled by whether a valid is found

scchan:
@jeffdaily I changed the traversal order, could you give it a try?
In the ascending case, not starting from zero is potentially racy since it won't know whether another thread has modified its state?

jeffdaily:
Inspecting younger ops got a seg fault.
```
#0  0x00007fff3779071f in ihipStream_t::locked_eventIsReady(ihipEvent_t*) () from /opt/rocm/hip/lib/libhip_hcc.so
#1  0x00007fff3780e4ba in hipEventQuery () from /opt/rocm/hip/lib/libhip_hcc.so
#2  0x00007fff461067c0 in stream_executor::gpu::GpuDriver::QueryEvent(stream_executor::gpu::GpuContext*, ihipEvent_t*)
    () from /usr/local/lib/python2.7/dist-packages/tensorflow/python/_pywrap_tensorflow_internal.so
```

scchan:
@jeffdaily I think I'm going to consider just having https://github.com/RadeonOpenCompute/hcc/pull/1046 to replace all the draining logic

whchung:
@jeffdaily do we have perf numbers for #1046?

jeffdaily:
Yes, I have TF CNN perf results for #1046.  
[TF_hcc_perf_20190220.xlsx](https://github.com/RadeonOpenCompute/hcc/files/2890571/TF_hcc_perf_20190220.xlsx)
This is comparing rocm 2.1 using `--local_parameter_device=cpu` as the baseline.  No significant variations in performance between 2.1 and #1046, so no regressions but also no improvements.

scchan:
superseded by https://github.com/RadeonOpenCompute/hcc/pull/1046
Implement a simple fix to the current design of asyncops vector without affecting functionality and performance proves to be difficult.  Let's skip that move to the improved design in 1046

PR #1049, opened 2019-02-20
update tests to fully qualify names in hc::    

david-salinas:
caused by  https://github.com/llvm-mirror/clang/commit/264d6736fec09b767e55b8671afb84b7fe0d051e

ambiguous names need to be fully qualified with their namespace.  Turns out there are a lot of instances where we do not do this in our tests.  The incoming upstream clang change exposes this, causing our tests to fail.  So we need to update them before we can merge this upstream change.


scchan (approved):
LGTM

PR #1050, opened 2019-02-21
use the proper triple in clamp-device

scchan:


scchan:
dup of https://github.com/RadeonOpenCompute/hcc/pull/1050

aaronenyeshi (commented):
I think this can be useful to set common TARGET_TRIPLE variable

PR #1054, opened 2019-02-26
Don't override CMAKE_CXX_FLAGS when building hcc-config

tstellar:
This fixes the build of hcc-config if a user configures with
-DCMAKE_CXX_FLAGS="-fPIC" and may prevent other similar kinds of
failures.


scchan (approved):
LGTM

PR #1060, opened 2019-02-28
improve HSAQueue::wait() performance

jeffdaily:
Add a marker to HSAQueue::wait() in case a valid future is never found, so that the next wait() will stop its search after fewer iterations.  This new marker caused a deadlock during HSAQueue::dispose() that was resolved by removing the wait() inside dispose().

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -1624,9 +1624,22 @@ public:
                     }
                     future = nullptr;
                 }
+                else {
+                    // asyncOps queue is initialized to all nullptr.
+                    // until the queue history fills up, this condition will occur
+                    break;
+                }
             }
         }
         if (future) future->wait();
+        else {
+            DBOUTL(DB_CMD2, "HSAQueue::wait() no future, enqueued marker");
+            auto marker = EnqueueMarker(hc::no_scope);
+            std::shared_future<void>* future = marker->getFuture();
+            if (future && future->valid()) {
+                future->wait();
+            }
+        }
     }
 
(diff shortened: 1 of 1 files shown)


jeffdaily:
@scchan Please review.  This PR fixes (and explains) why #1046 was slower than #1048 in some cases.

jeffdaily:
Initial benchmarking for TensorFlow, compared to #1046 and ROCm 2.1.


  | alexnet (1024) | googlenet (512) | inception3 (128) | inception4 (64) | resnet50 (128) | resnet152 (64) | vgg16 (128)
-- | -- | -- | -- | -- | -- | -- | --
ROCm2.1 | 1404.21 | 504.12 | 98.74 | 44.14 | 188.71 | 74.15 | 90.47
PR1046 | 1420.2 | 501.7 | 98.51 | 43.97 | 187.43 | 74.06 | 90.47
PR1060 | 1554.71 | 520.9 | 99.1 | 44.94 | 185.95 | 72.36 | 95.62

scchan:
@jeffdaily  I guess comparing to the previous asyncops implementation (e.g. vector, non-circular), the previous implementation will have from time to time a change to shrink the asyncops vector and therefore, making the number of ops to inspect much smaller.  
With the circular buffer, by enqueueing a new barrier removes the need to walk through the entire buffer.  
One potential issue I could think of is that if there isn't a rocrQueue associated to the HSAQueue object, then it will have to look for a different rocrQueue for enqueuing the new barrier packet but the probability of running into that is probably less often than the problem this PR is trying to solve

scchan (approved):
LGTM

PR #1061, opened 2019-03-04
Add error return for failed 2D SDMA copy

gargrahul:
Add error return for failed 2D SDMA copy to be used by HIP to fall back to 2D memcpy kernel.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -5462,10 +5462,8 @@ HSACopy::enqueueAsyncCopy2dCommand(size_t width, size_t height, size_t srcPitch,
         }
         
         status = hcc_memory_async_copy_rect(getCommandKind(), copyDevice, dstPtrInfo, srcPtrInfo, width, height, srcPitch, dstPitch, depSignalCnt, depSignalCnt ? &depSignal:NULL, _signal);
-        if(status != HSA_STATUS_SUCCESS) {
-            Kalmar::ctx.releaseSignal(_signal, _signalIndex);
+        if(status != HSA_STATUS_SUCCESS)
             return status;
-        }
     }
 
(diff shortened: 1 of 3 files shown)


gargrahul:
I have submitted required changes for this in HCC with https://github.com/RadeonOpenCompute/hcc/pull/1061

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -5456,14 +5460,15 @@ HSACopy::enqueueAsyncCopy2dCommand(size_t width, size_t height, size_t srcPitch,
                 << "  copyAgent=" << copyDevice
                 << "\n");
         }
-        isAsync = true;
         
-        hcc_memory_async_copy_rect(getCommandKind(), copyDevice, dstPtrInfo, srcPtrInfo, width, height, srcPitch, dstPitch, depSignalCnt, depSignalCnt ? &depSignal:NULL, _signal);
+        status = hcc_memory_async_copy_rect(getCommandKind(), copyDevice, dstPtrInfo, srcPtrInfo, width, height, srcPitch, dstPitch, depSignalCnt, depSignalCnt ? &depSignal:NULL, _signal);
+        if(status != HSA_STATUS_SUCCESS)
+            return status;
```
scchan, line 5466: This path needs to release the signal (see line 5427) before returning

scchan (changes requested):

lib/hsa/mcwamp_hsa.cpp
scchan: I took another look again.  I think my previous comment about the failing path needs to release the signal is wrong since the signal release would be taken care of by the dtor of HSACopy.  Could you revert your last commit?

scchan (approved):
LGTM

PR #1063, opened 2019-03-04
Set copy device prior to detecting stream deps.

jeffdaily:
Otherwise, a false dependency is detected even though the same copy engine is used.


jeffdaily:
@scchan please review.

scchan (approved):

PR #1066, opened 2019-03-05
Various asyncops fixes

scchan:
This fixes detectStreamDeps essentially by determining the copy device for the current first before analyzing the dependency.  It handles the case which the copy device is unknown to the hcc runtime.

This also adds a length of the dependency chain to HSAOp so that we could detect when the chain has become too long (e.g. 32 ops for now) and break it up by inserting a new marker.  This avoids the case where each enqueue op is dependent on its previous op and preventing resources from freeing up. 

There are a couple of fixes to make sure we delete the future before releasing a signal when disposing an HSAOp.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -870,13 +870,11 @@ public:
 private:
   hsa_status_t hcc_memory_async_copy(Kalmar::hcCommandKind copyKind, const Kalmar::HSADevice *copyDevice,
                                       const hc::AmPointerInfo &dstPtrInfo, const hc::AmPointerInfo &srcPtrInfo,
-                                      size_t sizeBytes, int depSignalCnt, const hsa_signal_t *depSignals,
-                                      hsa_signal_t completion_signal);
+                                      size_t sizeBytes);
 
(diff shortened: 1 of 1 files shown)


scchan:
@jeffdaily Could you take a look at this PR?  This is another PR related asyncops.  Thanks!

jeffdaily:
@scchan Any idea how performance is affected by this PR?  I think the first three commits were useful for fixing the seg fault we were seeing with the signal.  The dependency chain limit should catch latent bugs related to long dep chains, but I wonder if this affects performance since we would be inserting markers.  Does changing the depth from the current 32 to something else affect performance?  How did you arrive at 32?  Perhaps some fraction of the HCC_ASYNCOPS_SIZE should be the depth limit?

scchan:
@jeffdaily The dependency chain limit addresses a corner case where a queue is used for a huge amount of back-to-back P2P transfers, which the hcc runtime has no knowledge on which device was used for the copy.  The max value of 32 is just a SWAG :).  I tried caffe2 resnet50 with 2 and 4 GPU training and then changed the limit between 8 to 256 and I didn't observe any meaningful difference in performance.

jeffdaily:
LGTM

scchan:
rebased and resolved conflict with https://github.com/RadeonOpenCompute/hcc/pull/1061

PR #1068, opened 2019-03-06
Fix std::regex in printf causing bad_cast

aaronenyeshi:
In hc_printf.hpp, static const std::regex specifierPattern(... is causing std::bad_cast. Let's guard this with a macro so that preprocessor can optimize it out by default.

Issue: Certain user env cannot process hc_printf.hpp's std::regex syntax
Fix: Use macro HCC_ENABLE_ACCELERATOR_PRINTF to exclude regex and processPrintfPackets


scchan (approved):
LGTM

PR #1073, opened 2019-03-11
Fix Assembler script for early finalize and FCS

aaronenyeshi:
This fixes the missing function call flag for assembler code. 

we need https://github.com/RadeonOpenCompute/hcc-clang-upgrade/pull/172 first


scchan:
merged the clang patch, could update this to pick up a newer clang commit?

aaronenyeshi:
Okay I've updated the clang submodule to include the clang commit.

PR #1075, opened 2019-03-12
Add support for allocating finegrained GPU memory

mangupta:

PR #1076, opened 2019-03-13
Add support for allocating finegrained GPU memory

mangupta:


mangupta:
Replaces #1075 which had a rebase issue.

scchan (dismissed):
LGTM

scchan (changes requested):
This PR is causing a lot of unit tests to fail, could you validate this patch with `make test` and provide test results?  Thanks
http://hsautil.amd.com/job/RadeonOpenCompute/job/hcc/job/PR-1076/3/consoleFull

mangupta:
All the tests fail due to "No suitable runtime detected. Fall back to CPU!". However I don't see this on my development system. From what I see in the compiler source, this error is reported when we fail to detect ROCr. I will try debugging more on my end by trying on another system as well as trying inside a docker container.

scchan (approved):

PR #1080, opened 2019-03-14
Move amdgcn intrinsics from hc.amdgcn.ll to hc.hpp

david-salinas:
This is a minimal fix to get the known issues resolved. For a complete fix, we'd move all of the intrinsics in hc.amdgcn.ll from device-libs here.

### DIFF
--- a/include/hc.hpp
+++ b/include/hc.hpp
@@ -2976,7 +2976,7 @@ inline float __amdgcn_ds_permute(int index, float src) [[hc]] {
 /**
  * ds_swizzle intrinsic
  */
-extern "C" int __amdgcn_ds_swizzle(int src, int pattern) [[hc]];
+extern "C" int __amdgcn_ds_swizzle(int src, int pattern) [[hc]] __asm("llvm.amdgcn.ds.swizzle");
 inline unsigned int __amdgcn_ds_swizzle(unsigned int src, int pattern) [[hc]] {
   __u tmp; tmp.u = src;
   tmp.i = __amdgcn_ds_swizzle(tmp.i, pattern);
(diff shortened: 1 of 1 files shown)


arsenm (changes requested):
This still isn't using the clang builtins. This should use the __builtin_amdgcn_* builtins, and not use asm declarations

b-sumner:
I agree with Matt. This should use the clang builtins, and switch to update.dpp from mov.dpp.

david-salinas:
My latest commit, is just to get the known failing tests passing.  I'm working on the remaining functions in hc.amdgcn.ll, which I'll have early next week.

david-salinas:
@b-sumner and @arsenm , so using the builtins is an issue for HCC.  Because HCC invokes clang without specifying a  target, clang flags the use of the DPP builtins without enabling the DPP feature.  You can see the errors in the HCC build/test Check for this PR: http://hsautil.amd.com/job/RadeonOpenCompute/job/hcc/job/PR-1080/4/console

arsenm:
You need to place the +dpp feature on the calling function, e.g. __attribute__((target("dpp")))

b-sumner:
Sorry, I keep overlooking this serious problem with hcc, and I shouldn't.  We should probably go back and just make the minimal changes to the .ll file or wherever that fix the original problem.

david-salinas:
@scchan @b-sumner @arsenm 
ok, I updated the fix to use __asm again for llvm.amdgcn.update and llvm.amdgcn.ds.swizzle.  When I tried addind the attribute((target("dpp"))) I got a warning indicating that "dpp" is invalid for the target attribute, and it would be ignored.  The check just ran and all tests passed.

scchan (approved):
I think we'll settle with this workaround for now.  LGTM

PR #1094, opened 2019-03-25
Factor in sram-ecc and xnack ELF constants for ISA compat check

david-salinas:
Account for SRAM-ECC and XNACK when finding compatible Code Objects.

Change-Id: I2277030312815e0be8f8850fb14fc3bdfdd96ec1


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2657,6 +2657,7 @@ class HSADevice final : public KalmarDevice
     }
 
     bool IsCompatibleKernel(void* size, void* source) override {
+
```
AlexVlx: Why?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2686,7 +2687,12 @@ class HSADevice final : public KalmarDevice
             case hc::EF_AMDGPU_MACH_AMDGCN_GFX906 : triple.append("906"); break;
         }
 
-        const auto isa{get_isa_name_from_triple(std::move(triple))};
+        unsigned SRAM_ECC_ENABLED = reader.get_flags() & hc::EF_AMDGPU_SRAM_ECC;
```
AlexVlx: All caps is for macros, this isn't a macro.

AlexVlx (changes requested):

lib/hsa/mcwamp_hsa.cpp
david-salinas: let me clean that up ... one sec.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2686,7 +2686,12 @@ class HSADevice final : public KalmarDevice
...
+        unsigned XNACK_ENABLED = reader.get_flags() & hc::EF_AMDGPU_XNACK;
+        if (SRAM_ECC_ENABLED) isa.append("+sram-ecc");
+        if (XNACK_ENABLED) isa.append("+xnack");
```
scchan: Does this even work?  `isa` are used here but only gets defined 2 lines below

scchan (changes requested):

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2686,7 +2686,12 @@ class HSADevice final : public KalmarDevice
             case hc::EF_AMDGPU_MACH_AMDGCN_GFX906 : triple.append("906"); break;
         }
 
-        const auto isa{get_isa_name_from_triple(std::move(triple))};
+        unsigned sram_ecc_enabled = reader.get_flags() & hc::EF_AMDGPU_SRAM_ECC;
+        unsigned xnack_enabled = reader.get_flags() & hc::EF_AMDGPU_XNACK;
+        if (sram_ecc_enabled) triple.append("+sram-ecc");
```
AlexVlx: I believe the order is flipped here, XNACK comes before ECC, unless ROCr decided to be special about it.

lib/hsa/mcwamp_hsa.cpp
david-salinas: sorry.  I corrected this in the latest commit

lib/hsa/mcwamp_hsa.cpp
david-salinas: corrected in latest commit

lib/hsa/mcwamp_hsa.cpp
david-salinas: yup.  rocr currently has xnack first.

scchan:
@david-salinas some of the tests failed with `### HCC STATUS_CHECK Error: HSA_STATUS_ERROR_INVALID_ISA_NAME (0x1017) at file:mcwamp_hsa.cpp line:2724`, do we know the reason?

david-salinas:
> @david-salinas some of the tests failed with `### HCC STATUS_CHECK Error: HSA_STATUS_ERROR_INVALID_ISA_NAME (0x1017) at file:mcwamp_hsa.cpp line:2724`, do we know the reason?

@scchan so, I think it might just be that we're using an old ROCR.  This change only adds the feature flag strings to the triple iff the ELF flag is set.  Which could happen for some gfx###.  So, I'm trying to see if I can get an updated ROCR installed ... but I think we may need to do that manually.

scchan:
@david-salinas have you tested locally on your machine and what's the rocr version?

david-salinas:
> @david-salinas have you tested locally on your machine and what's the rocr version?

@scchan all of the tests pass when I run them outside of the docker on the same validation machine.  And that machine has hsa-rocr-dev 1.1.9-55-gbac2a9b installed

scchan:
@david-salinas are you planning update the docker image and restart the test?

david-salinas:
@scchan I think we have a chicken-egg problem.  This change will only work with ToT rocr.  I may need to add code to first try to validate the triple with the feature flags added, if it is not considered valid, then to leave them off ?  And down the road, we just remove that extra step ?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2708,14 +2708,28 @@ class HSADevice final : public KalmarDevice
...
 
         struct isa_comp_data {
             hsa_isa_t isa_type;
             bool is_compatible;
         } co_data;
 
-        status = hsa_isa_from_name(isa.c_str(), &co_data.isa_type);
+        status = hsa_isa_from_name(triple.c_str(), &co_data.isa_type);
+
+        if (status != HSA_STATUS_SUCCESS) {
```
AlexVlx: I don't quite see what the intent is here. I suspect that we do not want to support loading newer code objects on older RTs, and even if we do this will absolutely do the wrong thing, because you'd load the wrong thing. Furthermore, surely you don't want `else if`...

david-salinas: It's a kludge.  I'm not happy with it, but unless there is some way for HCC to validate the version of the RT, I'm not sure what else we can do.  

And, the "else if" is intended.  The std::erase will erase from the pos to the end of the string.  And since "+xnack" is first, we don't need to erase anymore, regardless of whether sram-ecc is set or not.  Honestly, my opinion is that none of this code should be in HCC/HIP.  It should all be contained in ROCR.  We should just be passing an ELF file to ROCR and getting back the isa_type.

AlexVlx: You can do "not this", which swallows errors assuming that they can only be due to mismatched RTs, and lies about the nature of the code that's in a code object. What are you actually trying to solve with this? Have you verified that it does anything for the problem that you are trying to solve?

david-salinas: ok, yeah I can change the "if (status != HSA_STATUS_SUCCESS)".  So, I've verified that this works with a RT that isn't expecting the feature flags.  I'm working on verifying it on a system with a RT that does expect the feature strings; which was the original problem.

AlexVlx: It cannot work on a RT that does not expect the feature flags, because at that point you are lying about what is in the code object. So at best you have an odd mismatch.

david-salinas: True.  This code, to remove the feature strings, is intended to be temporary.  We cannot commit the necessary changes to HCC/HIP without a RT that expects them.  And we cannot commit changes to the RT without a HCC/HIP that gives them.

david-salinas: So i've validated that this fix works with a RT expecting the feature string, as well as a RT not expecting the feature string.  My proposal is that (after I fix the top "if"), we use this until we get a RT in our build/test envs that expects the feature string, then we can remove the code to handle the case of a "old" RT.

Issue #1092, opened 2019-03-22
LIBCXX_HEADER paths in CmakeLists.txt


jithunnair-amd:
Are these paths in the CMakeLists.txt not a bit too restrictive? For eg. I had the headers installed but they were in /usr/include/c++/5 (I have gcc-5 installed). Creating a soft link from /usr/include/c++/5 -> /usr/include/c++/v1 helped as a workaround, but is that the proper solution?

```
# Detect libc++
if (USE_LIBCXX)
  find_path(LIBCXX_HEADER random PATHS /usr/local/include/c++/v1 /usr/include/c++/v1 NO_DEFAULT_PATH)
  message( "libc++ headers found at ${LIBCXX_HEADER}" )
endif( )
```

scchan:
If you are not trying to switch to the libc++ from llvm, could you try `cmake -DUSE_LIBCXX=OFF`?

jithunnair-amd:
While `cmake -DUSE_LIBCXX=OFF` does indeed work, the issue I was trying to highlight was the hardcoded path that the cmake file looks into for libc++ headers, specifically the `/v1` part. Is that expectation always correct?

scchan:
LIBCXX is a different C++ runtime implementation and that is different than the one used by GNU g++ by default.  The `v1` part is the location for the llvm implementation, which we no longer support.  The issue why you get into this is that there's some logic in the cmake to try to detect which C++ runtime implementation to use and that was confused by using clang as the build compiler.  Since we don't need/support that, I'll submit a patch to remove it.

PR #1096, opened 2019-03-27
Remove support for libc++

scchan:
hcc no longer maintains the support for using libc++ as the C++ runtime, remove it from the code to avoid unwanted side effects such as https://github.com/RadeonOpenCompute/hcc/issues/1092


AlexVlx (approved):
LGTM, thanks.

PR #1099, opened 2019-03-29
Remove the symbol  _binary_kernel_bundle_start from the runtime.

scchan:
This fixes the issue with dlopen() not being able to load
libmcwamp.so if the external symbol _binary_kernel_bundle_start
doesn't exist.

### DIFF
--- a/lib/mcwamp.cpp
+++ b/lib/mcwamp.cpp
@@ -7,6 +7,7 @@
 
 #include "hc_rt_debug.h"
 #include "mcwamp_impl.hpp"
+#include "../hc2/external/elfio/elfio.hpp"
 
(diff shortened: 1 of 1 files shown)


AlexVlx (approved):
LGTM, thanks. Do note that if we have more profound concerns here we'll probably have to do the same thing we did with HIP (inline functions with hidden visibility) for all of the state stuff. However, I don't think that there are use cases for that, so we probably can get away with it.

scchan:
I know this logic in hcc will go away eventually since hip rt has its own thing so I didn't bother doing a major surgery at this point for the hidden visibility or migrate this to use the program_state stuff

PR #1100, opened 2019-03-30
append feature flags to target triple for rt isa check

david-salinas:
Change-Id: I1285cf6a99b4ab534f049c820cb1bbbdb26259f9

I messed up the last PR (https://github.com/RadeonOpenCompute/hcc/pull/1094)

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -2706,6 +2706,11 @@ public:
             case hc::EF_AMDGPU_MACH_AMDGCN_GFX906 : triple.append("906"); break;
         }
 
+        unsigned sram_ecc_enabled = reader.get_flags() & hc::EF_AMDGPU_SRAM_ECC;
+        unsigned xnack_enabled = reader.get_flags() & hc::EF_AMDGPU_XNACK;
+        if (xnack_enabled) triple.append("+xnack");
+        if (sram_ecc_enabled) triple.append("+sram-ecc");
+
         const auto isa{get_isa_name_from_triple(std::move(triple))};
 
         struct isa_comp_data {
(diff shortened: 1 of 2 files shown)


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2714,7 +2719,9 @@ class HSADevice final : public KalmarDevice
         } co_data;
 
         status = hsa_isa_from_name(isa.c_str(), &co_data.isa_type);
-        STATUS_CHECK(status, __LINE__);
+
+	if (status != HSA_STATUS_SUCCESS)
+		return false;
```
scchan: please fix indentation

scchan (changes requested):

lib/hsa/mcwamp_hsa.cpp
scchan: Also `kernel_source` needs to be freed before returning or turn that buffer into an automatic object

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2730,6 +2739,7 @@ class HSADevice final : public KalmarDevice
                     return HSA_STATUS_SUCCESS;
                 },
                 &co_data);
+
```
scchan: could you remove this?

lib/hsa/mcwamp_hsa.cpp
david-salinas: done

lib/hsa/mcwamp_hsa.cpp
david-salinas: done

scchan (approved):
LGTM

PR #1109, opened 2019-04-10
add LC option -mattr=+sram-ecc for gfx906 only

david-salinas:
Change-Id: Icc484999e09a3f5303a2e820438c2f9b28ff3633


scchan (approved):
LGTM

PR #1111, opened 2019-04-11
Fix kernel name in profiling data

scchan:
If a kernel is launched from HIP, HCC RT can't depend on the kernel name string from HIP to be always available since its lifetime is unknown.  This cause the profiling data to output invalid kernel names.  The fix is to make a copy of the kernel name if the kernel is launched from HIP.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -1008,7 +1008,7 @@ class HSADispatch : public HSAOp {
 private:
     Kalmar::HSADevice* device;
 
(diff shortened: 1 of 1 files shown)


scchan:
@jeffdaily this addresses the kernel name issue in profiling data

jeffdaily:
I cherry-picked and built this PR against ROCm 2.2 and confirmed it fixes the issue of garbage names when using HCC_PROFILE=2.  Thank you, @scchan .

scchan:
@jeffdaily thanks

PR #1116, opened 2019-04-16
add amd_comgr to auto upstream merge docker file, since we build HIP

david-salinas:


aaronenyeshi (approved):
Thanks for adding this. LGTM

Issue #1056, opened 2019-02-26
libmcwamp_hsa.so has an executable stack


tstellar:
libmcwamp_hsa.so is built with an executable stack:

readelf -l libmcwamp_hsa.so | grep GNU_STACK -A 1
GNU_STACK      0x0000000000000000 0x0000000000000000 0x0000000000000000
                           0x0000000000000000 0x0000000000000000  RWE    0x10

I think this is happening because mcwamp_hsa.cpp.o hsa a note.GNU-stack section that is marked
executable.  Is this being done intentionally?  I haven't been able to figure out what is generating this section.

 readelf -S lib/hsa/CMakeFiles/mcwamp_hsa.dir/mcwamp_hsa.cpp.o | grep note.GNU-stack -A 1
  [29] .note.GNU-stack   PROGBITS         0000000000000000  00280aaa
                                                                 0000000000000000  0000000000000000   X       0     0     1

david-salinas:
@AlexVlx @scchan @whchung Do any of you recall if we need to have an exec stack in mcwamp.o?  I can try to add -noexecstack to our build of mcwamp.cpp and see if all still works.

david-salinas:
@tstellar This just seems like the default (on ubuntu for ELF), unless we explicitly remove the exec stack from mcwamp.o.    Do you have concerns with the exec stack being there?

FelixSchwarz:
From https://fedoraproject.org/wiki/Packaging_tricks#Executable_stack 
> rpmlint will complain if you generate code that permits stack execution, which is justified since this is rarely needed and is often used by attackers to exploit vulnerabilities.

tstellar:
@david-salinas As @FelixSchwarz mentioned we prefer not to ship executable stacks in Fedora, but I guess my question is less about libmcwamp_hsa.so having an executable stack and more about why does hcc compile objects with an executable stack by default.  I don't think this is Ubuntu specific since it also happens on Fedora.

david-salinas:
@tstellar sorry for the delay.  So, I believe this is an artifact of default Clang behaviour.  I'm trying to see if this can be disabled by default.

david-salinas:
[...]

PR #1126, opened 2019-04-25
remove executable stack from hcc shared libraries

david-salinas:
Change-Id: Ib3d10b0cfb97cf737f571c49156c304d92c2401e
for issue #1056

### DIFF
--- a/scripts/cmake/MCWAMP.cmake
+++ b/scripts/cmake/MCWAMP.cmake
@@ -6,6 +6,7 @@ endif()
 # gtest
 set(GTEST_SRC_DIR "${PROJECT_SOURCE_DIR}/utils")
 set(GTEST_INC_DIR "${PROJECT_SOURCE_DIR}/utils")
+set(CMAKE_SHARED_LINKER_FLAGS "-Wl,-z,noexecstack")
 
 # MCWAMP
 set(MCWAMP_INC_DIR "${PROJECT_SOURCE_DIR}/include")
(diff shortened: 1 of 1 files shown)


scripts/cmake/MCWAMP.cmake
```diff
@@ -27,7 +28,7 @@ macro(amp_target name )
   target_include_directories(${name} PRIVATE ${MCWAMP_INC_DIR})
   target_include_directories(${name} SYSTEM INTERFACE $<INSTALL_INTERFACE:$<INSTALL_PREFIX>/include>)
   target_compile_options(${name} PUBLIC -hc -fPIC)
-
+
```
aaronenyeshi: there was a random trailing space added here

aaronenyeshi (commented):
What is the context for this change?

scripts/cmake/MCWAMP.cmake
david-salinas: doh.  ok will remove.  And there is another one at line 71.

david-salinas:
> What is the context for this change?

@aaronenyeshi the context is that by default when we build libmcwamp_hsa.so, we leave in an executable stack in the SO.  Issue #1056 is the original complaint.  This is actually, just default behaviour for clang and the linker.  But because our SO gets linked in to user/customer executables, this executable section can introduce a security vulnerability.  So we can just add the linker option "noexecstack" to remove the "X" (execute) permission from the "GNU_STACK" entry in the elf.

scchan:
@david-salinas I don't think this PR addresses #1056 since this only addresses the shared libraries produced by HCC's cmake file.  #1056's concern is about objects produced by HCC so I think the fix should be somewhere's HCC's linker script.

PR #1131, opened 2019-04-29
expose an opaque pointer to accelerator::pDev

ssahasra:
Multiple accelerators can wrap the same pDev, and the equality
operator uses this to determine if two accelerators are "the
same". Exposing this as an opaque pointer allows clients to use the
pDev as a key in std::map etc when tracker per-accelerator
information.

This is used in HIP hostcall to track hostcall buffers that are
allocated per HIP device.


scchan (approved):
LGTM

PR #1132, opened 2019-04-29
remove accelerator_view::get_hsa_queue()

ssahasra:
This function returns the underlying hsa_queue_t for an
accelerator_view, but this mapping is not stable in the presence of
queue stealing. It was introduced for HIP, but it is not used
anywhere, and it is unsafe anyway.


scchan (changes requested):
Could you remove this test?

Failing Tests (1):
    HCC :: Unit/HC/test1.cpp

ssahasra:
> Could you remove this test?
> 
> Failing Tests (1):
> HCC :: Unit/HC/test1.cpp

Done.

PR #1134, opened 2019-04-29
remove signal pool

jeffdaily:
ROCR signal implementation has changed since needing the signal pool in the HCC runtime.  It was suggested by @skeelyamd to remove the HCC signal pool.  Some evidence to bolster the suggestion is a trace of HSA and HIP calls, showing that when the HCC signal pool must grow (by the default of 512), it will create a gap in kernel execution due to the overhead of growing the HCC signal pool.  Although the overhead of the pool is relatively small, removing the signal pool from HCC will simplify the code.

image

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -66,7 +66,6 @@
 #define KERNARG_BUFFER_SIZE (512)
 
 // number of pre-allocated kernarg buffers in HSAContext
-// Not required but typically should be greater than HCC_SIGNAL_POOL_SIZE 
 // (some kernels don't allocate signals but nearly all need kernargs)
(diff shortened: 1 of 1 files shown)


scchan (commented):
LGTM
What tests have we validate this with?

jeffdaily:
Excellent question.  I apologize for not being prepared with a test to validate this PR; I was simply running our TensorFlow release to confirm that this PR worked.  My motivation for this PR was not because this was a pressing issue with huge correctness or performance implications, but rather to not lose the knowledge exchanged with @skeelyamd about the latest signal implementation.

@scchan I echo your question -- are there existing tests for the HCC runtime that would stress signal creation/destruction?

scchan (approved):

PR #1135, opened 2019-05-01
don't add dependency if the last op has been completed

scchan:
When trying to detect stream dependency, check whether the last op and if it has been completed, then there's no need to report it as a dependency.  This check would avoid creating an extremely long dependency chain due to false dependencies.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -1533,7 +1533,9 @@ public:
             }
 
             if (needDep) {
-                if (hsa_signal_load_scacquire(*static_cast<hsa_signal_t*>(lastOp->getNativeHandle()))==0) {
+                hsa_signal_t* s = static_cast<hsa_signal_t*>(lastOp->getNativeHandle());
+                if (s->handle == 0 ||
+                    hsa_signal_load_scacquire(*s)==0) {
                     // if the last op has already been completed then don't need a dependency
                     return nullptr;
                 }
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -1529,10 +1529,15 @@ public:
                 if (FORCE_SIGNAL_DEP_BETWEEN_COPIES) {
                     DBOUT(DB_CMD2, "Set NeedDep for " << newOp << "(FORCE_SIGNAL_DEP_BETWEEN_COPIES) " );
                     needDep = true;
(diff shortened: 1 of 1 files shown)


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1529,10 +1529,15 @@ class HSAQueue final : public KalmarQueue
                 if (FORCE_SIGNAL_DEP_BETWEEN_COPIES) {
                     DBOUT(DB_CMD2, "Set NeedDep for " << newOp << "(FORCE_SIGNAL_DEP_BETWEEN_COPIES) " );
                     needDep = true;
+                    return lastOp;
```
jeffdaily: Why would we return lastOp here?  Why would we force the signal dep between copies even if the signal is already complete (as determined by line 1537 below)?

scchan: Fixed

jeffdaily (approved):

jeffdaily:
I just noticed one more thing -- you don't check `if (signal.handle)` prior to the load.  I think either case is a valid case to return `nullptr`.

jeffdaily (changes requested):
I just noticed one more thing -- you don't check if (signal.handle) prior to the load. I think either case is a valid case to return nullptr.

Also, should we move the `depAsyncOp = nullptr` (line 5680) as well as lines 5144 through 5147 (clears depAsyncOps in HSABarrier) out of the ::dispose() methods and into the waitComplete() methods?  It seems to me that if the HSAOps (copy and barrier) that have dependent ops/signals are completed, they should release the dep ops immediately instead of waiting for destruction.

scchan:
> I just noticed one more thing -- you don't check if (signal.handle) prior to the load. I think either case is a valid case to return nullptr.
> 
> Also, should we move the `depAsyncOp = nullptr` (line 5680) as well as lines 5144 through 5147 (clears depAsyncOps in HSABarrier) out of the ::dispose() methods and into the waitComplete() methods? It seems to me that if the HSAOps (copy and barrier) that have dependent ops/signals are completed, they should release the dep ops immediately instead of waiting for destruction.

Good catch, I added a check to ensure the signal is valid.
Actually, I considered taking PR#1134 and then wrap the signal around a shared_ptr.  In the HSAOp, instead of recording the dependencies as HSAOps, we should just record the shared_ptr<signal> instead.

jeffdaily (approved):

PR #1137, opened 2019-05-02
when finished waiting, decrement the ref count on the depenendent op

scchan:
as suggested by https://github.com/RadeonOpenCompute/hcc/pull/1135#pullrequestreview-232778635


jeffdaily (approved):
Exactly the change I suggested.  Hopefully this helps with dependency chains and the growing signal pool, as well.  Though I wonder if we should reconsider the default HSAOp queue size...

scchan:
I thought about reducing hsaop queue size but I'd do one change at a time so that QA has more time to test the effect of each change

PR #1138, opened 2019-05-02
full lazy init of HSAContext

jeffdaily:
HSAContext is no longer a module static object to be initialized during C++ static initialization.
HCC_LAZYINIT=ON is now the default behavior.

These changes allow dependent libraries and frameworks to be able to call fork() et al prior to
initializing the underlying HSA and kernel runtimes since HCC intialization is deferred until first use.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -147,7 +147,7 @@ char * HCC_PROFILE_FILE=nullptr;
             sstream << "\t" << *op << ";";\
     }\
    sstream <<  msg << "\n";\
-   Kalmar::ctx.getHccProfileStream() << sstream.str();\
+   Kalmar::ctx->getHccProfileStream() << sstream.str();\
 }
 
 // Track a short thread-id, for debugging:
@@ -2227,6 +2227,7 @@ private:
     uint16_t workgroup_max_dim[3];
 
     std::map<std::string, HSAExecutable*> executables;
+    std::once_flag executables_loaded;
 
     hcAgentProfile profile;
 
@@ -2312,6 +2313,17 @@ public:
 
 
 private:
+
+    void load_executables_once() {
+        std::call_once(executables_loaded, [=]() {
+            if (executables.size() == 0) {
+                // get default queue on the device
+                std::shared_ptr<KalmarQueue> queue = get_default_queue();
+                // load kernels on the default queue for the device
+                CLAMP::LoadInMemoryProgram(queue.get());
+            }
+        });
+    }
 
     // NOTE: removeRocrQueue should only be called from HSAQueue::dispose
     // since there's an assumption on a specific locking sequence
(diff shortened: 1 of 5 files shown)


scchan (commented):
Would it be sufficient to set this to `false`?  https://github.com/RadeonOpenCompute/hcc/blob/clang_tot_upgrade/lib/mcwamp.cpp#L411

I had a PR to make that change but reverted it later due to a bug in HIP, which I think has been fixed.

jeffdaily:
I would have simply set line 411 to false, but I wasn't sure about apps or libraries that might depend on the current default behavior of the immediate (i.e. non-lazy) init.  I reversed all the logic so that users could still access the old behavior by setting HCC_LAZYINIT=ON or HCC_LAZYINIT=0.  (I just pushed a fix for the latter.)

jeffdaily:
Looks like I misunderstood the C++ static initialization order.   Setting HCC_LAZYINIT=ON was sufficient for full lazy initialization.  Should I revert the changes to the mcwamp_hsa.cpp file but leave the mcwamp.cpp changes?

jeffdaily:
@scchan Please review again.   The changes are now only to the mcwamp.cpp source file to enable lazy init by default.

scchan (approved):
LGTM

jeffdaily:
I am looking into the failed tests.  Such as 
```
terminate called after throwing an instance of 'Kalmar::runtime_exception'
  what():  HSA executable NOT built yet!
```

```
Failing Tests (5):
    HCC :: Unit/HC/accelerator_get_all_views_mt.cpp
    HCC :: Unit/HC/memcpy_symbol1.cpp
    HCC :: Unit/HC/memcpy_symbol2.cpp
    HCC :: Unit/HC/memcpy_symbol3.cpp
    HCC :: Unit/HC/memcpy_symbol4.cpp
```

jeffdaily:
Down to 1 failure.
```
Failing Tests (1):
    HCC :: Unit/HC/accelerator_get_all_views_mt.cpp
```

jeffdaily:
@scchan I expect all tests to pass now.  The final failing tested needed to be updated for the new lazy init default since it changes the number of accelerator views created.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -2321,6 +2322,16 @@ class HSADevice final : public KalmarDevice
 
 private:
 
+    void load_executables_once() {
+        if (!executables_loaded) {
```
scchan: I think this is racy and probably needs to be protected by something like `std::call_once` ?

scchan (changes requested):

jeffdaily:
I'm not sure how our CI is set up, but an unrelated commit triggered a rebuild of the PR-1138 jenkins build and it is now failing, even though it was passing on build 7 of this PR.

scchan (dismissed):
LGTM in principle but now this PR has a merge conflict with the on-demand signal allocation patch you submitted earlier, could you take a look?

jeffdaily:
@scchan I don't know that the failed test is due to the PR.  Should we have jenkins retest?  All tests were passing on my local system.

scchan (approved):
LGTM

PR #1145, opened 2019-05-08
locked access to underlying hsa_queue_t

ssahasra:
This allows a client like the HIP runtime to lock the HSA queue assigned to an accelerator_view, so that it is not stolen while the client depends on a stable mapping. The hostcall infrastructure needs this stability so that the HIP runtime can associate a hostcall buffer with an HSA queue when a kernel is being launched.

### DIFF
--- a/include/hc.hpp
+++ b/include/hc.hpp
@@ -493,6 +493,22 @@ public:
     }
 
     /**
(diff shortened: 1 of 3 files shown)


scchan (approved):
LGTM

PR #1147, opened 2019-05-10
Fix PR #1138. Unconditionally pin printf buffer.

jeffdaily:
PR #1138 introduced a bug where the printf buffer was no longer unconditionally pinned.


scchan (approved):
LGTM

PR #1156, opened 2019-05-17
HCC_PROFILE=2 was printing HSABarrier dep ops after free

jeffdaily:
@scchan , the eager free of dep ops causes seg fault with HCC_PROFILE=2.

This PR is a quick fix, but we should consider alternatives.

We should consider moving any tracing/logging code out of the HSAOp::dispose() functions and into the HSAOp::waitComplete() functions.  Otherwise, we do not get any HCC_PROFILE=2 output until clean program termination when the HSAOps vector destructs.


scchan (approved):
LGTM

PR #1173, opened 2019-06-07
do not allocate a signal for every kernel dispatch

jeffdaily:


david-salinas:
Hi @jeffdaily.  Just confirming, does this PR address the performance issue in rocBLAS?

jeffdaily:
I have not had a chance to build and test the rocblas test with this PR in place.  I will need some more time for that.  As for how this affects performance of the various frameworks, it seems it does not make much of a difference for better or worse at least for TensorFlow.  I have not tested PyTorch.

jeffdaily:
Last CI run failed with "no space left on device".

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1993,13 +2034,13 @@ class HSAQueue final : public KalmarQueue
...
         std::shared_ptr<HSABarrier> barrier = std::make_shared<HSABarrier>(this, 0, nullptr);
-        // associate the barrier with this queue
-        pushAsyncOp(barrier);
 
         // enqueue the barrier
         status = barrier.get()->enqueueAsync(release_scope);
         STATUS_CHECK(status, __LINE__);
 
+        // associate the barrier with this queue
+        pushAsyncOp(barrier);
```
scchan: Would it make sense to move all these `pushAsyncOp` into the various `enqueueAsync` functions (where the queue is locked) such that the copy of the AQL packet into the user mode queue (and among other things) and the cache of the async op are atomic?

scchan (changes requested):

lib/hsa/mcwamp_hsa.cpp
jeffdaily: Yes!  Agree completely.

jeffdaily: `hcc_memory_async_copy` and `hcc_memory_async_copy_rect` never lock the qmutex -- when is it correct to call `pushAsyncOp` in those cases?

scchan: I think qmutex should be locked on entrance of these 2 functions because the analysis of the sync optimization needs protection.  Also, the call to the hsa async copy function and the push of the async op into the vector need to be atomic.

jeffdaily:
@scchan my last commit(s) need some explaining.  

The goal was as we discussed, the `pushAsyncOp` should be atomic w.r.t. the AQL packet insertion.  I tried to treat `hcc_memory_async_copy` and `hcc_memory_async_copy_rect` similar to how `dispatchKernel` is handled.  `dispatchKernel` takes the `hsa_queue_t*` as an arg after locking the qmutex.  So the two copy functions are used in the same way, having the caller acquire the locked queue prior, rather than locking the `qmutex` inside.

I felt `pushAsyncOp` should be used as a contract of sorts -- it creates the `std::shared_ptr<KalmarAsyncOp>` when handed an `HSAOp*`, taking ownership of the pointer and inserting it into the `asyncOps`.  As part of this contract, the `shared_future` must also be created.  Otherwise, the `qmutex` would be released before the `future` was created, and this could lead to a race condition with #1220.  Basically, once the `HSAOp` is in the `asyncOps` vector, it *must* have a `shared_future` that other threads can safely wait for.

Small bookkeeping change - the `std::shared_future` was duplicated in all `HSAOp` subclasses, so I moved it into `HSAOp`.  

Another small bookkeeping change - the interface was cleaner by removing the `_wait_complete_status` member.  I added that back when I re-worked the `asyncOps` for rocm 2.3.  I found it was cleaner without it.  I'm sure there are other small fixes like this that aren't critical, but help with maintainability.

I would like to understand performance impacts of this change, since our critical sections are becoming larger, and we added critical sections to the async copies.  I am preparing a branch based on roc-2.6.x with these commits cherry-picked.

jeffdaily:
@scchan the synchronous copy and kernel ops never call `pushAsyncOp`, so they never participate in the usual steps such as setSeqNumFromQueue(), youngestCommandKind, etc.  
Is this a bug?

jeffdaily:
Closed in favor of #1347 .

PR #1180, opened 2019-06-17
create hsa queue lazily

scchan:


scchan:
@gargrahul  this prevents hipGetDeviceCount() (or some other HIP APIs) from eagerly creating HSA queues, could you review this?  Thanks

jeffdaily:
This PR looks relatively innocent.  Which issues is this addressing?  Have we validated against frameworks yet?  If you would like, I could try this out on TensorFlow to make sure nothing breaks.  Let me know.

gargrahul:
@jeffdaily This PR should help delaying of creation of HSA resources ONLY on first actual use. So it addresses issue of creating queues with first HIP call when there isn't any actual work being done with that call. For example hipGetDeviceCount => which is a query API.

jeffdaily:
Thank you @gargrahul .  Do we know whether any frameworks have been tested with this PR yet?

scchan:
@jeffdaily No I haven't but will try with Caffe2 and PyTorch tomorrow.

scchan:
my framework tests are still running but a lot of them has already been run without issue and I don't see any regression so far, merging this now

PR #1183, opened 2019-06-20
Avoid queue creation when waiting on previously unused HSAQueue

scchan:
This fixes a problem where calling hipStreamDestroy on
an hipStream_t that has never been used before would cause
a new hsa queue being created unnecessary


scchan:
@gargrahul this is a follow up to the redundant queues fix earlier.  This addresses the issue when destroying a hipStream_t that has never been used would trigger queue creation (due to extra synchronization before the stream gets destroyed).  Patch already tested with Caffe2 and Pytorch

jeffdaily (approved):

PR #1184, opened 2019-06-21
do not abort if no gpu exists

scchan:
The runtime should not cause the program to exit if it couldn't find any GPU.


aaronenyeshi (approved):
LGTM !

srinivirt:
Wouldn't this leave 'def' unnitialized, causing KalmarContext get_default_dev() exit?
Any issues with dtor trying to delete stuff?

scchan:
It should exit when no device is available and a program still tries to use one.

PR #1185, opened 2019-06-26
Make sure the context gets cleanup

scchan:
This fixes the breakage of HCC_PROFILE not generating the profiling information of kernels.  This is a side-effect of the context not being properly cleaned-up when the runtime gets unloaded on the lazy init path.

### DIFF
--- a/lib/mcwamp.cpp
+++ b/lib/mcwamp.cpp
@@ -167,48 +167,55 @@ static RuntimeImpl* LoadCPURuntime() {
   return runtimeImpl;
 }
 
-RuntimeImpl* GetOrInitRuntime() {
-  static RuntimeImpl* runtimeImpl = nullptr;
-  if (runtimeImpl == nullptr) {
-    HSAPlatformDetect hsa_rt;
+static RuntimeImpl* GetOrInitRuntime_impl() {
+  RuntimeImpl* runtimeImpl = nullptr;
+  HSAPlatformDetect hsa_rt;
 
-    char* verbose_env = getenv("HCC_VERBOSE");
-    if (verbose_env != nullptr) {
-      if (std::string("ON") == verbose_env) {
-        mcwamp_verbose = true;
-      }
+  char* verbose_env = getenv("HCC_VERBOSE");
+  if (verbose_env != nullptr) {
+    if (std::string("ON") == verbose_env) {
+      mcwamp_verbose = true;
     }
+  }
 
-    // force use certain C++AMP runtime from HCC_RUNTIME environment variable
-    char* runtime_env = getenv("HCC_RUNTIME");
-    if (runtime_env != nullptr) {
-      if (std::string("HSA") == runtime_env) {
-        if (hsa_rt.detect()) {
-          runtimeImpl = LoadHSARuntime();
-        } else {
-          std::cerr << "Ignore unsupported HCC_RUNTIME environment variable: " << runtime_env << std::endl;
-        }
-      } else if(std::string("CPU") == runtime_env) {
-          // CPU runtime should be available
-          runtimeImpl = LoadCPURuntime();
-          runtimeImpl->set_cpu();
-      } else {
-        std::cerr << "Ignore unknown HCC_RUNTIME environment variable:" << runtime_env << std::endl;
-      }
-    }
-
-    // If can't determined by environment variable, try detect what can be used
-    if (runtimeImpl == nullptr) {
+  // force use certain C++AMP runtime from HCC_RUNTIME environment variable
+  char* runtime_env = getenv("HCC_RUNTIME");
+  if (runtime_env != nullptr) {
+    if (std::string("HSA") == runtime_env) {
       if (hsa_rt.detect()) {
         runtimeImpl = LoadHSARuntime();
       } else {
-          runtimeImpl = LoadCPURuntime();
-          runtimeImpl->set_cpu();
-          std::cerr << "No suitable runtime detected. Fall back to CPU!" << std::endl;
+        std::cerr << "Ignore unsupported HCC_RUNTIME environment variable: " << runtime_env << std::endl;
       }
+    } else if(std::string("CPU") == runtime_env) {
+        // CPU runtime should be available
+        runtimeImpl = LoadCPURuntime();
+        runtimeImpl->set_cpu();
+    } else {
+      std::cerr << "Ignore unknown HCC_RUNTIME environment variable:" << runtime_env << std::endl;
     }
   }
+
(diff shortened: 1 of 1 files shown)


jeffdaily:
To be safe, we should test this change with the PyTorch tests that use fork().

jeffdaily:
PyTorch seems fine.  LGTM.

jeffdaily (approved):

jeffdaily:
Alright, I swear it wasn't seg faulting when I tested it a few days ago.  But now it is.  It seems shutdown order again is a problem.  
```
Thread 1 "python3" received signal SIGSEGV, Segmentation fault.
getContext () at /root/hcc/lib/mcwamp.cpp:420
420       return static_cast<KalmarContext*>(CLAMP::GetOrInitRuntime()->m_GetContextImpl());
(gdb) bt
#0  getContext () at /root/hcc/lib/mcwamp.cpp:420
#1  0x00007fffe4beb923 in hc::accelerator::accelerator() () from /opt/rocm/lib/libhip_hcc.so
#2  0x00007fffe4c3f494 in hipFree () from /opt/rocm/lib/libhip_hcc.so
#3  0x00007fff4e7ec115 in _rocblas_handle::~_rocblas_handle() () from /opt/rocm/lib/librocblas.so.0
#4  0x00007fff4e7eceb7 in rocblas_destroy_handle () from /opt/rocm/lib/librocblas.so.0
#5  0x00007fffd357c7fe in miopen::Handle::~Handle() () from /opt/rocm/lib/libMIOpen.so.1
#6  0x00007fffd31d358e in miopenDestroy () from /opt/rocm/lib/libMIOpen.so.1
#7  0x00007fff6e1f8621 in std::unordered_map<int, at::native::(anonymous namespace)::Handle, std::hash<int>, std::equal_to<int>, std::allocator<std::pair<int const, at::native::(anonymous namespace)::Handle> > >::~unordered_map() () from /root/.local/lib/python3.6/site-packages/torch/lib/libcaffe2_hip.so
#8  0x00007ffff6cd6ff8 in __run_exit_handlers (status=0, listp=0x7ffff70615f8 <__exit_funcs>, run_list_atexit=run_list_atexit@entry=true) at exit.c:82
#9  0x00007ffff6cd7045 in __GI_exit (status=<optimized out>) at exit.c:104
#10 0x00007ffff6cbd837 in __libc_start_main (main=0x4d18a0 <main>, argc=10, argv=0x7fffffffe5f8, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffe5e8)
    at ../csu/libc-start.c:325
#11 0x00000000005e65e9 in _start ()
```

scchan:
This test also failed on Jenkins probably for the same reason, I haven't got a chance to look into that yet.  That's also why this PR hasn't been merged yet.

Failing Tests (1):
    HCC :: Unit/HC/early_finalize_1.cpp

jeffdaily:
I will take a look today as highest priority.  This is blocking some other analysis work.

scchan:
there shouldn't be a reason for libmcwamp.so to shutdown before libhip_hcc.so according to the dependence chain.

jeffdaily:
Are the exit handlers running after dlcose() of libmcwamp_hsa.so?  I've seen this particular stack trace before, when libmcwamp_hsa.so still had the one main static context.  The issue seemed to resolve when I added the ShutdownImpl(), but perhaps it merely caused libmcwamp.so to not properly shut down libmcwamp_hsa.so, so this exit handler ordering was avoided (incorrectly).

jeffdaily:
It seems if we move the function-static `runtime` variable out of the `GetOrInitRuntime()` function, it resolves the issue.  

I'm not an expert on the C++ standard, but it seems static objects are destructed in reverse order of first use.  So when we lazily create the `runtime` variable inside `GetOrInitRuntime()`, it is constructed very late in the execution of the program i.e. perhaps after MIOpen has registered its exit handler.  So our HCC runtime destructs first.  

So by moving the `static std::unique_ptr<RuntimeImpl> runtime` outside of the `GetOrInitRuntime()` function, into global scope, an empty unique_ptr is created during dlopen of libmcwamp.so.  It is populated the first time `GetOrInitRuntime()` is called.  Then it is statically destructed when libmcwamp.so is closed.  This seems to happen after exit handlers run.

jeffdaily:
@scchan  apparently I have write access to your fork's branch for this PR.  I pushed my two-line change.  Let's see if CI finishes now.  I am also testing and verifying locally.  PyTorch now runs and exits cleanly with this change.  I'm running the HCC test suite now locally, as well

jeffdaily:
When CI runs the test, early_finalize_1.cpp still fails.  When I run "make test" locally, it also fails.  But when I run the binary that the test linked, independent of the test environment, it passes.

scchan:
I observe the same behaviour in my environment.

jeffdaily:
Is there any way to run just the failing test but using the test runner?

jeffdaily:
I answered my own question about running just the failing test.  From the build directory, run:

./compiler/bin/llvm-lit -j 4 --filter early ./tests

jeffdaily:
@scchan it's seg faulting while destroying the printf buffer in libmcwamp_hsa.so when deleting the static context.  Can we just disable the printf buffer .... for ever and on all systems ... ?  Seems more trouble than it's worth at this point.

jeffdaily:
@scchan have you had any time to consider what to do about the failing test, given my triage of the printf buffer?

scchan:
@jeffdaily I was looking into that and confirmed the core dump was due to the removal of the printf buffer pointer from the memory tracker.  Due to lack of maintenance of this experimental feature, I'm going to disable it

scchan:
#1199 will address the printf issue

jeffdaily:
Since #1199 is merged, should I cherry-pick it or rebase this PR to verify it fixes the remaining test?

Edit: Well, I guess since it was your PR to start with, you should take the lead ...

scchan:
@jeffdaily Jenkins already took care of that :)

whchung:
@jeffdaily @scchan it seems commit introduced in this PR may brought unknown butterfly effects on TensorFlow... stay tuned.

jeffdaily:
@scchan some tensorflow tests were newly failing during shutdown with
```
Inconsistency detected by ld.so: dl-close.c: 811: _dl_close: Assertion `map->l_init_called' failed!
```
Removing RTLD_NODELETE from mcwamp.cpp seems to resolve the issue for TensorFlow, but more testing would be needed of course with other parts of the ROCm stack and other frameworks.

PR #1187, opened 2019-06-26
manual upstream merge 0626109

david-salinas:
Change-Id: Iff1012289f83d89c5ea1044d6a7ff58d6809d212


scchan:
I don't think we need this anymore?

PR #1205, opened 2019-07-15
Fix queue stealing issue with queue priorities 

scchan:
This fixes an issue in queue stealing that got exposed indirectly by lazy initialization.  For a given priority level, when there's no queue created yet for this particular level and that the maximum number of queues have been reached, queue stealing will get stuck.  The fix is to look for an available queue from a different priority level, adjust the priority level and reassign.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -1194,23 +1194,8 @@ struct RocrQueue {
     }
 
     RocrQueue(hsa_agent_t agent, size_t queue_size, HSAQueue *hccQueue, queue_priority priority)
-        : _priority(priority)
+        : _priority(priority_normal)
     {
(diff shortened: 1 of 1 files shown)


jeffdaily:
We might need a deeper code review here.  I understand the problem this PR is trying to fix, but I found another latent problem.  The `HSAQueue::acquireLockedHsaQueue()`, if the `rocrQueue` is null, will always `createOrstealRocrQueue(this)` without specify its priority and default to normal priority.  See https://github.com/RadeonOpenCompute/hcc/blob/8c88ffb398e581e95ce93cd5a6d3ef470b64774a/lib/hsa/mcwamp_hsa.cpp#L4050-L4056

scchan:
@jeffdaily Good catch, I added a commit to fix that.

jeffdaily (approved):

PR #1211, opened 2019-07-16
Dont make the imported hsa-runtime64 target global

pfultz2:
This causes problems if `find_package(hcc)` gets called twice.


david-salinas (approved):
LGTM

scchan (approved):

PR #1220, opened 2019-07-29
flush queue on wait()

jeffdaily:


jeffdaily:
Test suite passes when running locally, but I could see how this PR might cause the error

```
tests/Unit/AsyncPFE/async_av_dependent8.cpp:85: void test1D() [grid_size = 32, tile_size = 16]: Assertion `hc::accelerator().get_default_view().get_pending_async_ops() == expectedPendingOps' failed.
```

Depending on when the queue flush occurs, the number of pending async ops would be reset back to 1 after the flush.  The flush during wait(), as well as the get_pending_async_ops(), are both correctly using the qmutex.  I would be favor of updating the test logic.  I don't think it is safe to assume how many pending ops exist.

scchan:
@jeffdaily The value of this test is questionable, I'd suggest removing it.

jeffdaily:
@scchan there a few tests that assert or test on get_pending_async_ops() -- how should we proceed? 

- tests/Unit/HC/test2.cpp
- tests/Unit/AsyncPFE/async_av_dependent8.cpp
- tests/Unit/AsyncPFE/async_av_independent4.cpp
- tests/Unit/AsyncPFE/async_av_dependent7.cpp

jeffdaily:
@scchan The failing test never exercises the new code from this PR.  Unrelated.  I verified by adding `assert(false)` (and print statements) to the new code and building and running only the failing test from the last CI run.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1637,6 +1638,16 @@ class HSAQueue final : public KalmarQueue
             DBOUT(DB_CMD2, "No future found in wait, enqueued marker into " << *this << "\n");
         }
         back()->getFuture()->wait();
+        if (HCC_FLUSH_ON_WAIT) {
+            // aggressively cleanup resources
+            // but keep back() as a valid HSAOp, so decrement current insert index twice
+            int back_op_index = decrement(asyncOpsIndex); // index of back() op
+            int index = decrement(back_op_index); // index of first op to free
+            while (asyncOps[index] != nullptr && index != back_op_index) {
```
srinivamd, line 1646: Why is the asyncOps[index] != nullptr check needed in the while condition?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -119,6 +119,7 @@ int HCC_MAX_QUEUES = 20;
 #define HCC_PROFILE_SUMMARY (1<<0)
 #define HCC_PROFILE_TRACE   (1<<1)
 int HCC_PROFILE=0;
+int HCC_FLUSH_ON_WAIT=1;
```
srinivamd, line 122: How about defining the HCC_FLUSH_ON_WAIT as a count of asyncops to reclaim, so it can be tuned for performance analysis? A value of 0 or undefined env would be equivalent to no reclaim.

lib/hsa/mcwamp_hsa.cpp
jeffdaily: When the HSAQueue is created, its `std::vector< std::shared_ptr<HSAOp> > asyncOps` is default constructed with size `HCC_ASYNCOPS_SIZE`.  This should mean they are all initialized to `nullptr`.  The `asyncOps` vector is treated as a circular buffer via the `asyncOpsIndex` for the current insert index.  The flushing code starts at the `index` prior to the youngest op (`back_op_index`).   This `index` may refer to a nullptr instead of a valid HSAOp because we assign nullptr to HSAOps to trigger their ref-counted destructors.  The check will help short-circuit the while loop once we hit the first non-existent (`nullptr`) HSAOp in the circular buffer.  Otherwise, we would traverse the entire `HCC_ASYNCOPS_SIZE` items, currently 16k in size, for every `wait()`.

lib/hsa/mcwamp_hsa.cpp
jeffdaily: This is possible, sure.  Is it desirable to not recover all resources?

I imagine we would want to recover the resources associated with the oldest HSAOps first.  We do not maintain an index for the oldest op.

srinivamd: Isn't the worst case cleaning up HCC_ASYNCOPS_SIZE (16k) items? The env variable as a count to reclaim could be used as an upper limit to limit any intermittent/variable performance impact due to pending ops to, say, a more "predictable" behavior. Also, could be used to tune and collect data for analysis.

lib/hsa/mcwamp_hsa.cpp
srinivamd: If we want to avoid traversing large portions of the circular buffer, perhaps using the new environment variable as an upper bound count of items to reclaim could ensure that?

scchan: Partially cleaning up the async op vector would break the dependence chain.  It "may" still work but undesirable.  The threshold doesn't efficiently solve the problem that the short-circuit in the while loop could do here.

lib/hsa/mcwamp_hsa.cpp
scchan: I'd suggest the HCC_ASYNCOPS_SIZE being the first parameter to tune.  I think the current default is too large and there may be a sweet spot where it's sufficient large to accomodate a relatively large batch of dispatches without forcing a wait.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1637,6 +1638,16 @@ class HSAQueue final : public KalmarQueue
             DBOUT(DB_CMD2, "No future found in wait, enqueued marker into " << *this << "\n");
         }
         back()->getFuture()->wait();
+        if (HCC_FLUSH_ON_WAIT) {
+            // aggressively cleanup resources
+            // but keep back() as a valid HSAOp, so decrement current insert index twice
```
scchan, line 1643: it's unclear to me the reason for keep the recent HSAOp, could you elaborate?

scchan (changes requested):

lib/hsa/mcwamp_hsa.cpp
jeffdaily: During testing, I was getting seg faults if I didn't leave the last op.  I think I figured out why as I review the code.  `isEmpty()` is using `back()` without checking it for nullptr.

lib/hsa/mcwamp_hsa.cpp
jeffdaily: if hsa signals are the precious resource, perhaps we should make that the limit.  also, #1173 could help reduce the resource use, but it is a much bigger than than this current PR.

lib/hsa/mcwamp_hsa.cpp
srinivamd: It's kind of counter-intuitive to nullify a non-null pointer (even though it is just decrementing ref count). If non-null asyncOps pointer can be reclaimed, does that mean the getPendingAsyncOps() logic needs to be updated?

scchan: getPendingAsyncOps is still correct but just not optimized, it's a function for debugging anyway

lib/hsa/mcwamp_hsa.cpp
scchan: I think I'll merge in this PR as-is and address the isEmpty() issue in another PR.

scchan (approved):

PR #1227, opened 2019-08-04
use realpath to deduce absolute path of the script

zaliu:
Use realpath to deduce absolute path of the script so that clang_offload_bundler can be located.  This is important when extractkernel is picked up from /opt/rocm/bin as opposed to /opt/rocm/hcc/bin because clang_offload_bundler exists only under /opt/rocm/hcc/bin.

### DIFF
--- a/lib/extractkernel.in
+++ b/lib/extractkernel.in
@@ -2,7 +2,8 @@
 use strict;
 use File::Copy;
 use File::Spec;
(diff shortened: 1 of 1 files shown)


aaronenyeshi (approved):

david-salinas (approved):
LGTM

PR #1230, opened 2019-08-08
When inspecting a victim queue, use try_lock to avoid deadlock

scchan:


jeffdaily (approved):
LGTM.  I can't believe I missed the loop bound check in an earlier review.

david-salinas (approved):
LGTM

aaronenyeshi (approved):

PR #1243, opened 2019-08-16
remove use of RTLD_NODELETE from mcwamp.cpp

jeffdaily:
This fixes a problem in certain TensorFlow tests where the following assertion occurs during shutdown.
```
Inconsistency detected by ld.so: dl-close.c: 811: _dl_close: Assertion `map->l_init_called' failed!
```


scchan (approved):
LGTM

PR #1246, opened 2019-08-20
Introduce KMHACKISA environment variable for ISA-level code injection.

whchung:
When used with `-fno-gpu-rdc`, this allows applications to inject ISA into
object files. This flag is used for kernel developers to tweak ISA before an
optimization can be devised in the compiler.

The feature is developed for @asroy and @zjing14 . Need verification from their applications first.

### DIFF
--- a/lib/clamp-device.in
+++ b/lib/clamp-device.in
@@ -34,6 +34,9 @@ KMOPTLLC="${KMOPTLLC:="-O2"}"
 # enable LLVM hijacking
 KMHACKLLVM="${KMHACKLLVM:=0}"
 
+# enable ISA hijacking
+KMHACKISA="${KMHACKISA:=0}"
+
 # enable ThinLTO
 KMTHINLTO="${KMTHINLTO:=0}"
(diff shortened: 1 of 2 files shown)


scchan (commented):
I think this should work

whchung:
@scchan , @asroy and @zjing14 both could apply the patch onto their projects and do ISA-level tweaking now. Removing `[DO NOT MERGE]` tag now.

whchung:
+ @jehandadkhan for awareness.

with this PR you are able to inject additional or replace ISA coming out from the compiler.

whchung:
@scchan This feature is gradually getting picked up by kernel developers in MIOpen team. Wondering could you give this PR another go? Thanks.

scchan:
@whchung I'd like to get PR1245 first if possible since it's higher priority.  This PR will probably have be updated to adapt to those changes.

aaronenyeshi (changes requested):
Closing old prs

PR #1251, opened 2019-08-26
Changes to CMakeLists file to make sure that it can support SLES

adilad:
Addition of packages for SLES in CPACKRPM


adilad:
Will submit a new one.

PR #1256, opened 2019-08-30
use a smaller async ops vector size to reduce the number of signals

scchan:


scchan:
@gargrahul could you please review?

gargrahul (approved):
LGTM

jeffdaily:
How small is too small?  If it is too small, the host will block until the op signal clears.  Have we verified whether this affects performance of benchmarks and frameworks?

jeffdaily:
I can see a small asyncop size like this being okay, but only if we also managed to not allocate a signal for every kernel.  Otherwise, frequent queuing of kernels will block the host.  I am aware of one or more PyTorch workloads that queues many kernels, where we were already seeing the queue depth to be reached when it was still 16K.

scchan:
This PR was closed by mistake and it couldn't be reopen for some reason...

PR #1265, opened 2019-09-06
Use the path of libmcwamp to dlopen libmcwamp_hsa

scchan:


lib/mcwamp.cpp
```diff
@@ -428,6 +433,22 @@ KalmarContext *getContext() {
...
+        if (pos != std::string::npos) {
+          p.erase(p.begin()+pos, p.end());
+          CLAMP::library_path = std::move(p);
+          return 1;
+        }
+      }
+      return 0;
+    }, nullptr);
+
     bool to_init = false;
```
srinivamd, line 452: LGTM

srinivamd (approved):
LGTM

jeffdaily:
I'm getting a seg fault with pytorch test.sh.

jeffdaily:
Turning CLAMP::library_path into a static function seems to avoid the seg fault.  See
https://github.com/jeffdaily/hcc/pull/1
for the diff.

scchan:
> Turning CLAMP::library_path into a static function seems to avoid the seg fault. See
> [jeffdaily#1](https://github.com/jeffdaily/hcc/pull/1)
> for the diff.

I haven't seen any segfault... but perhaps it may have to do with the order of when the global objects get constructed.
I made a comment to your commit, could you try that?  If that works, could you submit a PR?

scchan:
replaced by https://github.com/RadeonOpenCompute/hcc/pull/1266

PR #1266, opened 2019-09-06
Use the path of libmcwamp to dlopen libmcwamp_hsa, as static function

jeffdaily:
Starting from #1265 , make the static variable CLAMP::library_path into a static function get_library_path() returning a std::string.  This avoids a seg fault caused by the previous PR attempt.

### DIFF
--- a/lib/mcwamp.cpp
+++ b/lib/mcwamp.cpp
@@ -91,7 +91,30 @@ namespace CLAMP {
 // Class declaration
 ////////////////////////////////////////////////////////////
 
-static std::string library_path;
+static std::string get_library_path()
+{
+    static std::string library_path;
+    static std::once_flag once;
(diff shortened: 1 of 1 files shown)


lib/mcwamp.cpp
```diff
@@ -90,6 +90,32 @@ namespace CLAMP {
 ////////////////////////////////////////////////////////////
 // Class declaration
 ////////////////////////////////////////////////////////////
+
+static std::string get_library_path()
```
scchan: can we return `std::string&` instead?

scchan:
This couldn't be built on Ubuntu 16.04 probably due to a bug in the implementation of std::call_once, replacing this PR with https://github.com/RadeonOpenCompute/hcc/pull/1266

PR #1282, opened 2019-10-09
DO NOT MERGE:  Move HCC to use llvm monorepo

david-salinas:
This is a dummy PR to demonstrate changes we will need (shortly) when we move to the LLVM monorepo.


aaronenyeshi:
Looks good, also conflicts look like it needs a rebase or ctu merge

PR #1283, opened 2019-10-09
migrate to monorepo

david-salinas:
This PR, is just a sneak-peek of the PR we will create tomorrow.  This has the latest submodules deleted.  And the gitref to llvm-project would need to be updated in the "real" PR.

Change-Id: If8fbc2f4e563582d6178dcf338bd93ddfc50ee02

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,24 +1,8 @@
-[submodule "llvm"]
-	path = compiler
-	url = https://github.com/RadeonOpenCompute/llvm.git
-	branch = amd-common
-[submodule "lld"]
-	path = lld
-	url = https://github.com/RadeonOpenCompute/lld.git
-	branch = amd-common
-[submodule "clang"]
-	path = clang
-	url = https://github.com/RadeonOpenCompute/hcc-clang-upgrade.git
-	branch = clang_tot_upgrade
-[submodule "compiler-rt"]
-	path = compiler-rt
-	url = https://github.com/RadeonOpenCompute/compiler-rt
-        branch = amd-common
 [submodule "rocdl"]
 	path = rocdl
 	url = http://github.com/RadeonOpenCompute/ROCm-Device-Libs.git
         branch = master
-[submodule "clang-tools-extra"]
-	path = clang-tools-extra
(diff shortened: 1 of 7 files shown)


.gitmodules
```diff
@@ -1,24 +1,8 @@
...
 	url = http://github.com/RadeonOpenCompute/ROCm-Device-Libs.git
         branch = master
-[submodule "clang-tools-extra"]
-	path = clang-tools-extra
-	url = https://github.com/RadeonOpenCompute/clang-tools-extra.git
-	branch = amd-common
+[submodule "llvm-project"]
+	path = llvm-project
+	url = https://github.com/david-salinas/llvm-project
+	branch = amd-common-monorepo-hcc
```
aaronenyeshi: is this supposed to be amd-stg-open-hcc on the RadeonOpenCompute/llvm-project?

aaronenyeshi (changes requested):

.gitmodules
aaronenyeshi: Nvm, you mentioned that in your commit. So the project isn't live yet

david-salinas: yes.  I was just using this repo/branch to validate that HCC could be built with a monorepo.  The "real" PR will have this correct, which will be github.com/llvm-project and branch amd-stg-open-hcc, when that branch is available.

david-salinas:
I've updated this PR, so that we can use it to move HCC to the monorepo, when RadeonOpenCompute/llv-project is updated with the amd-stg-open-hcc branch.

aaronenyeshi (approved):

scchan (changes requested):
LGTM, but still need to wait for the llvm-project and the test results to become available before approving this.

david-salinas:
I've updated this PR, so that we can use it to move HCC to the monorepo, when RadeonOpenCompute/llv-project is updated with the amd-stg-open-hcc branch.

> LGTM, but still need to wait for the llvm-project and the test results to become available before approving this.

And .... before we merge this .... I need to update the gitref for llvm-project to point to branch amd-stg-open-hcc, once it exists and is ready.

PR #1285, opened 2019-10-14
SWDEV-204509: re-enable printf on all platforms (including CentOS)

ssahasra:
printf was because std::regex is broken on CentOS. The dependence on
std::regex is now removed.

### DIFF
--- a/include/hc_printf.hpp
+++ b/include/hc_printf.hpp
@@ -6,7 +6,6 @@
 #include <cassert>
 #include <atomic>
 #include <string>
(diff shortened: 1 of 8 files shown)


jeffdaily:
`printf` was disabled first for CentOS reasons.  But it was later found to cause other faults during shutdown and was disabled for all platforms.  #1199

ssahasra:
Thanks for the pointer, @jeffdaily ! Discarding this PR for now.

PR #1286, opened 2019-10-15
IGNORE THIS: re-enable printf on all platforms (including CentOS)

ssahasra:
Created a PR just to trigger a Jenkins job, because I don't know of an easier method to do it.

printf was because std::regex is broken on CentOS. The dependence on
std::regex is now removed.

Change-Id: I401e4ed663df0d88ddbd4806ed44448a00f4836e


scchan:
Please reopen if needed

PR #1290, opened 2019-10-18
Printf no default

ssahasra:
printf was previously disabled for two reasons:

1. The format string was parsed using std::regex, which is broken on
   CentOS.

   This dependence on std::regex is now replaced with simpler
   parsing of the format string using string::find_first_of.

2. One unit test in particular, early_finalize, was crashing on the
   Jenkins build that uses an Ubuntu 16.04 container. The crash was
   traced to a segfault when trying to flush/delete the printf buffer
   when the application exits.

   The reason for the segfault is not yet understood, but simply
   disabling the deletion removes the crash. Note that the printf
   buffer is no longer valid at the point of deletion, and the
   contents of the buffer are flushed after every kernel returns.

In addition, this change sets HCC_ENABLE_PRINTF to 0 in the LIT tests
by default, and sets it to 1 only for the printf tests under tests/Unit/Printf.

See the following github PRs for more details:
#1185
#1199
#1169


include/hc_printf.hpp
```diff
@@ -98,14 +88,14 @@ enum PrintfPacketDataType {
...
+  void set(uint32_t d)     [[hc,cpu]] { type = PRINTF_UINT32_T;       data.uli = d; }
+  void set(int32_t d)      [[hc,cpu]] { type = PRINTF_INT32_T;        data.li = d; }
   void set(uint64_t d)     [[hc,cpu]] { type = PRINTF_UINT64_T;       data.uli = d; }
   void set(int64_t d)      [[hc,cpu]] { type = PRINTF_INT64_T;        data.li = d; }
   void set(unsigned long long d) [[hc,cpu]] { type = PRINTF_UINT64_T; data.uli = d; }
   void set(long long d)    [[hc,cpu]] { type = PRINTF_INT64_T;        data.li = d; }
-  void set(hc::half d)     [[hc,cpu]] { type = PRINTF_HALF;           data.h = d; }
-  void set(float d)        [[hc,cpu]] { type = PRINTF_FLOAT;          data.f = d; }
+  void set(hc::half d)     [[hc,cpu]] { type = PRINTF_HALF;           data.d = d; }
+  void set(float d)        [[hc,cpu]] { type = PRINTF_FLOAT;          data.d = d; }
```
scchan, line 98: Casting these to double would add a conversion which changes their value and their bit pattern.  If for example the format specifier is to print in hex that this will print out the wrong value.

scchan (changes requested):

include/hc_printf.hpp
ssahasra: There are a number of things to consider here:

1. The current printf is a template, but the actual printf is supposed to be a variadic function. When calling a variadic function, "standard promotions" are applied automatically, where all floating point types smaller than double are converted to double. This is what we will see when we eventually have the new printf.
2. The current printf converts data.h to float anyway before printing via std::cout.
3. There is no format specifier for printing the hex of a half value in CUDA. If we introduce one, then it is the job of this specifier to convert the argument to half before printing. Usual considerations about rounding mode apply in that case.

scchan:
LGTM

PR #1296, opened 2019-10-31
do not dlclose; better lazy init of HSA context

jeffdaily:


ssahasra (approved):
Looks good to me. After this is committed, I will check if I can renable the flush/delete of the printf buffer.

jeffdaily:
@ssahasra yes thank you.  Though this PR is needed to resolve a seg fault in one of our codes, I was concerned about the moving of the printf block of code.
Is there a corresponding PR for re-enabling printf?  Has it been tested against the previous failure cases (I would need to search our ticketing system)?

ssahasra:
Printf was re-enabled a few days ago with the following PR:
https://github.com/RadeonOpenCompute/hcc/pull/1290

It's already merged in clang_tot_upgrade.

jeffdaily:
@ssahasra when I merged this PR with upstream I mistakenly added the printf ifdef back in and I'm assuming that is what broke the few CI tests.  I just pushed the fix; let's see if CI recovers.

ssahasra:
> 
> 
> @ssahasra when I merged this PR with upstream I mistakenly added the printf ifdef back in and I'm assuming that is what broke the few CI tests. I just pushed the fix; let's see if CI recovers.

Good catch! I confused that macro with the other one "HCC_ENABLE_ACCELERATOR_PRINTF" when reviewing your original patch.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -5846,24 +5868,11 @@ HSACopy::syncCopy() {
...
-      });
-    }
-  }
-  return Kalmar::ctx;
+  return Kalmar::ctx();
 }
 
 extern "C" void ShutdownImpl() {
-  delete Kalmar::ctx;
+  delete Kalmar::__ctx;
```
scchan: It seems that `Kalmar::__ctx` should be initialized to `nullptr` and then we should check whether it's still `nullptr` here before deleting it.

scchan (changes requested):

scchan (approved):

PR #1309, opened 2019-11-13
Remove the runtime plugin infrastructure

scchan:
The runtime plugin was there to support multiple runtime backends in the past.  Since we've only
supported the ROCR runtime backend since ROCm 1.x and there has been no support for other runtime backends, remove the runtime plugin clutter.


jeffdaily:
What is the rationale for keeping libhc_am.so as a separate library?

scchan:
> What is the rationale for keeping libhc_am.so as a separate library?
I think some of the apps and libs explicitly link to libhc_am.so.  If I remove libhc_am.so then I'll have to go and update all the libs affected.  Logically it makes sense to remove libhc_am.so as well but I just don't have the time to go through that at the moment.

ssahasra (approved):
LGTM. Can we now have another patch that removes the lib/cpu directory?

scchan:
> LGTM. Can we now have another patch that removes the lib/cpu directory?

It's already gone

PR #1314, opened 2019-11-20
Revise roctracer activity callbacks get invoked via ROCR signal callbacks

whchung:
Current implementation invokes roctracer only when an HSAOp is disposed. This
could be too late for many long-running applications. In this commit we use
ROCR signal callbacks instead, which could effectively notify roctracer at a
much better timing.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -725,6 +725,14 @@ public:
 
     virtual void setSelf(const std::shared_ptr<HSAOp> &self) { _self = self; }
 
+    // Factory method to build HSAOp.
+    template<typename T, typename ... Ts>
+    static std::shared_ptr<T> buildOp(Ts ... args) {
+        auto op = std::make_shared<T>(args...);
+        op->setSelf(op);
+        return op;
+    }
+
 protected:
     uint64_t     apiStartTick;
     HSAOpCoord   _opCoord;
@@ -2012,8 +2020,7 @@ public:
         hsa_status_t status = HSA_STATUS_SUCCESS;
 
         // create shared_ptr instance
-        std::shared_ptr<HSABarrier> barrier = std::make_shared<HSABarrier>(this, 0, nullptr);
-        barrier->setSelf(barrier);
+        std::shared_ptr<HSABarrier> barrier = HSAOp::buildOp<HSABarrier>(this, 0, nullptr);
         // associate the barrier with this queue
         pushAsyncOp(barrier);
 
@@ -2045,8 +2052,7 @@ public:
         if ((count >= 0) && (count <= HSA_BARRIER_DEP_SIGNAL_CNT)) {
 
             // create shared_ptr instance
-            std::shared_ptr<HSABarrier> barrier = std::make_shared<HSABarrier>(this, count, depOps);
-            barrier->setSelf(barrier);
+            std::shared_ptr<HSABarrier> barrier = HSAOp::buildOp<HSABarrier>(this, count, depOps);
             // associate the barrier with this queue
             pushAsyncOp(barrier);
 
@@ -4254,8 +4260,7 @@ std::shared_ptr<KalmarAsyncOp> HSAQueue::EnqueueAsyncCopyExt(const void* src, vo
 
     // create shared_ptr instance
     const Kalmar::HSADevice *copyDeviceHsa = static_cast<const Kalmar::HSADevice*> (copyDevice);
-    std::shared_ptr<HSACopy> copyCommand = std::make_shared<HSACopy>(this, src, dst, size_bytes);
-    copyCommand->setSelf(copyCommand);
+    std::shared_ptr<HSACopy> copyCommand = HSAOp::buildOp<HSACopy>(this, src, dst, size_bytes);
 
     // euqueue the async copy command
     status = copyCommand.get()->enqueueAsyncCopyCommand(copyDeviceHsa, srcPtrInfo, dstPtrInfo);
@@ -4276,8 +4281,7 @@ std::shared_ptr<KalmarAsyncOp> HSAQueue::EnqueueAsyncCopy2dExt(const void* src, 
 
     //create shared_ptr instance
     const Kalmar::HSADevice *copy2dDeviceHsa = static_cast<const Kalmar::HSADevice*> (copyDevice);
-    std::shared_ptr<HSACopy> copy2dCommand = std::make_shared<HSACopy>(this, src, dst, width*height);
-    copy2dCommand->setSelf(copy2dCommand);
+    std::shared_ptr<HSACopy> copy2dCommand = HSAOp::buildOp<HSACopy>(this, src, dst, width*height);
 
     //euqueue the async copy command
     status = copy2dCommand.get()->enqueueAsyncCopy2dCommand(width, height, srcPitch, dstPitch, copy2dDeviceHsa, srcPtrInfo, dstPtrInfo);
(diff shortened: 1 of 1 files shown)


scchan:
For this to work, it assumes that the async op object is still alive when the callback is being invoked?

jeffdaily:
Instead of passing a raw pointer to the HSAOp, can a shared_ptr somehow be used and still ref count correctly, etc.?

whchung:
@scchan `signalCallback` would be triggered when the value of the associated HSA signal of a HSAOp changes. At that moment HSAOp is still alive and well.

whchung:
@jeffdaily looking at `activity_prof.h` it seems in `ActivityProf` merely retrieve the following information from a HSAOp:

- getCommandKind()
- getBeginTimestamp()
- getEndTimestamp()
- getStartTick()
- getSystemTicks()

And it doesn't keep a reference to the HSAOp. So as long as HSAOp is alive when `signalCallback` is invoked (which is the case), passing the raw pointer to `_activity_prof` should be fine.

whchung:
@scchan / @jeffdaily the purpose of the PR is to use ROCR runtime callback mechanism so HSA signals associated with HSAOp in HCC queue gets monitored by ROCR runtime, and user applications can received callbacks via this route:

a) ROCR runtime monitors signal values -> b) ROCR runtime notifies `signalCallback` in HCC runtime upon signal value changes -> c) HCC runtime notifies roctracer -> d) roctracer notifies client applications via activity callbacks in client applications.

If I understand the current implementation of HCC runtime correctly, HSAOp within a queue would be disposed and released in the following circumstances:

1) stream synchronization
2) queue overflow
3) end of program

In any of the case we'll flush the underlying queue, which means all signals should change value prior to the destruction of HSAOp instances. So when `signalCallback` gets invoked, the `reinterpret_cast<HSAOp*>(arg)` should still be good.

Please help review if my logic here is sound and valid.

scchan:
@whchung That's only true if the callback happens before the HSAOp has a chance to be destroyed.  That might be true most of the time in practice assuming the time between the signal change and the callback is very small.  However, the guarantee is not there.

whchung:
@scchan I understand your concern. How do you suggest we improve it so:

a) ensure HSAOp always outlives its associated signal
b) or, detect if the HSAOp is no longer valid within the callback and ignore notifying roctracer

jeffdaily:
What if we created a tiny wrapper around the `std::shared_ptr<HSAOp>` and passed that to the callback handler?  For example:

```C++
struct SharedWrapper {
    std::shared_ptr<HSAOp> op;
    SharedWrapper(std::shared_ptr<HSAOp> op) : op(op) {}
};

static bool signalCallback(hsa_signal_value_t value, void *arg) {
    if (arg != nullptr) {
        SharedWrapper *wrapper = reinterpret_cast<SharedWrapper*>(arg);
        wrapper->op->activityReport();
        delete wrapper;
    }
    return false; // do not re-use callback.

// called as such, perhaps in pushAsyncOp since it is an actual HSAOp by that time:
auto status = hsa_amd_signal_async_handler(signal, HSA_SIGNAL_CONDITION_LT, 1, signalCallback, new SharedWrapper(op));
}
```

I am not aware of how to otherwise pass a `shared_ptr` to a C callback `void*` and still maintain the ref counting.  This small wrapper will be deleted inside the signal call back, but after the activity is recorded using the `HSAOp` while it is still ref counted and alive.

Is it safe to free the HSA signal inside the HSA's signal handler?

whchung:
@jeffdaily I don't think it's a good idea to free the HSA signal in the signal handler. Right now we have a clear-defined semantic to release the signal, `HSAOp::dispose()`. We probably should want to keep such semantic.

Also the ctor of `SharedWrapper` doesn't look right to me, we are passing a vanilla `HSAOp` to `SharedWrapper` but `SharedWrapper` ctor expects a `shared_ptr<HSAOp>` and `shared_ptr` ctor is explicit according to:
https://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr

Plus we are essentially creating 2 `shared_ptr` trying to track the `HSAOp`, 1 in `HSAQueue::asyncOps`, 1 within `SharedWrapper`. This is going to have very bad consequences, refer to:

https://stackoverflow.com/questions/11223895/best-way-to-protect-from-multiple-shared-ptr-to-the-same-object

I'm still thinking about an alternative solution. Will post another memo.

whchung:
I think we want to protect the following scenario from happening:

a) HSA signal changes value
b) `signalCallback` gets invoked
c) corresponding `HSAOp` is no longer available
d) `signalCallback` dereferences invalid pointer and crash the program

@jeffdaily Wondering what if we amend your proposed changes so:

a) inside each `HSAOp`, it has an `weak_ptr` to itself. the `weak_ptr` is to track associated `shared_ptr<HSAOp` within `HSAQueue`.
b) `SharedWrapper` is constructed from the `weak_ptr`

This way the reference counting shall still be correct. I'll do some local experiments to verify.

whchung:
@scchan / @jeffdaily After my local experiments I'd like to propose amend @jeffdaily 's suggestion to:

```C++
struct SharedWrapper {
    std::shared_ptr<HSAOp> _op;
    SharedWrapper(const std::weak_ptr<HSAOp> &op) : op(op) {}
    ~SharedWrapper() { _op = nullptr; }
};

static bool signalCallback(hsa_signal_value_t value, void *arg) {
    if (arg != nullptr) {
        SharedWrapper *wrapper = reinterpret_cast<SharedWrapper*>(arg);
        wrapper->op->activityReport();
        delete wrapper;
    }
    return false; // do not re-use callback.
}

class HSAOp {
public:
  void setSelf(const std::shared_ptr<HSAOp> &self) {
    _self = self;
  }
private:
  std::weak_ptr<HSAOp> _self;
}

// called as such, perhaps in pushAsyncOp since it is an actual HSAOp by that time:
auto status = hsa_amd_signal_async_handler(signal, HSA_SIGNAL_CONDITION_LT, 1, signalCallback, new SharedWrapper(_self));
```

a) inside each `HSAOp`, it has an `weak_ptr` to itself. the `weak_ptr` is to track associated `shared_ptr<HSAOp` within `HSAQueue`.
b) a new `setSelf` method shall be invoked after an `HSAOp` is introduced. since it's a weak_ptr so it won't affect `HSAOp` life-cycle and memory management shall roctracer is not used.
c) `SharedWrapper` is constructed from the `weak_ptr` should we register a callback for a `HSAOp`.

I'll post a new commit to depict my idea.

whchung:
A self contained demo can be found at:
https://pastebin.com/st1FHLWb

build and run it with :
```
g++ -std=c++11 test.cc -pthread -o test
./test
```

You can see the object of `Foo` type is removed in the main thread, but since it's tracked by `FooWrapper` so its dtor is not invoked until the callback function is completed.

whchung:
@scchan / @jeffdaily I've pushed a new commit to incorporate aforementioned changes. Please kindly review it once again. I believe this should ensure:

a)  `HSAOp` never gets invalid before `signalCallback` is invoked.
b) `HSAOp` be properly destructed at the right timing ( when reference count goes to 0 ).

scchan:
I think this should work.

scchan (approved):

jeffdaily:
Can we wait to merge this?  I would like to manually merge this PR into #1317, in addition to some other changes that are related to this.

whchung:
@jeffdaily this PR can wait. as long as it hits ROCm 3.1 we are good.

whchung:
@scchan thanks for taking this PR. Would this feature land in ROCm 3.0? Or should we wait for 3.1?

+ @sunway513 for awareness.

PR #1315, opened 2019-11-20
remove __attribute__((used)) from math instrinsic wrappers

david-salinas:
an upstream clang commit: 8f089f2099d39021bbfb76a2cd575612382a7cf6 exposes this issue, where because we have __attribute__((used)) on our intrinsic wrappers, we end with superfluous calls to (actually) unused intrinsics.

https://github.com/llvm/llvm-project/commit/8f089f2099d39021bbfb76a2cd575612382a7cf6


scchan:
LGTM

PR #1316, opened 2019-11-20
disable ParallelSTL tests

david-salinas:


david-salinas:
We need to disable these tests for the time being.

scchan (approved):

PR #1317, opened 2019-11-21
remove use of shared_future, recursive_mutex in HSA runtime

jeffdaily:
Lower the overhead of HSA runtime by removing the use of shared_future.  Its functionality was redundant with the use of shared_ptr<HSAOp> and the HSA signal wait.

### DIFF
--- a/include/hc.hpp
+++ b/include/hc.hpp
@@ -378,9 +378,6 @@ public:
      */
     bool copy2d_ext(const void *src, void *dst, size_t width, size_t height, size_t srcPitch, size_t dstPitch, hcCommandKind copyDir, const hc::AmPointerInfo &srcInfo, const hc::AmPointerInfo &dstInfo, const hc::accelerator *copyAcc, bool forceUnpinnedCopy);
 
-    // TODO - this form is deprecated, provided for use with older HIP runtimes.
-    void copy_ext(const void *src, void *dst, size_t size_bytes, hcCommandKind copyDir, const hc::AmPointerInfo &srcInfo, const hc::AmPointerInfo &dstInfo, bool forceUnpinnedCopy) ;
-
     /**
      * Copies size_bytes bytes from src to dst.  
      * Src and dst must not overlap.  
@@ -1608,12 +1605,8 @@ inline void accelerator_view::copy_ext(const void *src, void *dst, size_t size_b
     pQueue->copy_ext(src, dst, size_bytes, copyDir, srcInfo, dstInfo, copyAcc ? copyAcc->pDev : nullptr, forceUnpinnedCopy);
 };
 
-inline void accelerator_view::copy_ext(const void *src, void *dst, size_t size_bytes, hcCommandKind copyDir, const hc::AmPointerInfo &srcInfo, const hc::AmPointerInfo &dstInfo, bool forceHostCopyEngine) {
-    pQueue->copy_ext(src, dst, size_bytes, copyDir, srcInfo, dstInfo, forceHostCopyEngine);
-};
-
 inline bool accelerator_view::copy2d_ext(const void *src, void *dst, size_t width, size_t height, size_t srcPitch, size_t dstPitch, hcCommandKind copyDir, const hc::AmPointerInfo &srcInfo, const hc::AmPointerInfo &dstInfo, const hc::accelerator *copyAcc, bool forceUnpinnedCopy) {
-    return (pQueue->copy2d_ext(src, dst, width, height, srcPitch, dstPitch, copyDir, srcInfo, dstInfo, copyAcc ? copyAcc->pDev : nullptr, forceUnpinnedCopy));
+    return pQueue->copy2d_ext(src, dst, width, height, srcPitch, dstPitch, copyDir, srcInfo, dstInfo, copyAcc ? copyAcc->pDev : nullptr, forceUnpinnedCopy);
 };
 
 inline completion_future
--- a/include/hc.hpp
+++ b/include/hc.hpp
@@ -1305,12 +1305,14 @@ public:
      *                     the expense of using one CPU core for active waiting.
      */
     void wait(hcWaitMode mode = hcWaitModeBlocked) const {
-        if (this->valid()) {
-            if (__asyncOp != nullptr) {
-                __asyncOp->setWaitMode(mode);
-            }   
-            //TODO-ASYNC - need to reclaim older AsyncOps here.
+        if (__amp_future.valid()) {
+            if (__asyncOp != nullptr) throw runtime_exception("completion_future expected amp, had async op", 0);
             __amp_future.wait();
+        }
+        if (__asyncOp != nullptr) {
+            if (__amp_future.valid()) throw runtime_exception("completion_future expected async op, had amp", 0);
+            __asyncOp->setWaitMode(mode);
+            __asyncOp->wait();
         }
 
         Kalmar::getContext()->flushPrintfBuffer();
@@ -1327,15 +1329,6 @@ public:
     }
 
     /** @} */
-
-    /**
-     * Conversion operator to std::shared_future<void>. This method returns a
-     * shared_future<void> object corresponding to this completion_future
-     * object and refers to the same asynchronous operation.
-     */
-    operator std::shared_future<void>() const {
-        return __amp_future;
-    }
 
     /**
      * This method enables specification of a completion callback func which is
@@ -1355,8 +1348,7 @@ public:
         // spawn a new thread to wait on the future and then execute the callback functor
         __thread_then = new std::thread([&]() __CPU__ {
           this->wait();
-          if(this->valid())
-            func();
+          func();
         });
       }
 #endif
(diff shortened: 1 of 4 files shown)


jeffdaily:
jenkins retest this please

scchan:
> jenkins retest this please

@david-salinas could you please restart the test once the CI issue has been resolved?

david-salinas:
> > jenkins retest this please
> 
> @david-salinas could you please restart the test once the CI issue has been resolved?

rerunning it now.

include/hc.hpp
```diff
@@ -1305,13 +1305,15 @@ class completion_future {
...
-            if (__asyncOp != nullptr) {
-                __asyncOp->setWaitMode(mode);
-            }   
-            //TODO-ASYNC - need to reclaim older AsyncOps here.
+        if (__amp_future.valid()) {
+            if (__asyncOp != nullptr) throw runtime_exception("completion_future expected amp, had async op", 0);
             __amp_future.wait();
         }
+        if (__asyncOp != nullptr) {
+            if (__amp_future.valid()) throw runtime_exception("completion_future expected async op, had amp", 0);
```
srinivamd, line 1310: Aren't new lines 1308-1309 and 1312-1313 identical in functionality unless the order of the conditions checked matter? If so, some comment would be helpful.
If not, would it be ok to factor it out as: 
`if (__amp_future.valid() &&  (__asyncOp != nullptr)) {`
`    throw runtime_exception("completion_future expected amp, had async op", 0);`
`}`

It could reduce indent of main code.

jeffdaily:
This PR is now updated to further improve runtime overhead.  The use of `recursive_mutex` has been removed.  In many internal APIs where a mutex was assumed to lock, a `NoLock` or `_no_lock` variant has been implemented for use in the methods that already had acquired the lock.

Further, a deprecated API `HSAQueue::copy_ext` was removed.  Compiling with `-Wall` detected infinite recursion in the old API.  Had any client code been calling this function, we would have heard about it by now due to this error.  It was safer to remove it than to continue to leave it deprecated.

These changes result in ~2-7% performance improvement in certain PyTorch benchmarks.

AlexVlx (approved):
LGTM, thanks.

scchan:
LGTM, waiting for the build to finish

scchan (approved):

jeffdaily:
@david-salinas could you please restart the test once the CI issue has been resolved?

jeffdaily:
Only one test failure.  Is this a known flaky test or not?
```
Failing Tests (1):
    HCC :: Unit/AmpMath/amp_math_expf_precise_math.cpp

  Expected Passes    : 671
  Expected Failures  : 44
  Unexpected Failures: 1
```

jeffdaily:
I ran that particular test on my local system in a loop 1000 times and it failed 13 times.  Perhaps the failure has more to do with the random distribution than any code changes?
```
for i in `seq 1000`; do ./tests/Unit/AmpMath/Output/amp_math_expf_precise_math.cpp.tmp.out; done
sum !< ERROR_THRESHOLD --- 0.104488 !< 0.100000
sum !< ERROR_THRESHOLD --- 0.111008 !< 0.100000
sum !< ERROR_THRESHOLD --- 0.109825 !< 0.100000
sum !< ERROR_THRESHOLD --- 0.107237 !< 0.100000
sum !< ERROR_THRESHOLD --- 0.102685 !< 0.100000
sum !< ERROR_THRESHOLD --- 0.105320 !< 0.100000
sum !< ERROR_THRESHOLD --- 0.102300 !< 0.100000
sum !< ERROR_THRESHOLD --- 0.100736 !< 0.100000
sum !< ERROR_THRESHOLD --- 0.100824 !< 0.100000
sum !< ERROR_THRESHOLD --- 0.101665 !< 0.100000
sum !< ERROR_THRESHOLD --- 0.102851 !< 0.100000
sum !< ERROR_THRESHOLD --- 0.100173 !< 0.100000
sum !< ERROR_THRESHOLD --- 0.102761 !< 0.100000
```

jeffdaily:
288fa876a93ddf8e56df4ca9e7bdde5406977e2e 
restores the previous behavior of freeing resources after waiting on a signal. This behavior was changed by
aaa7da1
in order to maintain thread safety.  Thread safety has since been restored using `std::atomic_flag`.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4890,6 +4894,7 @@ inline void
 HSABarrier::wait() {
     // Wait on completion signal until the barrier is finished
     hsa_signal_wait_scacquire(_signal, HSA_SIGNAL_CONDITION_LT, 1, UINT64_MAX, waitMode);
+    dispose();
```
scchan, line 4897: Calling `dispose()` in `wait()` doesn't seem right... 
If `wait()` is not called before the `dtor()` of HSAOp, will it cause a memory leak?
Would it be better to call `dispose()` from the `dtor()`? Then, you probably won't need the extra atomic flag since it's supposed to be destroyed by the last owning thread?

scchan (changes requested):

lib/hsa/mcwamp_hsa.cpp
jeffdaily: Admittedly, this is the least straightforward part of this PR.  Yes, `wait()` must be called before calling `dispose()`.  This is done in all `dtor()`s.  The reason for calling `dispose()` inside the `wait()` is to decrement the ref counts of stored dep ops as soon as possible.  This is what we were doing before this PR, but this PR accidentally changed that behavior to only decrement dep ops in the `dtor()`.  `dispose()` will free everything but the `_signal` --- the final `_signal` cleanup was moved to the `HSAOp` `dtor`.  `dispose()` should only be called by a single thread, that's why the `atomic_flag` was added.  You're right that the `dtor()` will only be destroyed by the last calling thread, and it will `wait()` just in case.  But any thread can call `wait()` and we should clean up resources there greedily.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -5258,42 +5114,21 @@ void HSAOp::registerSignalCallback() {
...
-
-    isSubmitted = false;
-
-    // clear reference counts for dependent ops.
-    depAsyncOp = nullptr;
-
-    return _wait_complete_status;
+    if (_signal.handle) {
+        hsa_signal_wait_scacquire(_signal, HSA_SIGNAL_CONDITION_LT, 1, UINT64_MAX, waitMode);
+        dispose();
```
scchan, line 5130: I think `dispose()` should be called at the end of this function regardless of whether `_signal.handle` is valid.  Then in `~HSACopy()`,  only `wait()` would be needed and `dispose()` becomes redundant.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -763,7 +763,7 @@ class HSAOp : public Kalmar::KalmarAsyncOp {
 
     activity_prof::ActivityProf _activity_prof;
 
-    hsa_status_t _wait_complete_status;
+    std::atomic_flag _dispose_flag;
```
scchan, line 766: I think it would be clearer to replace the type of this flag with `std::once_flag`.  Then, we can replace all the calls to `dispose()` with `std::call_once(_dispose_flag, dispose)`.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4660,54 +4604,27 @@ HSADispatch::dispatchKernel(hsa_queue_t* lockedHsaQueue, const void *hostKernarg
...
-        DBOUT(DB_MISC, "wait for kernel dispatch op#" << *this  << " completion with wait flag: " << waitMode << "  signal="<< std::hex  << _signal.handle << std::dec << "\n");
-
         // wait for completion
-        if (hsa_signal_wait_scacquire(_signal, HSA_SIGNAL_CONDITION_LT, 1, uint64_t(-1), waitMode)!=0) {
-            throw Kalmar::runtime_exception("Signal wait returned unexpected value\n", 0);
-        }
-
-        DBOUT (DB_MISC, "complete!\n");
+        hsa_signal_wait_scacquire(_signal, HSA_SIGNAL_CONDITION_LT, 1, UINT64_MAX, waitMode);
+        dispose();
```
scchan, line 4617: `dispose()` should be called regardless of whether the signal is valid.  Then, the call to `dispose` in `~HSADispatch()` could be removed.

scchan (changes requested):

lib/hsa/mcwamp_hsa.cpp
jeffdaily: If we ever go back to not having every dispatch create a signal, we cannot unconditionally `dispose()` here.  In fact, going back to signal-less dispatches will be an easier change after this PR, so I propose keeping this as-is for now.

lib/hsa/mcwamp_hsa.cpp
jeffdaily: The overhead of the `std::atomic_flag` is less than the alternative.  I tried to make it clear with the comment at the top of each `dispose()` implementation.

lib/hsa/mcwamp_hsa.cpp
jeffdaily: That might be true in this case of `HSACopy`, but isn't so for `HSADispatch` (see comment below).  I agree that in some cases the call to `dispose()` becomes redundant, but the `atomic_flag` at the top of each `dispose()` implementation should be quite cheap.

scchan (approved):

PR #1347, opened 2019-12-10
do not allocate a signal for every kernel dispatch

jeffdaily:
This replaces #1173 .


jeffdaily:
Is CI okay?  I'm seeing errors in the test logs such as `There is no device can be used to do the computation`.

jeffdaily:
@david-salinas could you double-check CI health for me?  Thanks.

scchan (changes requested):
Don't we have to update `HSAQueue::wait_no_lock` and `HSAQueue::isEmptyNoLock` to take `HSAQueue::asyncOpsWithoutSignal` into account?  For example if in `wait_no_lock`, I think we may have to enqueue a barrier if there exists ops without a signal?

jeffdaily:
tl;dr No, those functions already handle the no signal case.

Recall that this PR inserts the new `shared_ptr<HSAOp>` into the `asyncOps` vector just like always.  The primary difference is that an extra copy (ref count) of this same `shared_ptr<HSAOp>` is kept in the new `asyncOpsWithoutSignal` vector.  That said, the `HSAQueue::back()` that we use for checking the last inserted op will return the signal-less HSAOp from the `asyncOps` vector just like always.

`HSAQueue::wait_no_lock` should be fine.  It already had logic for this case from back when we didn't create signals for dispatches in earlier ROCm releases.
https://github.com/RadeonOpenCompute/hcc/blob/37c25cae64cd21ca9957d887a11824525222326e/lib/hsa/mcwamp_hsa.cpp#L1635-L1642.  

Same case for `HSAQueue::isEmptyNoLock()`.  It checks for a signal, as well.
https://github.com/RadeonOpenCompute/hcc/blob/37c25cae64cd21ca9957d887a11824525222326e/lib/hsa/mcwamp_hsa.cpp#L1585-L1606

scchan:
@jeffdaily thanks for the explanation.
For some reason, I somehow got the impression that an asyncop without a signal would only get pushed into asyncOpsWithoutSignal but not into the asyncOps vector.  Now it makes sense.

scchan (approved):
LGTM

PR #1348, opened 2019-12-10
fix deadlock in isEmpty

jeffdaily:
Deadlock introduced by 3694bb41312db2a7220c1c6db51352085501a4e0.


jeffdaily:
This deadlock would be rare, if present at all, until #1347 is merged.  It's not clear whether at this time we can create any `HSAOp` without a signal.  I recall at one time there were cases where an `HSACopy` can be created without a signal.  Otherwise, most `HSAOp`s have a signal and would not trigger the locked extra op insertion that this PR fixes.

scchan (approved):
LGTM

PR #1363, opened 2020-01-09
remove amd_math_expf_precise_math.cpp

david-salinas:


scchan:
This is to fix PR1359?

david-salinas:
> This is to fix PR1359?

Yup.  PR1359 just added the .donotrun version of the file, but didn't include actually removing the original file.

PR #1377, opened 2020-01-25
increase KERNARG_BUFFER_SIZE from 512 to 4k

jeffdaily:
Decrease HCC_ASYNCOPS_SIZE from 16k to 1k.
HCC_KERNARG_BUFFER_SIZE is now an environment variable.
HCC_KERNARG_POOL_SIZE is now an environment variable.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -61,7 +61,10 @@
 // kernel dispatch speed optimization flags
 /////////////////////////////////////////////////
 
-// size of default kernarg buffer in the kernarg pool in HSAContext
+// Size of default kernarg buffer in the kernarg pool in HSAContext, in bytes.
+// Increased from 512 to 4k to match CUDA default. See
+// https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#function-parameters
+// When this size is exceeded, on-demand allocation of the kernarg buffer is slow.
 #define KERNARG_BUFFER_SIZE (4096)
 
 // number of pre-allocated kernarg buffers in HSAContext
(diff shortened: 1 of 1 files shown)


jeffdaily:
Since this PR also reduces the asyncops size, it could replace #1261 .

emankov:
Justification for all the numbers is needed.

jeffdaily:
@emankov 

1. CUDA default kernarg size is 4k.  `__global__ function parameters are passed to the device via constant memory and are limited to 4 KB.`  From https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#function-parameters .
2. PyTorch translate model uses a number of kernels with kernargs > 512 bytes, the current default.  Changing the default kernarg buffer size results in a 30% performance improvement since kernargs are no longer allocated on demand.
3. Since kernarg buffer size is increased by 8 times, HCC_ASYNCOPS_SIZE is reduced by 16 times to keep memory use roughly the same in the worst case, assuming two streams fully queuing to the same device.

jeffdaily:
@emankov The most important change in this PR is the increase in the default kernarg buffer size.  If needed, would such a change be acceptable without the other changes?

emankov:
@jeffdaily, thank you for explanation. Could you please add just a few words in comments?

jeffdaily:
@emankov comments added in commit https://github.com/RadeonOpenCompute/hcc/pull/1377/commits/f0e2b40f13086f73565de6142a792f889f29b7b9.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -3848,6 +3853,8 @@ void HSAContext::ReadHccEnv()
     GET_ENV_INT(HCC_MAX_QUEUES, "Set max number of HSA queues this process will use.  accelerator_views will share the allotted queues and steal from each other as necessary");
     GET_ENV_INT(HCC_LAZY_QUEUE_CREATION, "Control lazy HSA queue creation (enabled by default).  Set to 0 to disable it.");
 
+    GET_ENV_INT(HCC_KERNARG_BUFFER_SIZE, "Maximum size of kernel arguments, in bytes. Default 4096.");
+    GET_ENV_INT(HCC_KERNARG_POOL_SIZE, "Number of kernel argument buffers to allocate each time pool is grown. Default 1024.");
```
scchan: I'm working on changing how memory is allocated for kenarg pool and how it's managed so could we remove these knobs for now since they may become obsolete in a near future?  I'm fine with the rest of this PR.

scchan (changes requested):

lib/hsa/mcwamp_hsa.cpp
jeffdaily: Done.  Started over with these two commits and omitted the env var changes.

PR #1388, opened 2020-02-12
restore error return for 2D SDMA copy, fix completion_future valid()

jeffdaily:
PR #1317 inadvertently backed out changes to add reporting of a failed 2D SDMA copy,
originally added in PR #1061. This commit restores those changes.

Further, the implementation of hc::completion_future valid() was incorrect after #1317.
valid() should also check if an asyncOp is present, since we no longer construct a shared_future.

### DIFF
--- a/include/hc.hpp
+++ b/include/hc.hpp
@@ -1280,7 +1280,15 @@ public:
      * completion_future is associated with an asynchronous operation.
      */
     bool valid() const {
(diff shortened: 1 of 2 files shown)


scchan:
LGTM
@gargrahul  Could you review this PR? thanks

gargrahul (approved):
LGTM

PR #1395, opened 2020-02-23
kernarg buffer allocation strategy improvement

scchan:
- add the capability to allocate kernarg buffer in coarse grained GPU memory to leverage caching (off by default)
- implemented a new algorithm to manage kernarg buffers
   * provides a control point when to recycle old kernel buffers (for coarse grained buffers)
   * more efficient free kernarg buffer lookup
   * supports multiple pools with kernarg buffer sizes


scchan:
branch rebased and resolved conflicts

PR #1400, opened 2020-02-27
Installs the text file into a bin directory

rakeshsv:
Text file will be installed into hcc bin directory
`


scchan:
Could you please provide the motivation of this change and then reopen?  Thanks

PR #1401, opened 2020-02-28
remove chmod 777 /opt/rocm from docker files

david-salinas:
This RUN command is no longer needed; and is interfering with image/container creation since rocm 3.1


scchan (approved):
LGTM

PR #1407, opened 2020-03-04
Address a deadlock issue in multi-GPU scenario

mangupta:
- Fixes SWDEV-219322

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -4176,13 +4176,13 @@ void HSAQueue::dispose() {
 
         Kalmar::HSADevice* device = static_cast<Kalmar::HSADevice*>(getDev());
 
+        wait();
+
         // NOTE: needs to acquire rocrQueuesMutex and then the qumtex in this
         // sequence in order to avoid potential deadlock with other threads
         // executing createOrstealRocrQueue at the same time
         std::lock_guard<std::mutex> rl(device->rocrQueuesMutex);
         std::lock_guard<std::mutex> l(this->qmutex);
-
-        wait_no_lock();
 
         // clear asyncOps to trigger any lingering resource cleanup while we still hold the locks
         // this implicitly waits on all remaining async ops as they destruct, including the


jeffdaily:
The error message for the failing tests is curious.  `error while loading shared libraries: libhsa-runtime64.so.1: cannot open shared object file: No such file or directory`

jeffdaily:
I kicked off a unit test run on my local system to see if I could reproduce the failures.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -4176,14 +4176,14 @@ void HSAQueue::dispose() {
 
         Kalmar::HSADevice* device = static_cast<Kalmar::HSADevice*>(getDev());
 
+        wait();
+
```
srinivamd, line 4180: Some comment would help. It seems calling 'wait()' is equivalent to holding `qmutex` and `wait_no_lock()` which is what is lines 4185 below after locking `rocrQueuesMutex`. Does this fix imply that `rocrQueuesMutex` should NOT be held before `qmutex` as it may cause deadlock? If so, shouldn't line 4185 locking of `qmutex` be removed?
Or, could moving acquiring of `rocrQueuesMutex` to just before calling `removeRocrQueue()` help?

scchan:
> The error message for the failing tests is curious. `error while loading shared libraries: libhsa-runtime64.so.1: cannot open shared object file: No such file or directory`

That's a known problem that @david-salinas will address

scchan:
@jeffdaily do we still need an explicit wait if we clear the asyncops vector?

lib/hsa/mcwamp_hsa.cpp
jeffdaily: `wait_no_lock()` could potentially call `EnqueueMarkerNoLock()`.  If the `HSAQueue` does not hold a rocr queue, it will end up calling `createOrstealRocrQueue()` and lock `rocrQueuesMutex`.

Moving the lock on `rocrQueuesMutex` until just before line 4206 might work, too.

jeffdaily:
> @jeffdaily do we still need an explicit wait if we clear the asyncops vector?

Isn't the `wait` is necessary due to `HCC_OPT_FLUSH`?

PR #1409, opened 2020-03-05
PR for auto_update_202003041942_EST

rocm-hcc:


david-salinas (approved):
Failing check is because of a known environment issue in docker container.

PR #1410, opened 2020-03-05
change docker run arguments; remove --device and --group-add

david-salinas:


Jenkinsfile
```diff
@@ -75,7 +75,7 @@ node( 'hcctest' )
   }
 
 // JENKINS-33510: the jenkinsfile dir() command is not workin well with docker.inside()
-  hcc_build_image.inside( '--privileged --device=/dev/kfd --device=/dev/dri --group-add video' )
+  hcc_build_image.inside( '--privileged' )
```
scchan: can we remove `--privileged` and keep the other flags instead?

scchan (changes requested):

Jenkinsfile
david-salinas: yup.  trying that now.

PR #1412, opened 2020-03-05
isEmpty should only rely on underlying HSA queue

jeffdaily:
An external library can call `acquireLockedHsaQueue()` and insert its own packets into the HSA queue.  #1382 caught the case where doing so would put the HSAQueue into an inconsistent state.  However, it did not go far enough.  `isEmpty()` should reflect the contents of the underlying HSA queue (if any), not the contents of the `HSAQueue::asyncOps` vector.

This will become a blocker for future HIP/hcc runtime improvements.

### DIFF
--- a/lib/hsa/mcwamp_hsa.cpp
+++ b/lib/hsa/mcwamp_hsa.cpp
@@ -1606,25 +1606,46 @@ public:
     }
 
     bool isEmptyNoLock() {
-        if (!has_been_used) return true;
-
-        auto lastOp = back();
-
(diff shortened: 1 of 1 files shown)


jeffdaily:
This PR seems to break HIP unit tests.  Needs more testing.

jeffdaily:
HIP unit tests are passing with current PR.  Removed dontmerge labels.

jeffdaily:
After discussion with @scchan, instead of changing `isEmpty` et all, we should find a way to notify HCC's dep tracking.

PR #1413, opened 2020-03-06
disable clock2 test until SWDEV-225934 is resolved

saiislam:
It is currently blocking all automatic merges in the hcc branches of LLVM.
  renamed:    tests/Unit/AMDGPU/clock2.cpp -> tests/Unit/AMDGPU/clock2.cpp.donotrun


david-salinas (approved):
the check will fail because of a known issue with some tests loading/finding libhsa-runtime64.so.1.  so for now you can ignore these failures.  I'm working on a solution - its just an environment issue with the docker container.

PR #1414, opened 2020-03-06
update required packages for HCC for centos 8

david-salinas:


CMakeLists.txt
```diff
@@ -604,6 +604,9 @@ if("${DISTRO_ID}" MATCHES "opensuse" OR
    "${DISTRO_ID}" MATCHES "suse" OR
    "${DISTRO_ID}" MATCHES "sles" )
    set(HCC_GENERAL_RPM_DEP "coreutils, findutils, libelf-devel, pciutils, file, perl-File-Copy-Recursive, perl-File-Listing, perl-File-Which")
+elseif("${DISTRO_ID}" MATCHES "centos" AND 
+       "${DISTRO_RELEASE}" MATCHES "^8")
```
scchan: Could you add similar check for a comparable RHEL release?
We should match for the older version (e.g. CentOS 7) to use pth and have others defaulting to npth

scchan (changes requested):

CMakeLists.txt
david-salinas: ok.  I've updated the change.

scchan (approved):

PR #1420, opened 2020-03-09
add accelerator_view::create_marker(hsa_barrier_and_packet_t)

jeffdaily:
Similar to how `hc::accelerator_view::dispatch_hsa_kernel` allows the caller to pass their own `hsa_kernel_dispatch_packet_t`, add a new function `accelerator_view::create_marker(hsa_barrier_and_packet_t)`.  It behaves nearly identically to `accelerator_view::create_marker(hc::memory_scope)`, but now the caller can specify a nearly complete aql packet.  This could be used to indicate both acq and rel scope as well as an associated hsa signal with a non-default starting value.

### DIFF
--- a/include/hc.hpp
+++ b/include/hc.hpp
@@ -268,6 +268,24 @@ public:
 
     /**
      * This command inserts a marker event into the accelerator_view's command
+     * queue. This marker is returned as a completion_future object. When all
+     * commands that were submitted prior to the marker event creation have
+     * completed, the future is ready.
+     *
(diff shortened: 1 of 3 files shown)


scchan (approved):
LGTM, thanks

jeffdaily:
@scchan I don't think we need this PR now, since we found a solution in HIP that doesn't need this new HCC API.  Pending more testing.

jeffdaily:
Closing this PR.  This change is no longer needed.  We can revisit at a later time if the need comes up again.

PR #1424, opened 2020-03-14
new implementation to handle signal-less ops

jeffdaily:


lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1459,37 +1457,40 @@ class HSAQueue final : public KalmarQueue
...
-                    << " signal="<< std::hex  << ((hsa_signal_t*)op->getNativeHandle())->handle << std::dec
-                    << " commandKind=" << getHcCommandKindString(op->getCommandKind())
-                    << " "
-                    << (op->getCommandKind() == hcCommandKernel ? ((std::static_pointer_cast<HSADispatch> (op))->getKernelName()) : "")  // change to getLongKernelName() for mangled name
-                    << std::endl);
-            if (!op->asyncOpsWithoutSignal.empty()) {
-                throw Kalmar::runtime_exception("new op does not have empty async ops list!", 0);
+        auto op_getting_replaced = asyncOps[asyncOpsIndex]; // ref counts up, if it was a valid HSAOp
+        if (op_getting_replaced != nullptr) {
+            if (op_getting_replaced->getNativeHandle() != 0) {
```
scchan: `getNativeHandle()` returns the address of `_signal`, which is a class member so it will never be `null`
Also, it seems the original check makes more sense to detect whether the signal is valid or not?

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1459,37 +1457,40 @@ class HSAQueue final : public KalmarQueue
...
-                throw Kalmar::runtime_exception("after swap, signal list not empty!", 0);
+            else {
+                // op_getting_replaced has a signal, so wait on it.
+                // This tells us we can free any pending signal-less ops.
+                op_getting_replaced->wait();
+                asyncOpsWithoutSignal.clear();
             }
         }
+
+        if (op->getNativeHandle() == 0) {
```
scchan: see question about `getNativeHandle()` in my previous comment

scchan (changes requested):

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1459,37 +1457,39 @@ class HSAQueue final : public KalmarQueue
...
-                    << (op->getCommandKind() == hcCommandKernel ? ((std::static_pointer_cast<HSADispatch> (op))->getKernelName()) : "")  // change to getLongKernelName() for mangled name
-                    << std::endl);
-            if (!op->asyncOpsWithoutSignal.empty()) {
-                throw Kalmar::runtime_exception("new op does not have empty async ops list!", 0);
+        auto op_getting_replaced = asyncOps[asyncOpsIndex]; // ref counts up, if it was a valid HSAOp
+        if (op_getting_replaced != nullptr) {
+            if (static_cast<hsa_signal_t*>(op_getting_replaced->getNativeHandle())->handle == 0) {
+                // No signal present in the op_getting_replaced, so it won't wait() on it as part of its dtor.
+                // Add it to this queue's vector of signal-less ops to free once we know they're finished.
+                asyncOpsWithoutSignal.push_back(std::move(op_getting_replaced));
```
scchan, line 1465: May I suggest to rename `asyncOpsWithoutSignal` to `asyncOpsWithoutSignalWaitingNextSync`?  I think it would help remembering what this is for.

lib/hsa/mcwamp_hsa.cpp
```diff
@@ -1459,37 +1457,39 @@ class HSAQueue final : public KalmarQueue
...
-            DBOUT(DB_SIG, "attaching " << asyncOpsWithoutSignal.size() << " ops without signals to " << *op
-                    << " signal="<< std::hex  << ((hsa_signal_t*)op->getNativeHandle())->handle << std::dec
-                    << " commandKind=" << getHcCommandKindString(op->getCommandKind())
-                    << " "
-                    << (op->getCommandKind() == hcCommandKernel ? ((std::static_pointer_cast<HSADispatch> (op))->getKernelName()) : "")  // change to getLongKernelName() for mangled name
-                    << std::endl);
-            if (!op->asyncOpsWithoutSignal.empty()) {
-                throw Kalmar::runtime_exception("new op does not have empty async ops list!", 0);
+        auto op_getting_replaced = asyncOps[asyncOpsIndex]; // ref counts up, if it was a valid HSAOp
+        if (op_getting_replaced != nullptr) {
```
scchan, line 1461: I'd prefer that it checks `asyncOps[asyncOpsIndex]` for `!nullptr` and then move it into `op_getting_replaces`.  This makes the lifetime of the ops replaces much clearer (e.g. without this scope).  Otherwise, it will get cleanup later when move-assigning a newer op in this is slot.

scchan (changes requested):

PR #1431, opened 2020-03-20
SWDEV-225038: disable relaxed uniform regions in structurizer

ssahasra:
The LLVM structurizer includes a local relaxation on the criteria that
define uniform regions, that was introduced in [1] and eventually
enabled by default in [2].

[1] https://reviews.llvm.org/D62198
[2] https://reviews.llvm.org/D63198

But this relaxation is affecting correctness. For example, in
SWDEV-225038, for different combinations of compiler version and app
version, it is possible to either make the app hang or produce
incorrect results.

For now, we disable this relaxation in HCC until we have a complete
resolution in LLVM itself.

### DIFF
--- a/lib/clamp-device.in
+++ b/lib/clamp-device.in
@@ -30,7 +30,7 @@ KMOPTOPT="${KMOPTOPT:="-O3"}"
 # KMOPTLLC can be used to pass last-minute options to llc in the backend
 # add "--frame-pointer=none" as deafult option to llc to enable frame
 # pointer elimination, so saved registers can be used for SGPR spill
-KMOPTLLC="${KMOPTLLC:="--frame-pointer=none"}"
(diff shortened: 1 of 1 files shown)


ssahasra:
The build seems to have failed due to unrelated reasons (missing libhsaruntime). How do I retrigger the build?

AlexVlx (approved):
LGTM, thanks - word of warning since you're in the neighbourhood, `--frame-pointer=none` no longer works, since LLC now handles frame pointer persistence based on function attributes; we'll have to modify the FE to pass `-fomit-frame-pointer` for the equivalent result although this is a dubious workaround to begin with.

ssahasra:
This is being fixed in upstream LLVM, so discarding the local HCC change.

PR #1439, opened 2020-03-30
fix creation of /opt/rocm soft link

david-salinas:
We need to install ROCr (HSA RT) after we create the soft link, otherwise the symlinks in the HSA lib do not get created properly on the ROCr install.


aaronenyeshi (approved):

david-salinas:
There may be an issue with CMAKE_INSTALL_PREFIX ... investigating.

PR #1440, opened 2020-03-31
extractkernel objdump args to use double dashes

aaronenyeshi:
More recent llvm-objdump has changed disassemble and mcpu arguments to require double dashes.


scchan (approved):

eidenyoshida (commented):
LGTM

PR #1442, opened 2020-04-08
An optimization to the function of findGlobalPool

zhaozhangjian:
In findGlobalPool, return HSA_STATUS_INFO_BREAK instead of HSA_STATUS_SUCCESS when a fine grained global pool is found, so as to avoid useless iterations when called in the constructor of UnpinnedCopyEngine. Also add an ErrorCheck after this call.


zhaozhangjian:
Any one to review this?

PR #1444, opened 2020-04-15
correct HCC package requirements for centos 7

david-salinas:
My original fix got the order of conditional expressions backwards, OR is evaluated last.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -605,9 +605,9 @@ if("${DISTRO_ID}" MATCHES "opensuse" OR
    "${DISTRO_ID}" MATCHES "suse" OR
    "${DISTRO_ID}" MATCHES "sles" )
    set(HCC_GENERAL_RPM_DEP "coreutils, findutils, libelf-devel, pciutils, file, perl-File-Copy-Recursive, perl-File-Listing, perl-File-Which")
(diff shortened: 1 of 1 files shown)


aaronenyeshi (approved):
LGTM

CMakeLists.txt
```diff
@@ -605,9 +605,8 @@ if("${DISTRO_ID}" MATCHES "opensuse" OR
    "${DISTRO_ID}" MATCHES "suse" OR
    "${DISTRO_ID}" MATCHES "sles" )
    set(HCC_GENERAL_RPM_DEP "coreutils, findutils, libelf-devel, pciutils, file, perl-File-Copy-Recursive, perl-File-Listing, perl-File-Which")
-elseif( "${DISTRO_ID}" MATCHES "centos" OR 
-        "${DISTRO_ID}" MATCHES "redhatenterpriseserver" AND
```
scchan, line 609: I think you should just add the parentheses around the `OR` expression.

scchan (changes requested):

CMakeLists.txt
david-salinas: yeah.  updated.

scchan (approved):

